PR #7, opened 2022-01-18
Allow custom form_field wrapper classes

fcarlislehg:


mplatts:
Nice.. I was actually about to implement something like this... sometimes I don't need that margin but still want to use form_field

fcarlislehg:
I just wanted it tightened up to `mb-2`. :)

PR #11, opened 2022-02-15
Solid Heroicons are meant to be w-5 h-5

doughsay:
According to the Heroicons website, the solid icons are supposed to use `w-5 h-5` and the outline icons use `w-6 h-6`.

This PR fixes this discrepancy, and also introduces a sort in the mix task when generating the modules, so that future re-generations don't cause unnecessary churn in the diffs (sorry this one is so big...).  I also re-generated using Heroicons `v1.0.3`.


mplatts:
Cool thanks for that!

PR #12, opened 2022-02-19
Allow custom attributes to be passed to root elements

MortadaAK:
This PR allows users to pass additional data to the root element of the components where it will make testing easier


mplatts:
Thanks, I resolved the conflicts locally and did a separate [pull request](https://github.com/petalframework/petal_components/pull/15) (not sure how to resolve in this PR). I renamed "extra_attributes" to "extra_assigns" as we had used that variable name previously.

Still TODO: update the heroicons generator to use "extra_assigns" instead of "extra_attributes"

PR #23, opened 2022-03-09
Forward class attribute to table components

soundmonster:
Ran into this while trying to hide table columns on a small screen. It's possible to hide `<td>` elements but it didn't
work for `<th>`. This patch makes sure to forward the `class` assign to every component in `PetalComponenets.Table`.


nhobes:
Thanks for that. Merged.

PR #28, opened 2022-03-25
Do not accidentally override form classes

soundmonster:
- Do not leak the `class` assign as it's already set from `classes`
- Fix tests


nhobes:
Thanks for this! Merged.

PR #29, opened 2022-04-02
Begin adding SlideOver component

ethancrawford:
This introduces a basic SlideOver component adapted from the existing Modal.
TODO:
- Work out how to handle max-height/max-width, if at all.
- Add RTL/writing mode support.
- Make any further changes as requested by the Petal components core developers.

Some gifs:
Mobile view:
slide_over_mobile2

Desktop view:
slide_over

### DIFF
--- a/lib/petal_components.ex
+++ b/lib/petal_components.ex
@@ -18,6 +18,7 @@ defmodule PetalComponents do
         Pagination,
         Link,
         Modal,
+        SlideOver,
         Tabs,
         Card,
         Table
--- a/lib/petal_components/slide_over.ex
+++ b/lib/petal_components/slide_over.ex
@@ -1,0 +1,142 @@
+defmodule PetalComponents.SlideOver do
+  use Phoenix.Component
+  alias Phoenix.LiveView.JS
+  # prop title, :string
+  # prop size, :string
+  # slot default
+  def slide_over(assigns) do
+    origin = assigns[:slide_over] || "start"
+    assigns =
+      assigns
+      |> assign_new(:classes, fn -> get_classes(assigns) end)
+      |> assign_new(:extra_assigns, fn ->
+        assigns_to_attributes(assigns, ~w(
+          classes
+          class
+          title
+          size
+          slide_over
+        )a)
+      end)
+
+    ~H"""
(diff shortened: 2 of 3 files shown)


ethancrawford:
As mentioned in the commit message/top comment above, I'm also still in the process of deciding what to do with `max-width`/`max-height` - the Modal sets a max-width, but a similar size restriction may need to be handled differently for the SlideOver given that the longest dimension of the content block may be either horizontal _or_ vertical. Open to suggestions...

I'm interested in any other comments about my initial implementation so far as well.

mplatts:
Code looks good to me. Will have more of a play with it today. Yeah sizing could be an issue. I noticed Tailwind UI leave a gap on the side even on mobile. Not sure if that's better or worse though - will chat about with Nic today

CleanShot 2022-04-04 at 06 28 20

mplatts:
Hi @ethancrawford , we are happy enough to pull this in and make some minor style changes. What did you mean by "Add RTL/writing mode support." ?

ethancrawford:
My thoughts were just to use [logical properties and values]https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties where useful (or otherwise handle RTL writing with JavaScript if necessary) and ensure that different [writing modes](https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode) would be supported 🙂

ethancrawford:
(Such that by adding something like `dir="rtl"` to the html, things 'just worked')

mplatts:
maybe for now anyone who wants that can do it inside the inner_block? eg.

```
<.slide_over>
  <div dir="rtl">Write this text right-to-left!</div>
<./slide_over>
```

ethancrawford:
Happy for you to merge now if you want to. I think ultimately using logical properties/values where they can be is a good step in the long run, if you're happy to consider it - but sure, it's not an immediate blocker 👍

mplatts:
Yep maybe for now we just get it out and it will cover most use cases, and then can accept PR's to fine tune it. Thanks a lot for this

ethancrawford:
You're welcome, looking forward to using it!

PR #34, opened 2022-04-28
Generate prettier classes

moogle19:
`build_class/2` is basically `Enum.join/2` with some improvements
for building the class names. For example empty strings are not joined
into the class names and the default joiner is set to a single whitespace to avoid specifying it every time.

Before:
Screenshot 2022-04-28 at 09 54 49

After:
Screenshot 2022-04-28 at 10 33 03

### DIFF
--- a/lib/petal_components/alert.ex
+++ b/lib/petal_components/alert.ex
@@ -1,5 +1,7 @@
 defmodule PetalComponents.Alert do
   use Phoenix.Component
+  import PetalComponents.Class
+
   alias PetalComponents.Heroicons
 
   # prop color, :string, options: ["info", "success", "warning", "danger"]
@@ -57,7 +59,7 @@ defmodule PetalComponents.Alert do
             </div>
 
             <%= if @close_button_properties do %>
-              <button class={Enum.join(["p-2 mouse-hover flex hover:rounded ", " ", get_dismiss_icon_classes(@color), " "])} {@close_button_properties}>
+              <button class={build_class(["p-2 mouse-hover flex hover:rounded", get_dismiss_icon_classes(@color)])} {@close_button_properties}>
                 <Heroicons.Solid.x class="self-start w-4 h-4" />
               </button>
             <% end %>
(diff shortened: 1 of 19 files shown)


mplatts:
@moogle19 thanks for this! 

We've since discovered that you can actually just pass a list to the class attribute and HEEX will concatenate them for us:

```html
<h1 class={["bg-black", "p-5", ""]}>Hi there</h1>

<!-- Outputs to: -->
<h1 class="bg-black p-5 ">Hi there</h1>
```

However, as you can see your solution is better as HEEX doesn't remove the whitespace, so happy to merge this.

Another thing we'd love is the ability to override previous classes. 

eg imagine this was in a component:

```
<h1 class={build_class(["bg-black", "p-5"], @class)}>Hi there</h1>
```

I might not want the "p-5", so I pass in `class="p-3"` - in this case it should override the "p-5". Obviously this could be a bit of a tough task given the number of potential classes. But maybe we could start with just the padding/margin classes...  do you have any thoughts/ideas on how we could go about this? It may be too difficult.

moogle19:
My first naive approach would be something like this:

```elixir
defmodule Class do
  @doc """
  Builds a class based on a set of default classes and specified overrides for them
  """
  def build_class(defaults, overrides) when is_binary(defaults),
    do: build_class(String.split(defaults), overrides)

  def build_class(defaults, overrides) when is_binary(overrides),
    do: build_class(defaults, String.split(overrides))

  def build_class(defaults, overrides) when is_list(defaults) and is_list(overrides) do
    defaults_map = get_class_map(defaults)
    overrides_map = get_class_map(overrides)

    Map.merge(defaults_map, overrides_map)
    |> Map.values()
    |> Enum.join(" ")
  end

  # Converts a list of classes to a map identified by their class name prefix
  defp get_class_map(classes) do
    classes
    |> Enum.map(fn class -> {get_class_id(class), class} end)
    |> Map.new()
  end

  # Gets the prefix of a class name (without the part after the last `-` character)
  defp get_class_id(class) do
    class
    |> String.reverse()
    |> String.split("-", parts: 2, trim: true)
    |> case do
      [class] -> class
      [_ | [prefix]] -> prefix
    end
    |> String.reverse()
  end
end
```


Since the class order is not important, we can split the classnames by whitespace and then put them into a map, identified by their class prefix (everything before the last `-`, e.g. `"p-5 place-items-start" => %{"p" => "p-5", "place-items" => "place-items-start"}`).

Then we just merge the overrides map over the defaults map and get the remaining values.

I am not 100% sure if this would always work with tailwindcss and the naming scheme used there.

This also would override class names like `table` with `table-auto` during merging, because they share the same prefix, which could also be undesirable.

PR #40, opened 2022-05-06
Switch should use primary colour instead of blue

abradburne:
The form switch component is currently using a hardcoded colour blue rather than the configured primary colour.


mplatts:
Yeah I think I originally had it as primary, then changed it to blue as I wasn't sure if people wanted a form input to be branded. I guess having it blue is signalling some kind of color, so it may as well be on brand. So happy to merge this if you're all good @nhobes

nhobes:
All good with me. Thanks @abradburne!

PR #44, opened 2022-05-22
Fix heroicons render method

RobinBoers:
In 538f2b1348e31537b6f41d68b044d5655d2edd2e the docs for the Heroicons component were updated to use `type="icon_name"` instead of `icon={:icon_name}`:

```heex
<PetalComponents.Heroicons.Solid.render icon={:home} class="w-6 h-6" />
<PetalComponents.Heroicons.Solid.render type="home" class="w-6 h-6" />
```

But it doesn't work since the render function wasn't updated.

### DIFF
--- a/lib/mix/tasks/heroicons/generate.ex
+++ b/lib/mix/tasks/heroicons/generate.ex
@@ -17,22 +17,21 @@ defmodule Mix.Tasks.Heroicons.Generate do
       @moduledoc \"\"\"
       Icon name can be the function or passed in as a type eg.
       <PetalComponents.Heroicons.Solid.home class="w-5 h-5" />
-      <PetalComponents.Heroicons.Solid.render type="home" class="w-5 h-5" />
+      <PetalComponents.Heroicons.Solid.render icon="home" class="w-5 h-5" />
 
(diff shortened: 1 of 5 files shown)


mplatts:
Thanks! 

- Could you modify `apps/petal_components/test/petal/heroicons_test.exs` to test this?
- More of an Elixir question - you use `String.to_existing_atom` - how are all the atoms existing? Do function names count as atoms or something?
- Can you make it work both ways (string or atom)?

mplatts:
Just realised it should only be `icon`, not `type`. Not sure why `type` was in the comment, as I've been using `icon` elsewhere in this project. By both ways I mean:

```
<Heroicons.Solid.render icon="home" />
<Heroicons.Solid.render icon={:home} />
```

mplatts:
> > More of an Elixir question - you use String.to_existing_atom - how are all the atoms existing? Do function names count as atoms or something?
> 
> I just assumed they were since module names are atoms too, and it seems to work (I tested it with a project that uses a lot of different icons and they all showed up correctly using the new syntax for render).

Yeah it must be the case. Thanks!

PR #45, opened 2022-05-26
Make compatible with liveview 0.18

### LINKED ISSUE #42: FYI fails to compile on HEAD from phoenix_live_view
I have been running on `0.18.0-dev` of phoenix_live_view (commit `e24527536890be2f0565c4a6d122a89e819df58e` and later) for the bug fixes to the heex formatter. Obviously this is a pre-release commit, but I wanted to let you know that petal_components fails to compile.

```bash
❯ mix deps.compile petal_components
==> petal_components
Compiling 25 files (.ex)

== Compilation error in file lib/petal_components/modal.ex ==
** (CompileError) lib/petal_components/modal.ex:77: cannot find or invoke local \\/2 inside match. Only macros can be invoked in a match and they must be defined before their invocation. Called as: js \\ %JS{}
    (phoenix_live_view 0.18.0-dev) expanding macro: Phoenix.Component.__pattern__!/2
    lib/petal_components/modal.ex:77: PetalComponents.Modal.hide_modal/1
could not compile dependency :petal_components, "mix compile" failed. Errors may have been logged above. You can recompile this dependency with "mix deps.compile petal_components", update it with "mix deps.update petal_components" or clean it with "mix deps.clean petal_components"
```

Everything works fine on `0.17.9` assuming once you downgrade you run `mix deps.compile phoenix_live_view` first.

dvic:
The next version of liveview introduces a `link` component, not sure if this can be used here, but this PR makes it compatible with the next version by not importing the `Link` module.

Fixes #42

### DIFF
--- a/lib/petal_components/button.ex
+++ b/lib/petal_components/button.ex
@@ -3,9 +3,9 @@ defmodule PetalComponents.Button do
 
   alias PetalComponents.Loading
   alias PetalComponents.Heroicons
+  alias PetalComponents.Link
 
   import PetalComponents.Class
-  import PetalComponents.Link
 
   # prop class, :string
   # prop color, :string, options: ["primary", "secondary", "info", "success", "warning", "danger", "gray"]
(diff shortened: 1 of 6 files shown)


mplatts:
Thanks @dvic . Wondering if we should either remove this component completely or rename it to not clash with the official `<.link>`? eg we could call ours `<.a>`.

dvic:
I think it would be best to deprecate this component and remove it in a later release. I guess it depends on how long you want to support liveview < 0.18 :)

Issue #30, opened 2022-04-13
Modal - should be able to pass a target to close_modal 


mplatts:
Allow modals to target `@myself` for when a modal is inside a live_component.

```html
<.modal title="Modal" close_modal_target={@myself}>
    <.p>Content</.p>
</.modal>
```

```elixir
# In the component code something like:
JS.push("close_modal", target: @close_modal_target)
```

mplatts:
Oh yeah I did implement this - forgot to close this issue. Thanks haha

PR #58, opened 2022-07-05
Only send close_modal event to target if provided

nikkocampbell:
Seems related to #30 and some recent changes that were added to support modals in components. As-is, this was still sending the `close_modal` event to the LiveView page when used inside a component and a specific target is passed, as well as to the target. This'll make it so only the target gets the event

### DIFF
--- a/lib/petal_components/modal.ex
+++ b/lib/petal_components/modal.ex
@@ -90,12 +90,11 @@ defmodule PetalComponents.Modal do
         },
         to: "#modal-content"
       )
-      |> JS.push("close_modal")
 
     if close_modal_target do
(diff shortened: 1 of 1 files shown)


mplatts:
Thanks! That was my bad

mplatts:
pushed a release with this in it: 0.17.1

Morzaram:
Hey I think it'd be a good idea to update the [modal doc](https://petal.build/components/modals) about being able to assign the 
`close_modal_target={@myself}`

Not sure if it's worth opening an issue about this, hence my comment here. Sorry in advance!

PR #59, opened 2022-07-11
Fix avatar passing src as empty string and name together

addstar34:
If you pass `src` and `name` to avatar and `src` is `""` then it will still try to render the avatar with an img tag instead of a div with initials.

E.g. this is broken when `src == ""`:
```
<.avatar size="sm" src={@user.avatar_url} name={@user.name} />
```


nhobes:
Thanks for this!

PR #61, opened 2022-07-21
Make accordion items dynamic

hengermax:
The `accordion` component currently works by employing `<:item>` slots. The downside of this approach is that one cannot wrap slot items in a loop, e.g.

```html
<.accordion>
  <%= for thing <- @my_things do %>
    <:item><!-- snip --></:item>
  <% end %>
</.accordion>
```

This PR adds an assign called `entries` to the `accordion` such that it is possible to add dynamic entries into the component. This approach with `entries` was taken because of the `heading` that is expected for each `<:item>`. The `accordion` logic expects the `heading` to be set for each `<:item>`, and currently phoenix liveview does not allow retrieving and using a slot value within the same HTML node. e.g.

```html
<:item let={my_item} heading={my_item.my_value}></:item>
```

Is not allowed.

There are other approaches:
- Rewriting the `accordion` into `accordion` and `accordion_item`, where accordion uses CSS that sets the borders of the first/last element using CSS selectors.
- Rewriting the `accordion` such that it uses slots `:heading` and `:item`.
- Making a `dynamic_accordion` and a `static_accordion`.


mplatts:
Thanks @hengermax ! Could you add a test or two to `accordion_test.exs`?

hengermax:
Hey @mplatts , tests are in there now. Good that you asked me, because I completely overlooked some possibilities. During testing I found out that your phoenix (liveview) dependencies are a bit outdated (`render_component` testing function has a bug that has been presently fixed). Do you want me to bump them too?

mplatts:
Tests look good 🥇 . Yeah bumping them would be great thanks - then I'll merge

hengermax:
@mplatts Versions are bumped without touching mix.exs

PR #64, opened 2022-07-22
Fix 'modal' IDs used at SlideOver component

hengermax:
The `SlideOver` component was using the modal overlay/content IDs `modal-overlay` and `modal-content`. Now it is an edge case, but if you had both a `SlideOver` and a `Modal` opened, then closing the `Modal` would cause the `SlideOver` to disappear as well (but without any user interaction possible in the area that was covered by the `SlideOver`).

To fix this, these IDs are now `slide-over-overlay` and `slide-over-content`. A truly "perfect" solution would generate an ID per instance of `SlideOver`/`Modal` that is on the page, such that multiple modals/slideovers are can be present on a single page.

Tests are not included, since proper testing depends on executing javascript.

### DIFF
--- a/lib/petal_components/slide_over.ex
+++ b/lib/petal_components/slide_over.ex
@@ -20,7 +20,7 @@ defmodule PetalComponents.SlideOver do
     ~H"""
     <div {@rest} id="slide-over">
       <div
-        id="modal-overlay"
+        id="slide-over-overlay"
         class="fixed inset-0 z-50 transition-opacity bg-gray-900 dark:bg-gray-900 bg-opacity-30 dark:bg-opacity-70"
         aria-hidden="true"
       >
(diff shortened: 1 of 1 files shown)


hengermax:
WIP: Apparently there still is a user-interaction bug when closing a modal when a slide-over is open, one moment.

hengermax:
Nope, nevermind, that had to do with not resetting the `enable` flag in the code:

```heex
<%= if @enable do %>
<.modal><!-- snip --></.modal>
<% end %>
```

hengermax:
Hm, perhaps don't put this in yet. A question that should be answered first is how something should behave when both a slide-over and a modal are opened. Right now clicking on a modal with a slide-over underneath it will close the slide-over (due to the `phx-click-away` handler)

So how should this behave? Should pressing `esc` close both of them? Should clicking outside of the modal region close *just* the modal, or both the modal and the slideover. I suppose that `esc` should close the topmost overlayed component, and that clicking outside the modal should *only* close the modal. 

Thoughts?

mplatts:
I think what you have now is great. To me a slide over is like another version of a modal and I can't really think of a use case for nested modals.

PR #67, opened 2022-08-01
Allow heroicon names containing "-"

RobinBoers:
In #44 I fixed the render method for heroicons to work with strings. Before the names of the icons were atoms and the dashes in their names were replaced with underscores, but since we're using strings now there is no reason to use underscores.

This PR uses `String.replace` to replace all "-" with "_" before calling the function which makes the names from heroicons.com work out of the box, but icon names with underscores should keep working.

Edit: 
Regarding the atoms existing in the last PR
> I just assumed they were since module names are atoms too, and it seems to work (I tested it with a project that uses a lot of different icons and they all showed up correctly using the new syntax for render).

Function names do not count as atoms. I think they all exist because they are created in the generator.


mplatts:
Nice, makes sense. Can you add a test to heroicons_test.exs?

RobinBoers:
Yeah, sure!

PR #72, opened 2022-08-09
add title elements to SVGs

cleaver:
- optional <title> tag for SVGs
 - improves accessibility


cleaver:
I created this PR to improve the accessibility of the SVG icons by adding a [title element](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title)

mplatts:
Hey @cleaver, thanks for contributing! 

The `outline.ex` + `solid.ex` files are actually auto generated by a mix command - see `generate.ex`. So if you'd like to modify outline/solid you will need to modify the `generate.ex` file and then run these commands to generate:

```
# in petal_components root
git clone https://github.com/tailwindlabs/heroicons.git
mix heroicons.generate
```

Since the svgs are copied over from the heroicons library - you will have to do either a String.replace or maybe use https://github.com/philss/floki or something to get the title attribute in there. Hope this all makes sense

cleaver:
Cool, makes sense. I'll have a look at the best way to do that.

cleaver:
I went with the regex rather than adding a dependency. Seems to do the job.

cleaver:
I haven't added anything to documentation. I see the generated documentation, but it looks like the website is private. Is that right?

If the PR looks OK to you so far, I was thinking to just add a line to the example code in the moduledoc, unless you were thinking of a way to improve on that.

mplatts:
Looks good. Yeah, we will handle the documentation.

The only thing is that we may introduce other icon libraries in future. So `PetalComponents.Heroicons.Attributes` might be better as just `PetalComponents.Svg`. As `<Svg.title title="Home" /> makes sense and can be used with other svg icons libs.

mplatts:
Thanks - hmm is the file size resulting in a big problem when you deploy? I haven't seemed to notice a problem deploying to fly.io at least. I think unless it's truly painful, it's not worth introducing another step.

cleaver:
> Thanks - hmm is the file size resulting in a big problem when you deploy? I haven't seemed to notice a problem deploying to fly.io at least. I think unless it's truly painful, it's not worth introducing another step.

I've been thinking about this over the weekend and it does seem a bit over-complex. I don't think it would be an issue at run time. File under premature optimization. :)

Issue #77, opened 2022-09-01
Use Heroicons Elixir library


mplatts:
[Heroicons](https://heroicons.com/) recently upgraded to 2.0.

We could either upgrade our own icon component builder, or just import this separate Hex lib: https://github.com/mveytsman/heroicons_elixir

chvanikoff:
My vote is for the lib. Let your lib focus on the components, not the icons. It should be trivial to migrate.

mplatts:
We have renamed `Heroicons` to `HeroiconsV1`. To use Heroicons V2, please use https://github.com/mveytsman/heroicons_elixir

PR #79, opened 2022-09-08
Update heroicons generator for v2

f0rest8:
Update to the heroicons generator. Would also need to update documentation as some name changes (e.g. "mail" -> "envelope"). I just separated into two functions, `loop_directory/1` and `loop_mini_directory/1`. I imagine you probably want to streamline it a bit? 😊

By updating your existing generator, you can keep your same functionality:

```elixir
# 24
<PetalComponents.Heroicons.Solid.home class="w-5 h-5" />
<PetalComponents.Heroicons.Solid.home title="Optional title for accessibility" class="w-5 h-5" />
<PetalComponents.Heroicons.Solid.render icon="home" class="w-5 h-5" />

<PetalComponents.Heroicons.Outline.home class="w-6 h-6" />
<PetalComponents.Heroicons.Outline.home title="Optional title for accessibility" class="w-6 h-6" />
<PetalComponents.Heroicons.Outline.render icon="home" class="w-6 h-6" />

# 20 (Mini)
<PetalComponents.Heroicons.Mini.Solid.home class="w-5 h-5" />
<PetalComponents.Heroicons.Mini.Solid.home title="Optional title for accessibility" class="w-5 h-5" />
<PetalComponents.Heroicons.Mini.Solid.render icon="home" class="w-5 h-5" />

<PetalComponents.Heroicons.Mini.Outline.home class="w-6 h-6" />
<PetalComponents.Heroicons.Mini.Outline.home title="Optional title for accessibility" class="w-6 h-6" />
<PetalComponents.Heroicons.Mini.Outline.render icon="home" class="w-6 h-6" />
```


mplatts (changes requested):
Maybe through a test in heroicons_test.exs for the mini icon?

f0rest8:
Thanks! I updated and tested in the Petal Development umbrella app. Just need to add a test now. 😊

f0rest8:
Okay I think I cleaned up little typos, added tests for the `Heroicons.Mini.Solid`, updated tests that were calling v1 heroicons, and updated the `loop_directory/1` based on your suggestion. 

Although instead of multiple arguments, I went with one argument. Hmm... perhaps multiple arguments is preferable?

```elixir
# Looks like this:

Enum.each(
  [
    ["24/solid", "Solid", "solid"],
    ["24/outline", "Outline", "outline"],
    ["20/solid", "Mini.Solid", "mini/solid"]
  ],
  &loop_directory/1
)
```

f0rest8:
Here's also the documentation for upgrading to v2. This might be useful to include the link and/or icon name change list in documentation for people: [Heroicons v2.0.0 Release Notes](https://github.com/tailwindlabs/heroicons/releases/tag/v2.0.0)

mplatts:
Looking better, but now the variable name "folder" doesn't accurately label what it is.

Though after some thought I think it might be best to maybe shelve this PR and use another [Heroicons library](https://github.com/mveytsman/heroicons_elixir) as mentioned in this issue: https://github.com/petalframework/petal_components/issues/77

I feel bad as you have spent some time on this - but if we use this other library then perhaps it will save more time in future as it seems to be well maintained and saves us duplicating these upgrades. What do you think?

f0rest8:
I agree and think it'd be better to drop and go with the Heroicons library. No worries, I enjoyed tinkering.

PR #84, opened 2022-09-20
Accepts function to define the path in Pagination component

Wigny:
Allows the Pagination component to receive a function as a parameter that will define the path of the page.

This change will bring flexibility so that we can use helper functions like the [Phoenix Routing](https://hexdocs.pm/phoenix/routing.html#path-helpers) ones to generate the path:  

```elixir
<.pagination path={fn page -> Routes.user_path(Endpoint, :index, [page: page]) end} current_page={1} total_pages={10} />
```

> The above function should dynamically generate the path `/users?page=#{page}`.

### DIFF
--- a/lib/petal_components/pagination.ex
+++ b/lib/petal_components/pagination.ex
@@ -98,15 +98,25 @@ defmodule PetalComponents.Pagination do
     build_class([base_classes, active_classes, rounded_classes])
   end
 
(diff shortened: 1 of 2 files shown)


mplatts:
🙏  much better

mplatts:
will go into our next release in the coming days

PR #86, opened 2022-09-23
Fix background color for disabled fields in dark mode

moogle19:
Added a better background color for disabled fields in dark mode and also removed a duplicate `dark:border-gray-600` for select fields.


mplatts:
👍 thanks!

PR #97, opened 2022-10-17
New declarative assigns

nhobes:


mohammedzeglam-pg:
thats is awesome but what about using this embedded pages

https://github.com/phoenixframework/phoenix_live_view/blob/cd93e0f3645d89b47d60a22eb8a085205d933df0/lib/phoenix_component.ex#L1386

mplatts:
wow that's cool - hadn't seen that. How do you propose we use it in Petal Components? Could help with installing them.

mohammedzeglam-pg:
> wow that's cool - hadn't seen that. How do you propose we use it in Petal Components? Could help with installing them.

https://github.com/mohammedzeglam-pg/e_commerce/blob/c869cdfa4bbeb234a654b5313c83b3d5b1d20eba/lib/e_commerce_web/shared/elements.ex#L1

Rio517:
Looking forward to seeing this merged!

Issue #95, opened 2022-10-12
Add tags to the card content


obsidienne:
Hi,

I think the <.card_content> component should allow a tags attributes. Something like this

```
<.card_content category="My category" heading="Awesome heading" tags={[tag1, tag2, ..., tagn]}>
</.card_content>
```

The tag being a <.badge color="primary" variant="outline">

Is it a pull request interesting for you ?

nhobes:
Sounds good! Do you have an example of the use cases/what it would look like?

obsidienne:
Hi,

Something like that :)

<img width="429" alt="image" src="https://user-images.githubusercontent.com/1252097/195794151-6311b2a9-aa3a-4d3a-a19c-0745c8f932fa.png">

nhobes:
That does look great and I can see the use case. Feel free to submit a PR if you get around to it and I'll take a look :)

nhobes:
Closing for now.

PR #98, opened 2022-10-17
Add tags to card header

obsidienne:
Hi,

Following ticket #95

### DIFF
--- a/lib/petal_components/card.ex
+++ b/lib/petal_components/card.ex
@@ -2,6 +2,7 @@ defmodule PetalComponents.Card do
   use Phoenix.Component
 
   import PetalComponents.Helpers
+  alias PetalComponents.Badge
 
   # prop class, :string
   # prop variant, :string
@@ -66,6 +67,7 @@ defmodule PetalComponents.Card do
       |> assign_new(:category_color_class, fn -> "text-primary-600 dark:text-primary-400" end)
       |> assign_new(:inner_block, fn -> nil end)
       |> assign_new(:category, fn -> nil end)
+      |> assign_new(:tags, fn -> nil end)
       |> assign_new(:heading, fn -> nil end)
       |> assign_rest(~w(class category_color_class inner_block category heading)a)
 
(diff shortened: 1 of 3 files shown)


mplatts:
Thanks, can you remove the version change? We will update this when we push to hex.pm

mplatts:
Also maybe add a new test on the end of `card_test.exs`

obsidienne:
Done

mplatts:
Thanks for that. I just had a look.

CleanShot 2022-10-23 at 10 26 41

Design-wise I think it might get a bit busy up the top once you have tags - they probably should be gray too so they don't look so prominent. 

eg. I've been playing around with them too and putting them down on the bottom 

CleanShot 2022-10-23 at 10 28 51

This means you can just put them in the card's inner block and customise them exactly how you want (eg, you might want them to be clickable).

mplatts:
Maybe we just make the top right an optional slot? 

```
    <.card>
        <:top_right>
          <.tag content="Tag 1" />
          <.tag content="Tag 2" />
        </:top_right>

        <.card_content category="Article" heading="Enhance your Phoenix development">
          <div class="mt-4 font-light text-gray-500 text-md">
            Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vivamus eget leo interdum, feugiat ligula eu, facilisis massa. Nunc sollicitudin massa a elit laoreet.
          </div>
        </.card_content>
      </.card>
```

mplatts:
Though if we do that, we should probably convert all inner card stuff to slots:

- `<.card_media src="">`  => `<:card_media />`
- `<.card_content>` => `<:card_content>`
- `<.card_footer>` => `<:card_footer>`

mplatts:
Thoughts @obsidienne  or anyone?

obsidienne:
Hi,

I love the idea of using slots for the tags because maybe some tags need to be inverted or outlined according to the webapp design... and IMHO using slots creates less clutered HEEX.

Adding them in each part of a card is interesting too.

obsidienne:
Regarding creating a slot for each part of card, is it ok to have slots inside of slots ?

obsidienne:
Hi, do you need some help for this pull request ? Do I need to change the code ?

mplatts:
Sorry @obsidienne I've been off work due to health issues. Let me discuss with Nic. I think ultimately we need something that is quite customisable and so the slot approach might be better for the top right area than tags. A problem with tags is if there is a long category and more than say 3 tags then what will happen with the wrapping? It has the potential to get messy

nhobes:
@obsidienne I agree with @mplatts's sentiments about placing the tags in the top right and that while it looks great for 2 or 3, you start to experience issues when you have more (see img). It might be be better as @mplatts suggested to place them down the bottom and defaulting to a more muted gray color.

CleanShot 2022-11-15 at 13 52 53

obsidienne:
Sure, do you want an updated PR or are you doing the work on your side ?

mplatts:
Closing for now. Just submit a new PR if you make progress. Cheers

Issue #113, opened 2022-12-14
Button missing attribute `phx_disable_with`


krns:

tmaszk:
<.button> is also missing `phx_click` and `data` attributes

mplatts:
Closing but if you still have issues just give a reproducible example and I'll fix it!

PR #129, opened 2023-01-12
Update form_attrs with data to stop warnings

mrnovalles:
While compiling our project with the latest update we started receiving some warnings:

```
warning: undefined attribute "data" for component PetalComponents.Button.button/1
  lib/tool/live/path/show.html.heex:110: (file)

```

This commit adds these attributes to the allow-list to remove such warnings


paynegreen:
You need to change the implementation from `data={[confirm: "Are you sure?"]}` to `data-confirm="Are you sure?"` Reference issue: https://github.com/petalframework/petal_components/issues/113

PR #131, opened 2023-01-13
Add missing supported attributes

philipgiuliani:
This just adds some attributes which were missing in our application. All of them are in the html specification.


mplatts:
Thanks, released on v0.19.10

PR #134, opened 2023-01-23
Allow data field in `button`

moogle19:
Allow `data` field in button


mplatts:
Not sure we need this? 

I think `data` should always be used like: `data-foo="bar"`

PR #148, opened 2023-02-25
fix(form_field): fixed compilation warning caused by hidden_input not…

BobbieBarker:
I noticed that the docs have an example for using hidden input with `form_field/1`

```
<!-- Explicit option  -->
<.hidden_input form={f} field={:your_field} />

<!-- All inclusive option  -->
<.form_field type="hidden_input" form={f} field={:your_field} />
```

But the `@input_types` module attribute for `PetalComponents.Form` does not include `hidden_input` which causes a compilation warning when trying to use form_field with hidden_input. I also added an extra function clause to `form_field/1` to handle hidden input. My change prevents the wrapper elements from being generated with the hidden input. I realize this is a little opinionated, but `hidden_input` is unique.  It was kind of a surprise to me when I noticed the hidden input showing up in the DOM. If you don't want this change, let me know and I'll drop it from the PR.

### DIFF
--- a/lib/petal_components/form.ex
+++ b/lib/petal_components/form.ex
@@ -59,8 +59,10 @@ defmodule PetalComponents.Form do
     "checkbox",
     "checkbox_group",
     "radio_group",
-    "switch"
+    "switch",
+    "hidden_input"
   ]
+
   attr(:form, :any, doc: "the form object", required: true)
   attr(:field, :atom, doc: "field in changeset / form", required: true)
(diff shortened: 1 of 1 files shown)


mplatts:
Thanks @BobbieBarker! Yes you're right - the wrapper elements are not needed.

mplatts:
Out now - https://github.com/petalframework/petal_components/releases/tag/v1.0.3

PR #149, opened 2023-02-26
Disabled buttons and links should not do anything when clicked

forest:


lib/petal_components/link.ex
```diff
@@ -9,35 +9,55 @@ defmodule PetalComponents.Link do
...
+    assigns = update_in(assigns.rest, &Map.drop(&1, [:"phx-click"]))
+
+    ~H"""
+    <button class={@class} disabled={@disabled} {@rest}>
+      <%= if @label, do: @label, else: render_slot(@inner_block) %>
+    </button>
+    """
+  end
+
+  def a(%{disabled: true} = assigns) do
```
mplatts, line 54: What if we moved this to the top, and always drop phx-click and always render a button. Then we don't need to change the others as any disabled `a` will hit this first?

mplatts: Or, if we still want <a> s to appear, just move both these disabled true function to the top, then don't modify the others

forest:
Good idea. This for sure can be cleaned up. Just want to make sure this is
generally right approach.

lib/petal_components/link.ex
forest: I think we want both `button` and `a` since those render differently.

mplatts:
Cheers

mplatts:
Released in 1.0.4  https://github.com/petalframework/petal_components/releases/tag/v1.0.4

PR #150, opened 2023-02-27
feat: Add an attribute to show the modal on mounted

RyoWakabayashi:
I suggest adding an attribute that allows the modal to be hidden when mounted

if show_on_mounted is set to false, the modal is not displayed when mounted

modal

### DIFF
--- a/lib/petal_components/modal.ex
+++ b/lib/petal_components/modal.ex
@@ -19,6 +19,11 @@ defmodule PetalComponents.Modal do
     doc: "modal max width"
   )
 
+  attr(:show_on_mounted, :boolean,
+    default: false,
+    doc: "if show_on_mounted is set to false, the modal is not displayed when mounted"
+  )
+
   attr(:rest, :global)
   slot(:inner_block, required: false)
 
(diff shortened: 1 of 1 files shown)


lib/petal_components/modal.ex
```diff
@@ -19,6 +19,11 @@ defmodule PetalComponents.Modal do
     doc: "modal max width"
   )
 
+  attr(:show_on_mounted, :boolean,
+    default: false,
```
mplatts, line 23: Hmm this would mean everyone who upgrades will have their modals stop working? maybe we could reverse it - "hide_on_mounted", defaults to false?

lib/petal_components/modal.ex
```diff
@@ -64,6 +69,17 @@ defmodule PetalComponents.Modal do
     """
   end
 
+  def init_modal(show_on_mounted) do
+    unless show_on_mounted do
+      %JS{}
+      |> JS.remove_class("overflow-hidden", to: "body")
+      |> JS.hide(to: "#modal-overlay")
```
mplatts, line 76: just wondering why we need to hide so many things?

RyoWakabayashi: Certainly there was no need to hide so much.

I'll Add commits immediately.

nhobes:
Thanks for this, and apologies it sat unanswered for so long. The modal has been rewritten a couple of times since this was opened (it is now driven by LiveView.JS with a different show/hide flow), so this patch no longer applies to the current code. Closing it as part of a PR backlog cleanup. If you are still seeing the mount flash on a current version, please open an issue and we will get it sorted.

PR #153, opened 2023-02-27
feat: add conditional to JS.push event for slide over to accommodate live_co…

gsperka:
I am running into an issue where I created a `live_component` for the slide over,  however I am getting a handle_event error saying the "close_slide_over" event is being looked for in the parent, which I don't want. 

I am hoping for the same functionality achieved used with the modal component here: 

```
Close modal target
If you’re nesting a modal inside a live component, the close event will go to the parent live view, instead of the component.

To fix this, you can change the target to be the live component:

<.modal title="Modal" close_modal_target={@myself}>
  <.p>Content</.p>
</.modal>
Underneath the hood, this gets forwarded to the JS.push function. Here’s what it looks like:

# in petal_components/modal.ex
JS.push(js, "close_modal", target: close_modal_target)
```

### DIFF
--- a/lib/petal_components/slide_over.ex
+++ b/lib/petal_components/slide_over.ex
@@ -113,7 +113,12 @@ defmodule PetalComponents.SlideOver do
       },
       to: "#slide-over-content"
     )
-    |> JS.push("close_slide_over")
+
+    if close_slide_over_target do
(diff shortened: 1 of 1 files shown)


mplatts:
released in 1.0.4 https://github.com/petalframework/petal_components/releases/tag/v1.0.4

PR #158, opened 2023-03-07
Minor code readability improvement for Badge component

Rio517:
The private function `get_color_classes` had some fairly redundant pattern matching. Given that the badge component constrains the incoming attribute already, it seems safe to skip the pattern matching that was being performed in favor of simpiler, more readable code.

Of course, please let me know if I've misunderstood something.

### DIFF
--- a/lib/petal_components/badge.ex
+++ b/lib/petal_components/badge.ex
@@ -49,94 +49,16 @@ defmodule PetalComponents.Badge do
     end
   end
 
-  defp get_color_classes(%{color: "primary", variant: variant}) do
+  defp get_color_classes(%{color: color, variant: variant}) do
     case variant do
       "light" ->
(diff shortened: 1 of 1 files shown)


mplatts:
Thanks yeah we initially wrote it all out so the tailwind classes would be picked up (couldn't do interpolation) - but now with a stylesheet defined it doesn't matter. Cheers

Rio517:
Ah makes sense. It seems like there are many of these instances where those considerations drove the formatting, and now interpolation could be used.  Mind if I start submitting PRs to remove/simplify pattern matching?  Another example is the buttons component. 

Of course, don't let me stand in the way of progress. I would just do this when I have small gaps in other work.

mplatts:
PRs would be awesome! I’m working on something else atm

PR #172, opened 2023-04-19
Add tooltip to `icon_button`

fredwu:
An icon button sometimes isn't very useful as the user might recognise what it is or does. So let's allow for a tooltip.

This is what it looks like:
image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Unreleased
 
 - New: Add `separator_class` option to `<.breadcrumb>`
+- New: Add `tooltip` option to `<.icon_button>`
 
 ### 1.0.5 - 2023-04-03 09:33:31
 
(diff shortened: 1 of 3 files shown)


mplatts:
Nice, though how does this work? Doesn't it need javascript?

fredwu:
Oh oops, I was pushing ahead without thinking... this requires Flowbite. 🤦

fredwu:
I guess we'll close this PR, perhaps when petal component adds tooltips we can revisit.

fredwu:
Actual, hang on... I can make it work without Flowbite... Will update this PR.

fredwu:
Updated. Now it works with just Tailwind (but also works with Flowbite if it's there). :)

mplatts:
Thanks, can you add a test to `button_test.exs`?

fredwu:
Test added.

PR #175, opened 2023-04-22
Tooltip tweaks

fredwu:
- Add fade-in transition to make it nicer
- Remove the unneeded `label`
- Make it more portable by self-contain the `flex` class
- Make the grouping isolated so it can be nested in parent groups

### DIFF
--- a/assets/default.css
+++ b/assets/default.css
@@ -187,7 +187,7 @@
 }
 
 .pc-icon-button__tooltip {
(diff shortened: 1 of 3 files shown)


fredwu:
Also noticed we no longer need to supply the `label`, so removed it too.

fredwu:
Another update: the `flex` classes changes the behaviour of the `icon_button`. Let's add an inner container to contain the `flex` elements.

fredwu:
Also, I feel like the styling could use some tweaks. Though it's subjective so I haven't added it here.

Before:
image

After:
image

This is done by:

```css
.pc-icon-button__tooltip__text {
  @apply px-3 text-sm rounded-md;
}
```

mplatts:
🙏

PR #185, opened 2023-05-09
Fix radio group's state

fredwu:
So the radio group doesn't actually remember the state, so this PR fixes it.

However, there are currently no tests to test the form's state behaviour, I'm not sure what the best way to do it is. @mplatts @nhobes


mplatts:
I'll try write one

mplatts:
Done. Published in v1.1.6

PR #186, opened 2023-05-09
Make textarea v2's height the same as v1

fredwu:
Please feel free to close this PR, but I was hoping v2 and v1's default styling are consistent. So adding `rows="4"` to match v1's.


mplatts:
Yeah they should be consistent. Thanks!

mplatts:
Done in v1.1.6

PR #191, opened 2023-05-16
Move custom css classes to last so they can potentially override default ones

fredwu:
I haven't done a full scan of the code base for any other instances, just simply searched for "user_added_classes".


mplatts:
Smart

PR #190, opened 2023-05-16
Add `tooltip` to `<.button>`

fredwu:
Ported the same tooltip from `icon_button` to `button`.

PR #192, opened 2023-05-17
Add `tooltip` to `<.button>`

fredwu:
Second attempt of #190.

Text only:

image

With loading:

image

With icon:

image

With icon and loading:

image

### DIFF
--- a/assets/default.css
+++ b/assets/default.css
@@ -28,6 +28,22 @@
   @apply inline-flex items-center justify-center font-medium transition duration-150 ease-in-out border rounded-md focus:outline-none;
 }
 
+.pc-button__tooltip__group {
+  @apply relative flex flex-col items-center;
+}
+
+.pc-button__tooltip {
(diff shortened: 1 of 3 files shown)


mplatts:
Looking better! One thing I also missed last time is we don't put classes in the components anymore - we name classes using the BEM naming convention and put the CSS in `default.css`. This allows people to override it if they want. Couple of classes still in there if you want to have a go at a

fredwu:
Done. The only exception is that the group class needs to stay in the markup per Tailwind:

> Groups can be named however you like and don’t need to be configured in any way — just name your groups directly in your markup and Tailwind will automatically generate the necessary CSS.

https://tailwindcss.com/docs/hover-focus-and-other-states#differentiating-nested-groups

mplatts:
Didn't know that

fredwu:
Closing in favour of simply using https://atomiks.github.io/tippyjs/ as suggested by @mplatts .

mplatts:
( via https://petal.build/components/tooltip )

PR #201, opened 2023-05-31
Add `empty_message` attr for `checkbox-group` and `radio-group`

fredwu:
When there's no options it'll be great if we could add a message (different to error label).

### DIFF
--- a/assets/default.css
+++ b/assets/default.css
@@ -596,6 +596,10 @@
 .pc-radio-group__item--col {
   @apply inline-flex items-center gap-3 cursor-pointer;
 }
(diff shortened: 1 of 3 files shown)


mplatts:
Never thought of this.. What is the use case? Happy to include it as it shouldn't affect anyone

fredwu:
For my use case, I have m2m associations where the associated collection is user-generated content, so if the content is missing I'd like to display a message.

mplatts:
https://github.com/petalframework/petal_components/releases/tag/v1.2.9

PR #203, opened 2023-06-01
Add more details to upgrade guide

mindok:
It took a while to figure out the additional steps needed to complete an upgrade, so this PR includes some of the additional steps that were required.


mplatts:
🙏 Thanks @mindok !

Issue #224, opened 2023-07-11
Modal hide event is broken


sezaru:
The dialog will only work correctly if the user don't adds an id to the component.

That's because one of the JS events being pushed in the `hide_modal` function has `"#modal"` hardcoded where it should be `"##{id}"`

https://github.com/petalframework/petal_components/blob/453bffc2c832408fbe830e34caffe52ffc93589c/lib/petal_components/modal.ex#L93

nhobes:
@sezaru thanks for the fix!

https://github.com/petalframework/petal_components/pull/225

PR #225, opened 2023-07-11
fix: Fixes Modal hide_modal function when id is not default

sezaru:


sezaru:
This fixes issue https://github.com/petalframework/petal_components/issues/224

sezaru:
This also fixes the `phx-remove` call which was missing the `@close_modal_target`, meaning that the `@id` variable was being passed as the `close_modal_target` function parameter

PR #227, opened 2023-07-16
Fix bg colour for input checkbox

fredwu:
This PR fixes the bg colour of the checked checkboxes.

Dark mode works fine:

Screenshot 2023-07-16 at 5 30 24 pm

However, light mode currently looks like this, note that the first three boxes are checked, the last one is not, but they all look the same:

Screenshot 2023-07-16 at 5 30 14 pm

With this PR:

Screenshot 2023-07-16 at 5 29 57 pm

Took me a while to realise this was caused by styling. I was going crazy thinking my auto-checking plugin was broken. 🙈


fredwu:
cc @mplatts

mplatts:
this has been released btw..just forgot to add it to the changelog

Issue #229, opened 2023-07-20
`.pc-accordion-item--last` class applied to all but the last accordion item.


jpramassini:
I was working on modifying the existing [Accordion](https://petal.build/components/accordion) component (which is great, by the way) to have a fully rounded appearance instead of a rounded top and flat bottom. In doing so I was following the instructions outlined in the [Modifying Components page](https://petal.build/components/modifying-components) to update the `.pc-accordion-item--last` class to add a rounded bottom to the last accordion selector item, which is what it seems like this selector is meant to apply to. For reference, the companion selector `.pc-accordion-item__content-container--last` works for styling the content box in the same way as expected, only applying the style to the last content box in the accordion. What actually happens with the `accordion-item--last` class is that it styles all buttons that _aren't_ the last item instead. 

It seems like this is due to the `unless` here: https://github.com/petalframework/petal_components/blob/ffa946c92f9a1a51ca12505b43651543cdf8e176/lib/petal_components/accordion.ex#L59C17-L63C19

Happy to get a fix in for this, but before doing so just want to confirm if this is intended behavior or not. Thanks!

jpramassini:
After playing around with this a bit more locally, it seems like changing the behavior of this to try to do the rounding stuff (converting the unless to if, etc) doesn't do what I expected, so I think I probably just misunderstood the role this class. Closing, sorry!

mplatts:
Hmm it does seem wrongly named. Seems like it should be called `pc-accordion-item--all-except-last`

jpramassini:
Yeah, that name would probably make more sense given how it's applied. It would be good to change it to add clarity for people trying to do custom styling stuff, but I can also see how it could be a bit dangerous to rename if people are relying on that name for existing custom style overrides already.

nhobes:
@jpramassini you're right about the naming convention, certainly more appropriately named something like @mplatts suggested e.g `pc-accordion-item--all-except-last`. I can update that now, hopefully not too many people have modified that and will discover this conversation. The current component also doesn't make it easy to modify the last item's corners to rounded. Perhaps it could be an attribute even. A PR would be welcome if you end up solving that.

jpramassini:
Awesome, appreciate the name update! Not sure when I'll get to it, but if I end up messing around with it I'll definitely put up a PR 👍

PR #236, opened 2023-07-27
Add support for styling last accordion item

jpramassini:
Following up on Issue https://github.com/petalframework/petal_components/issues/229, here's what I came up with. This takes advantage of the new naming convention to allow for new styles that are intended for the last item only. As described in the issue, the main use case I had in mind when wanting this was rounding the bottom of the accordion to allow for a fully border-radius'd appearance instead of rounded on top and flat on bottom. The problem is that sometimes the bottom of the table is the last accordion button itself and sometimes it's the last accordion item's content box. (Guessing this is why the accordion was made flat on the bottom to begin with)

This PR doesn't explicitly add styles for rounding the table by default, but it adds logic to apply an always on class: `pc-accordion-item--last` (let me know if you want to make this a different name, I'm _slightly_ concerned about this conflicting with the old name for what is now `all-except-last`) as well as a conditional class: `pc-accordion-item--last--closed`, which is only applied when the last accordion item is closed. 

This can be used in conjunction with `pc-accordion-item__content-container--last` to style the bottom of the table to be rounded via style overrides and some `rounded-b-lg` on them both. I tried my best to follow naming conventions here, and apologies if this is a bit rough around the edges. There's also a small range of frames where this does look a _bit_ janky if you're using the smooth expand from Alpine. I don't think it's too bad, but any ideas on how to get around this would be great.

### DIFF
--- a/lib/petal_components/accordion.ex
+++ b/lib/petal_components/accordion.ex
@@ -59,7 +59,7 @@ defmodule PetalComponents.Accordion do
                 build_class([
                   "pc-accordion-item accordion-button",
                   if(i == 0, do: "pc-accordion-item--first"),
-                  unless(i == length(@item) - 1, do: "pc-accordion-item--last"),
+                  unless(i == length(@item) - 1, do: "pc-accordion-item--all-except-last"),
                   if(i == length(@item) - 1,
                     do:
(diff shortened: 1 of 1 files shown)


nhobes:
Thanks for this! Looks great!

PR #240, opened 2023-08-02
Allow custom slots in pagination component + format project using styler

RobinBoers:
Hi! I extended the pagination component to allow custom slots, like this:

```heex
<.pagination class="mb-5" current_page={@page} total_pages={@total_pages}>
  <:prev>
    <span phx-click="prev-page" class="pc-pagination__item__previous">
      <Heroicons.chevron_left solid class="pc-pagination__item__previous__chevron" />
    </span>
  </:prev>
  <:page :let={page}>
    <span phx-click="to-page" phx-value-page={page}>
      <%= page %>
    </span>
  </:page>
  <:next>
    <span phx-click="next-page" class="pc-pagination__item__next">
      <Heroicons.chevron_right solid class="pc-pc-pagination__item__next__chevron" />
    </span>
  </:next>
</.pagination>
```

It is completely optional. If the slots don't exist, they won't be used. I also added an test to test this change.

I removed the brackets in `attr`s for a lot of components too, and formatted the project using [styler](https://github.com/adobe/elixir-styler).

PR #241, opened 2023-08-02
 Add option to send events in pagination component

RobinBoers:
This PR adds the `event` flag to the pagination component. If enabled, the `link_type` and `path` options will be overwritten, and clicking the buttons in the component will send events to the parent LiveView. Here's an example:

```heex
<.pagination event={true} current_page={@page} total_pages={@total_pages} />
```

```elixir
@impl true
def handle_event("goto-page", %{"page" => page}, socket) do
  {:noreply, push_patch(socket, to: ~p"/posts/#{page}")}
end
```

This PR is a replacement for #240


nhobes:
Released in 1.4.2!

PR #244, opened 2023-08-08
Fix: Icon button setting info to primary

Cameron-Kurth:
Description
---

The info color arg for icon buttons was being set to the primary class. This PR just makes this small fix and adds some test coverage to ensure the colors are set correctly.


nhobes:
Thanks for this! Merged.

PR #246, opened 2023-08-11
Add `type=button` to close slide over button

tegon:
Ideally `button` elements should always specify a `type` attribute to avoid unexpected behavior, since each browser might implement a different default.

In our app, we do HTML forms inside slide overs. When the user clicks to close a slide over, the form ends up being submitted, given that the close button is inside the form. Setting the button's type prevents this issue.


nhobes:
Thanks for this!

PR #249, opened 2023-08-16
Add support for disabled attribute on dropdown menu items

RobinBoers:
Hi! I added support for the `disabled` attribute on the `dropdown_menu_item` component.

I don't know whether this is the correct way to apply styling. I tried to mirror the approach that the `button` component uses.

Also, sorry for the giant diff! I removed the annoying brackets the formatter adds everywhere. Btw, that has been fixed a while ago in ElixirLS, see elixir-lsp/elixir-ls#526.

### DIFF
--- a/assets/default.css
+++ b/assets/default.css
@@ -496,7 +496,7 @@
 :not(.phx-no-feedback).pc-form-field-wrapper--error textarea {
   @apply border-danger-500 focus:border-danger-500 text-danger-900 placeholder-danger-700 bg-danger-50 dark:text-danger-100 dark:placeholder-danger-300 dark:bg-danger-900 focus:ring-danger-500;
 }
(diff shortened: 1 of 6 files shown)


mplatts:
Thanks - looks good. Yeah we used string interpolation elsewhere in the project, although now I think the best way is to use a list.. eg `"#{@class} pc-dropdown__menu-item #{get_disabled_classes(@disabled)}"` => `[@class, "pc-dropdown__menu-item", get_disabled_classes(@disabled)`.

Is it possible to add a test in `dropdown_test.exs`?

RobinBoers:
On it!

RobinBoers:
So the `Link.a` component that's used under-the-hood expected a `:string` in the attrs, so I changed that to `:any`, since I don't think the syntax supports union types.

mplatts:
🙏

nhobes:
Released in 1.4.6. Thanks for this!

PR #253, opened 2023-08-23
Add 'close_on_click_away' attr on modal

dvic:


fredwu:
Nice! I think we should also add a similar option for the escape key. :)

mplatts:
Released: https://github.com/petalframework/petal_components/releases/tag/v1.4.8

@fredwu I also added `close_on_escape`

PR #257, opened 2023-09-02
Add `target` attribute for sending pagination events to LiveComponents

RobinBoers:
In #241 I added an option to make the pagination component send events instead of using `live_patch`/`live_redirect`. 

This PR adds an optional `target` option, that can be used to set the `phx-target` for those events, which makes it also work in LiveComponents :)


nhobes:
Thanks Robin!

Issue #254, opened 2023-08-25
Add `disabled` attribute to `Tabs` `.tab` component


jpramassini:
I have a situation where I'd like to be able to selectively disable access to specific tabs. This requires disabling clicks (or at least linking to the underlying route the tab is connected to). Conveniently, the underlying `Link` component supports this attribute already and passing the `disabled` attribute through the `.tab` components already works perfectly, but raises warnings about the `disabled` attribute being undefined. Adding this attribute to the tabs themselves would allow for some nice free added functionality without wrestling with attribute warnings. 

Happy to put up a PR for this, just wanted to make sure that was something that made sense to you before doing it.

mplatts:
Hey yep a PR would be accepted for this!

PR #258, opened 2023-09-04
Add `disabled` attribute to Tab component

jpramassini:
Sorry for the delay on this, following up on the quick change referenced in https://github.com/petalframework/petal_components/issues/254. Passes through a `disabled` attribute to the `Link` component for `Tab`s.

(I also fixed a small doc typo I noticed while I was in there for the underline attribute.)


mplatts:
Cheers. Released in 1.5.1

PR #264, opened 2023-09-29
Add pagination attribute allowing always shown prev and next buttons

### LINKED ISSUE #140: Request: keep pagination previous & next buttons always visible
Currently the pagination component's previous and next page buttons get hidden from the DOM when you're on page 1 or the last page. This then causes the element to jump around a bit, which degrades the UX a little bit.

Instead of hiding the buttons completely, they could be grayed out & disabled.

angelasheppard:
Adds a boolean attribute to the pagination component allowing the previous & next buttons to always be shown on the boundary pages. Defaults to current behavior (prev not shown on page 1, next not shown on last page).

Buttons shown on those boundary pages are disabled and reduced opacity.

[Loom of the live boilerplate page](https://www.loom.com/share/15f0d94008874bc58ae7cc007fe41a7a?sid=59d436c3-537c-4c4f-b933-fade87bb7dfe)

Resolves #140

### DIFF
--- a/assets/default.css
+++ b/assets/default.css
@@ -843,13 +843,13 @@
   @apply inline-flex items-center justify-center leading-5 px-3.5 py-2 border border-gray-200 dark:border-gray-700;
 }
 .pc-pagination__item__previous {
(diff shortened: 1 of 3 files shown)


nhobes:
This is great, thanks so much for contributing! 🙌

PR #268, opened 2023-10-13
Add 'close on' attributes to slideover

samuelpordeus:
Add the `close_on_click_away` and `close_on_escape` attributes to the `slideover` so it matches the `modal`

I was chatting with @mplatts about adding the `close_on_click_away` attr yesterday, but since we are here i also added the `close_on_escape` as well giving more flexibility to the component 🙂

lmk if i'm missing anything ❤️


mplatts:
released in 1.7.0

mplatts:
also added fade in transition

PR #283, opened 2023-12-16
Update modal component to fix duplicate `close_modal` event

axelclark:
Problem:  With the current implementation, the `close_modal` event gets sent twice to the LiveView if you `push_patch` from the `close_modal` `handle_event` in the LiveView.  First when `phx-click` calls `hide_modal/2` and then again when `phx-remove` is executed after the `push_patch` removes the modal from the DOM.

Solution:  Calls [JS.push/2](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.JS.html#push/2) before executing the function in `phx-remove` and removes the `JS.push/2` from the `phx-remove` function.  This update fixes the bug where `close_modal` event fires twice with server `push_patch`.

Also, adds `on_cancel` attr with default to existing `JS.push/2` functionality so library users have the options of passing [JS.patch/1](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.JS.html#patch/1) to avoid the round trip through the `handle_event` or `%JS{}` if they want to simply close the modal and remain on the current page without sending the `close_modal` event.

This PR borrows patterns from the latest `modal` component generated in a new Phoenix project, see [here](https://github.com/phoenixframework/phoenix/blob/a66e7c33019a3d564c4f7f2e48b77efb63c54ad7/priv/templates/phx.gen.live/core_components.ex#L39-L90).

### DIFF
--- a/lib/petal_components/modal.ex
+++ b/lib/petal_components/modal.ex
@@ -24,6 +24,11 @@ defmodule PetalComponents.Modal do
     default: false,
     doc: "whether or not the modal should have a close button in the header"
 
+  attr :on_cancel, JS,
+    default: JS.exec("data-cancel-default"),
+    doc:
+      "a JS function to execute when the modal is closed. Defaults to pushing close_modal event"
+
   attr :max_width, :string,
     default: "md",
     values: ["sm", "md", "lg", "xl", "2xl", "full"],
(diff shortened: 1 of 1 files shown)


mplatts:
Many thanks!

PR #295, opened 2024-01-30
Fix modal class

periclesd:
Currently, the `class` passed into the modal component is part of the `rest` global attr, since it's not explicitly declared, which causes `custom_classes` in `get_classes/1` to always be `""`.


nhobes:
Thanks, released in [1.9.1](https://github.com/petalframework/petal_components/releases/tag/v1.9.1)!

periclesd:
Thanks for doing an awesome job!

PR #298, opened 2024-02-12
Put id attr on <table> when not using :col slots

soundmonster:
The `id` attribute on a table is only used on `<tbody>` when `:col` slots are provided. The attribute goes unused if no
`:col` slots are used. This proposal puts the `id` on the `<table>` element in case column slots are not used. An
alternative can be to put the `id` on the `<table>` unconditionally, but the `<tbody>` needs an `id` as well, so we'd
have to "invent" one, e.g. `id={@id <> "_tbody"}`, not sure if this is desirable. What are your thoughts on this?


mplatts:
Hm yeah I think put on table unconditionally and use `id={@id <> "_tbody"}`. Can you update to this?

PR #311, opened 2024-03-17
Fix table id

fredwu:
Recent changes to the `<.table>` component broke the `id` attribute.


mplatts:
Thanks the ID was just moved to tbody (but still not a good solution). See https://github.com/petalframework/petal_components/pull/298

PR #326, opened 2024-07-01
Make Petal Components Accessible

tylerbarker:
This gets us a pretty nice starting a11y test we can run against all components rendered in `a11y_live.ex`. It uses `:phoenix_playground` to make this work without needing to include an entire LiveView application just for testing. This caught 3-4 existing a11y errors, but there may be things we can do to improve the test in future PRs.


tylerbarker:
This is passing now. I'm a little skeptical I've covered all the components sufficiently in the test, however I think it'd be good to get this merged as a first step before exploring further. 

I'm just waiting until [a contribution](https://github.com/phoenix-playground/phoenix_playground/pull/10) I made to `:phoenix_playground` is released so I can bump the version and remove a redundant `<script>` - then we're good.

PR #329, opened 2024-08-02
Cleanup type of class attrs

nallwhy:
`class` can take string and list of string also.


lib/petal_components/pagination.ex
```diff
@@ -43,7 +43,7 @@ defmodule PetalComponents.Pagination do
 
   def pagination(assigns) do
     ~H"""
-    <div {@rest} class={"#{@class} pc-pagination"}>
+    <div {@rest} class={["pc-pagination", @class]}>
```
nallwhy, line 46: This kind of problems exist in petal pro also.

mitkins:
Thanks @nallwhy!

PR #331, opened 2024-08-08
a11y improvements

tylerbarker:
Here I've gone over the component set and added additional aria tags and roles where appropriate. There was pretty decent coverage of this already, so I've only made changes to 9 components.

I'm not really a web accessibility expert, so I'd be hesitant to say these changes are _all_ that is necessary to be 100% accessible. That said, I think these changes ought to help a lot, and we should consider this effort an ongoing one to uphold a good standard.


tylerbarker:
The a11y is flaky ATM, the SlideOver issues it's reporting in the latest CI run aren't always happening locally 🧐

PR #333, opened 2024-08-15
Remove `for` attribute in `radio-group` label

tegon:
The `for` attribute in `label` is supposed to match the `id` of a visible `input` element on the page. 
Since the `radio-group` component renders multiple `radio` inputs, it's not possible to have the label match a single element with a unique ID.

Having the `label` unmatched with an `input` causes an accessibility issue to show up on the Chrome dev tools.

Screenshot 2024-08-14 at 15 58 09


mitkins:
Nice

PR #334, opened 2024-08-22
Hide dropdown when pressing escape

RobinBoers:
I added a `phx-` attribute for hiding dropdown menus when pressing the escape key.


mitkins:
Thanks @RobinBoers!

PR #335, opened 2024-08-25
Add default_js_lib config

nallwhy:
This PR adds `default_js_lib/0` to address the inconvenience of having to set `js_lib` for the component every time it is used.


mitkins:
Awesome. Thanks @nallwhy

Issue #328, opened 2024-07-08
Problem with loading of heroicons


rakso-startkoden:
Due to the slight difference in how heroicons are being loaded compared to how phoenix 1.7 suggest I'm having problems using petal_components with other libraries that also require heroicons.

Pheonix default https://github.com/phoenixframework/phoenix/blob/main/installer/templates/phx_single/mix.exs
suggest doing this. 
```
{:heroicons,
       github: "tailwindlabs/heroicons",
       tag: "v2.1.1",
       sparse: "optimized",
       app: false,
       compile: false,
       depth: 1}
       
```
And a recent update in the backpex library changes it to do the same https://github.com/naymspace/backpex/blob/develop/guides/get_started/installation.md#track-the-heroicons-github-repository.

This now causes issues for me becase If i add the heroicons in a similar way as petal_components suggest
```
{:heroicons, "~> 0.5.3"}
```
 it does not work for reasons I can't understand. 
And if i override the phoenix default way it causes petal_component to crash on start.
 
 Is there anyone that have similar issue, or is there a way to configure this so that both libraries work together?

CharlesIrvineKC:
Yes, same problem here.

mitkins:
I'm in the process of investigating this.

The dependency that's generated by Phoenix pulls directly from the `tailwindlabs/heroicons` GitHub repo. It just happens to have the same name as the [heroicons package](https://hex.pm/packages/heroicons).

A workaround for now is to rename the dependency created by Phoenix. E.g.:

```elixir
{:core_heroicons,
  github: "tailwindlabs/heroicons",
  tag: "v2.1.5",
  app: false,
  compile: false,
  sparse: "optimized"},
```

To keep the above dependency functional, you'll need to update `assets/tailwind.config.js`:

```javascript
plugin(function({matchComponents, theme}) {
  let iconsDir = path.join(__dirname, "../deps/core_heroicons/optimized")
  ...
}
```

Finally, the core components `<.icon />` function component will be overridden by the Petal Component version (depending on where you put `use PetalComponents`). If you want to use the core components version, you might have to rename the function in `lib/hello_web/components/core_components.ex`:

```elixir
  def core_icon(%{name: "hero-" <> _} = assigns) do
    ~H"""
    <span class={[@name, @class]} />
    """
  end
```

Then the dependencies won't clash and the `:heroicons` in the Petal Component library will continue to work as they always have. Of course, this means that there will be a double up of heroicons - but only if you use both versions.

[...]

PR #340, opened 2024-09-06
Aligns .icon with latest heroicon method

mitkins:
This PR aims to address https://github.com/petalframework/petal_components/issues/328

In the current version of Petal Components, heroicons are rendered using the [Heroicons](https://github.com/mveytsman/heroicons_elixir) package:

```elixir
  {:heroicons, "~> 0.5.3"}
```

Recently, an alternative solution was implemented in the Phoenix Framework. See the discussion in the following issue:

  - https://github.com/phoenixframework/phoenix/issues/5622

In the latest version of Pheonix, if you generate a new project, it'll add heroicons like so:

```elixir
  {:heroicons,
    github: "tailwindlabs/heroicons",
    tag: "v2.1.5",
    app: false,
    compile: false,
    sparse: "optimized"},
```

Even though both dependencies have the same name (`:heroicons`) - they are not referring to the same dependency package (causing a dependency name clash). The heroicons solution introduced by the Phoenix Framework introduces some new benefits (i.e. reduced deployment size and client-side caching). This PR aligns Petal Components to the new method.

### Differences between the old and new method

Before now, you would render a heroicon like so:

```heex
<Heroicons.chevron_down solid />
<PetalComponents.Icon.icon name={chevron_down} solid />
```

With the new method, you render a heroicon like this:

```heex
<PetalComponents.Icon.icon name="hero-chevron-down-solid" />
```

The main difference is that you would refer to the heroicon as a string (instead of an atom) using the "hero-" prefix. This is important, because in the new solution, the heroicons are generated as a set of CSS classes (with an inline svg image). If you don't refer to the CSS class name, then it willl be trimmed from the `app.css` file by the Tailwind JIT.

Another benefit of implementing heroicons via CSS is that the resulting html moves from server-side rendering to client-side. Using the old method, if you used the same icon twice, it would be twice the download. Using CSS, the browser can cache the data client-side.

You can refer to heroicon variations with different class names. E.g:

```
hero-home
hero-home-solid
hero-home-mini
hero-home-micro
```

This adds greater flexibility to components like the Button. Take the following example from the documentation:

```heex
<.button icon={:home} label="Mini" />
<.button with_icon link_type="a" to="/">
  <Heroicons.home solid class="w-5 h-5" /> Solid
</.button>
<.button with_icon link_type="a" to="/">
  <Heroicons.home class="w-5 h-5" /> Outline
</.button>
```

In the past, if you didn't use `with_icon`, the `.button` component would assume that you're generating a mini icon. Here's what you can do now:

```heex
<.button icon="hero-home-mini" label="Mini" />
<.button icon="hero-home-solid" link_type="a" to="/" />
<.button icon="hero-home" link_type="a" to="/" />
```

Note that the first line refers to "hero-home-mini". This is the first problem that's introduced by the new method - it's not obvious you should replace `:home` with `hero-home-mini` to keep the look of the Button consistent (with the past). 

Another problem that's not so obvious is that if you want to control the size of the icon, you'll need to use the `with_icon` method. This problem is particularly confusing with the `.icon_button`. You control the size of the container with the `size` attribute, but the size of the icon is controlled by the content. So for example, here's a small Icon Button using the old method:

```heex
<.icon_button tooltip="Clock" size="sm" color="secondary">
  <Heroicons.clock solid />
</.icon_button>
```

And here's what I had to do to make it work with the new method:

```heex
<.icon_button tooltip="Clock" size="sm" color="secondary">
  <.icon name="hero-clock-solid" class="h-6 w-6" />
</.icon_button>
```

I'm not sure that the addition of the `class` is obvious (particularly when upgrading).

One final problem, when generating a new Phoenix project, code is added to `tailwind.config.js` to generate the heroicon CSS. If users are upgrading an existing project then they may need to add this code manually.

### The path forward

I think in most cases, issues can be addressed with appropriate documentation. For example:

* The documentation can contain instructions on how to add the JavaScript code to `tailwind.config.js`
* It can provide a note that an upgrade process is necessary and the little gotchas (mentioned above)
* Copy and paste examples will be updated in the case where the user is creating a new interface

While on the topic of upgrading, I did consider creating "Heroicons V3" - but that would only be possible if we could keep the original `:heroicons` dependency. Unfortunately, I think the best way forward is to help users through the upgrade process via the [documentation](https://petal.build/components/heroicons).


lib/petal_components/menu.ex
```diff
@@ -287,9 +285,13 @@ defmodule PetalComponents.Menu do
       ) %>
     <% end %>
 
-    <%= if is_binary(@icon) do %>
+    <%= if is_binary(@icon) && String.match?(@icon, ~r/svg|img/) do %>
```
mitkins, line 288: Because the heroicon name is no longer an atom and it is now a class name (string), I updated this comparison. if this fails, then it falls through to the next `if` statement (where it collects all strings). 

In other words, if it's not an `svg` or an `img` it will be considered to be a heroicon class name.  Any thoughts?

assets/default.css
```diff
@@ -1217,7 +1217,8 @@
 }
 
 .pc-vertical-menu-item__toggle-chevron__icon {
-  @apply w-3 h-3 ml-2 transition duration-200 transform;
+  /* !important is needed because heroicon class overrides width and height */
+  @apply w-3 h-3 ml-2 transition duration-200 transform !important;
```
mitkins, line 1223: This was a little tricky. this is what the icon looks like in the template:

https://github.com/petalframework/petal_components/blob/01a3bbbac194cb456162c119ed40efb7554d30c1/lib/petal_components/menu.ex#L257-L261

And here's how it looks rendered:

```html
<span class="hero-chevron-right pc-vertical-menu-item__toggle-chevron__icon" ... />
```

Because of CSS specificity, the `hero-chevron-right` class overrides what has been setup in `pc-vertical-menu-item__toggle-chevron__icon` class:

```css
@import "tailwindcss/base";
@import "../../deps/petal_components/assets/default.css"; /* <-- Petal Components are added here */
@import "tailwindcss/components"; /* <-- Heroicons are generated by default in here */
@import "tailwindcss/utilities";
``` 

Effectively the width/height is unintentionally mangled

It seems that the simplest solution is to add `!important` to the `pc-vertical-menu-item__toggle-chevron__icon` class. Although adding `!important` is undesirable, I think it encourages the end user to alter the `default.css` file. Since it's fairly unlikely that the user will want to change this part of the menu component, then this seems like an okay compromise.

Any objections?

lib/petal_components/icon.ex
```diff
@@ -1,23 +1,58 @@
...
+
+    attr :name, :any, required: true
+    attr :class, :any, default: nil, doc: "svg class"
+  end
 
-      <.icon name={:arrow_right} class="w-5 h-5 text-gray-600 dark:text-gray-400" />
-  """
-  def icon(assigns) do
-    apply(Heroicons, assigns.name, [assigns])
+  for heroicon_name <- heroicon_names do
```
mplatts, line 55: Could you loop over twice and do it for the atoms? That way we could still use atoms? Or not worth it?

mplatts: (For backwards compatibility)

mitkins: Unfortunately, I don't think so. You need to use the heroicon class in the calling code - so that Tailwind JIT doesn't remove that class. If you mapped every atom to a heroicon class name, then no heroicon class would be trimmed.

PR #341, opened 2024-09-08
Use Phoenix.Components.used_input?/1 

joepstender:
Use Phoenix.Components.used_input?/1 to only render errors when the form fields are used. 

See https://hexdocs.pm/phoenix_live_view/1.0.0-rc.6/Phoenix.Component.html#used_input?/1

The commit is not backwards compatible with earlier versions of live_view. Any suggestions for this?


mitkins:
Petal Components will be upgraded to 1.0 when it's released. We could merge this PR at that point in time. Is that what you're getting at?

joepstender:
I'm just worried about the backwards compatibility. It's up to what the maintainers want to support. You could do something like (not checked):
```elixir
  if Code.ensure_loaded?(Phoenix.Component) and function_exported?(Phoenix.Component, :used_input?, 1) do
    defp field_has_errors?(%{form: form, field: field}) when is_map(form) do
      case Keyword.get_values(form.errors || [], field) do
        [] -> false
        _ -> Phoenix.Component.used_input?(form[field])
      end
    end
  else
    defp field_has_errors?(%{form: form, field: field}) when is_map(form) do
      case Keyword.get_values(form.errors || [], field) do
        [] -> false
        _ -> true
      end
    end
  end

  defp field_has_errors?(_), do: false
```
Or just clearly state it won't work with earlier versions of live_view, which would force people to update.

mitkins:
I think it makes sense that once Live View 1.0 is released, it'll become the base line for future development (which would include `used_input?/1`). Once there is a release that refers to Live View 1.0, then users will either need to upgrade their project to Live View 1.0 or use an older version of Petal Components. 

When the time comes, we'll update the change log and the upgrade notes (with regards to the change to Live View 1.0). However, if this causes a pain point for people, then we can we can certainly re-evaluate what our options are at that point.

mitkins:
I'm just going to close this for now. If you want to add anything or re-raise the issue around the release for 1.0, feel free to do so

nallwhy:
@mitkins @joepstender Now that Phoenix LiveView 1.0 has been officially released, I think it’s a good time to revisit it.

joepstender:
I will see if I can make a PR.

nhobes:
PR would be welcome! I've reopened this for now.

mitkins:
I'm going through Petal Components right now - with regards to `phx-feedback-for` and `used_input?/1`

joepstender:
I updated my PR, don't know how to fix the mix.lock conflict?

mitkins:
Hi @joepstender. I've used your updates in the related PR to update Forms V1 and V2 to LiveView 1.0

PR #347, opened 2024-09-19
Ensure css transitions don't cause flaky tests

tylerbarker:
The a11y tests have been failing intermittently. @nhobes found [the culprit](https://github.com/angelikatyborska/a11y-audit-elixir/commit/b56c1439c680276c1e4972a9c4e6d879eec836b0). This implements the fix outlined there.


mitkins:
Excellent!

PR #350, opened 2024-09-24
Add ID to table row containing empty state slot

dgigafox:
Running tests on an app that passes stream to the table component produces the following error:
```
** (ArgumentError) setting phx-update to "stream" requires setting an ID on each child. No ID was found on:

     <tr class="pc-table__tr hidden only:table-row"><td class="pc-table__td " colspan="2">
```
because the row that contains the empty state slot has no ID.

Based on https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#stream/4-handling-the-empty-case, the row with the empty state should also have an ID.

Versions used:
elixir          1.17.2-otp-27
phoenix_live_view          0.20.14

### DIFF
--- a/lib/petal_components/table.ex
+++ b/lib/petal_components/table.ex
@@ -55,7 +55,7 @@ defmodule PetalComponents.Table do
         </thead>
         <tbody id={@id} phx-update={match?(%Phoenix.LiveView.LiveStream{}, @rows) && "stream"}>
           <%= if length(@empty_state) > 0 do %>
-            <.tr class="hidden only:table-row">
+            <.tr id={@id <> "-empty"} class="hidden only:table-row">
               <.td
                 :for={empty_state <- @empty_state}
                 colspan={length(@col)}


mitkins:
Just to clarify. Is this a test that's generated using `phx.gen.live`? I'd like to reproduce the circumstance that other people might find themselves in (if that's possible)

dgigafox:
> Just to clarify. Is this a test that's generated using `phx.gen.live`? I'd like to reproduce the circumstance that other people might find themselves in (if that's possible)

@mitkins nope. It's a live view page test on our project using `phoenix_live_view 0.20.14`. Do you want me to create a sample project if that would help?

mitkins:
`phx.gen.live` does use streams, so I thought maybe you had uncovered something I had missed (I've been working on the generators for Petal Pro). I assume that your test is just checking each child element for an id?

dgigafox:
@mitkins correct me if I'm wrong I'm using 2.0.6 and did try to use the generator, example:
`mix petal.gen.live Blog Post posts title description:text published_at:datetime published:boolean`

And it assigns a post like this:
```
defp assign_posts(socket, params) do
    starting_query = Post
    {posts, meta} = MyApp.DataTable.search(starting_query, params, @data_table_opts)
    assign(socket, posts: posts, meta: meta)
  end
```
So I'm guessing it is not using streams. Plus the generator is using the data_table. But maybe I'm wrong I could be using an outdated generator?

Anyway, one way or another I still use the latest live_view and petal_components versions and still get the same error when running a test with a table/1 components and streams.

mitkins:
Ah, the generator has been updated in Petal Pro `main`. It hasn't been released (but will be soon). 

The default generator will create a `Table` (with streams). You can choose to add the Data Table with the `--data-table` option. 

Thanks for the reply - I think that covers everything

PR #366, opened 2024-10-29
Fix slide-over wrapper/overlay not hiding on close

dshoreman:
When closing the slide over component, the main wrapper's `display: block` rule from `JS.show` doesn't get removed, leading to underlying content being blocked.

This PR fixes it by also hiding the wrapper at the other end of the chain.


mitkins:
Thanks for the fix. Nice find!

PR #368, opened 2024-11-04
Address Button Group warnings

### LINKED ISSUE #367: Button group usage throws compiler errors
I initially reported this [here](https://github.com/phoenixframework/phoenix_live_view/issues/3487) as I wasn't sure it was an expected use case that was missed, please see that issue for more details.

As the `button` slot in button groups has a `do` block ([here](https://github.com/petalframework/petal_components/blob/main/lib/petal_components/button_group.ex#L64)) the `validate_attrs` opt is ignored, so using attrs like `phx-click` throws errors - you can see this in LiveView 0.20.17+, despite the docs for button group saying that fixed it.

mitkins:
Resolves #367

### DIFF
--- a/lib/petal_components/button_group.ex
+++ b/lib/petal_components/button_group.ex
@@ -61,17 +61,20 @@ defmodule PetalComponents.ButtonGroup do
     doc: "class to customize the button border styles"
 
   # We mark validate_attrs false so passing phx-click etc. does not emit warnings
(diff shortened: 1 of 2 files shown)


nhobes:
Thanks! Is this merge-ready?

mitkins:
So yes it is :)

PR #411, opened 2025-03-03
prepares css file for tw4

IdoLeshkowitz:
This pr aligns the css file with the new tw4 requirements - deprecated design tokens were removed and the `@utility` directive was added to the existing rules.


mitkins:
Thanks for taking the time to do this. I learned a few things! I have a different branch where the classes are added to the component layer (instead of the utility layer):

https://github.com/petalframework/petal_components/tree/tailwind_4

The main reason I did this is because of this statement in the Tailwind 4 documentation:

https://tailwindcss.com/docs/adding-custom-styles#adding-component-classes

> By defining component classes in the components layer, you can still use utility classes to override them when necessary:
> ```HTML
> 
> 
> <div class="card rounded-none">
>   
> </div>
> ```

I like the idea of using the `@utility` directive, but I think adding them in the `components` layer will give greater flexibility. I was just wondering, am I missing something?

IdoLeshkowitz:
> Thanks for taking the time to do this. I learned a few things! I have a different branch where the classes are added to the component layer (instead of the utility layer):
> 
> https://github.com/petalframework/petal_components/tree/tailwind_4
> 
> The main reason I did this is because of this statement in the Tailwind 4 documentation:
> 
> https://tailwindcss.com/docs/adding-custom-styles#adding-component-classes
> 
> > By defining component classes in the components layer, you can still use utility classes to override them when necessary:
> > ```
> > 
> > <div class="card rounded-none">
> >   
> > </div>
> > ```
> 
> I like the idea of using the `@utility` directive, but I think adding them in the `components` layer will give greater flexibility. I was just wondering, am I missing something?

that makes sense, want me to close this one?

PR #422, opened 2025-04-03
Bubble event when input is cleared

szajbus:
**Current behavior**

Clearable form field does not trigger form's `phx-change` when clicking the clear button and the LiveView doesn't receive the updated form state.

**Expected behavior**

`phx-change` is triggered on the form just as it is when typing in something in the input.

**Fix**

Updated the clear button logic to dispatch the 'input' event with `{ bubbles: true }`, allowing the event to propagate up to the form and properly trigger `phx-change`.


nhobes:
Thanks so much for this!

PR #423, opened 2025-04-08
Add: ChatBubble and test

GraphiteSprite:
Added ChatBubble based on example at https://flowbite.com/docs/components/chat-bubble/ minus dropdown menu. Awaiting feedback.

### DIFF
--- a/assets/default.css
+++ b/assets/default.css
@@ -1051,6 +1051,24 @@
   .pc-checkbox-group__item--col {
     @apply inline-flex items-center gap-3 cursor-pointer;
   }
+  .pc-radio {
+    @apply w-4 h-4 border-gray-300 cursor-pointer text-primary-600 focus:ring-primary-500 dark:bg-gray-800 dark:border-gray-600;
+  }
+  .pc-radio-group {
+    @apply pt-2;
+  }
+  .pc-radio-group--row {
+    @apply flex flex-row gap-4 mb-3;
+  }
+  .pc-radio-group--col {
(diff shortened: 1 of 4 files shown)


nhobes:
Thanks for this! I think you may still need to push up the css for the `default.css` file?

GraphiteSprite:
Ah! I missed that part 🙏 Will review 😎

nhobes:
Hey @GraphiteSprite, really sorry this one got left hanging for so long - that is on us, and thanks again for the work you put in here and for following up on the CSS feedback at the time.

Closing it now as part of a backlog cleanup, but I want to explain the thinking rather than just hit the button. Since this was opened the library has moved to Tailwind v4 with a new theming and typography system, so the implementation would need a ground-up redo against the current tokens. More fundamentally, we are curating the component set around a shadcn-style core, and a chat bubble sits closer to app-level UI that we would rather see composed from primitives like avatar and card than shipped as a large component we maintain forever. If demand for a chat pattern shows up we will revisit it against the current styling.

Genuinely appreciate the contribution and hope to see you around the repo.

GraphiteSprite:
Thanks @nhobes good to know what happened to it! It was @mplatts suggestion to try the chat bubble as a learning exercise so I think it did help me get more familiar with the structure you're working with and some key concepts but maybe got stalled on momentum! If you have any other current issues that would be suitable for beginners I'd be happy to take a look if that would be any help :pray: :heart:

PR #491, opened 2026-07-05
Fix basic inputs dropping value, placeholder and other global attrs

### LINKED ISSUE #488: Regression: search_input (and other render_basic_input inputs) ignore value, placeholder, and other @form_attrs globals since 3.0.2
### Summary

`PetalComponents.Form.search_input/1` (and several sibling inputs) no longer forward globals such as `value` and `placeholder` to the rendered `<input>`, although they did in **3.0.1**. This breaks common LiveView patterns (pre-filled search fields, `phx-update="ignore"` on inputs, etc.).

### Affected versions

- **Broken:** 3.0.2+, verified on 3.2.2 and 4.0.6
- **Last known good:** 3.0.1

### Regression commit

https://github.com/petalframework/petal_components/commit/d2f957c1f5df10b8860f199aad7a4a292841ae47 ("Refactor form.ex")

### Root cause

`render_basic_input/2` correctly merges `assigns[:rest]` into `Form.*_input/3` options, but Phoenix only populates `:rest` when the **immediately preceding** function declares `attr(:rest, :global, include: @form_attrs)`.

After the refactor, only `text_input/1` has that `attr` before a `render_basic_input` caller. `search_input/1`, `email_input/1`, `number_input/1`, `password_input/1`, `telephone_input/1`, `url_input/1`, `datetime_local_input/1`, and `date_input/1` do not — so `@rest` is `%{}` and attributes like `value` are dropped.

In 3.0.1, each of those functions had its own `attr(:rest, :global, include: @form_attrs)` block (see tag `v3.0.1`).

### Steps to reproduce

1. Checkout `v3.2.2` (or current `main`).
2. Add the test below to `test/petal/form_test.exs` (same style as existing `"text_input"` test).
3. Run `mix test test/petal/form_test.exs`.

```elixir
test "search_input forwards value and placeholder" do
  assigns = %{}

  html =
    rendered_to_string(~H"""
    <.form :let={f} for={%{}} as={:user}>
      <.search_input
        form={f}
        field={:name}
        value="abc-123"
        placeholder="eg. Jane Doe"
      />
    </.form>
    """)

  assert html =~ ~s(value="abc-123")
  assert html =~ ~s(placeholder="eg. Jane Doe")
end
```

### Expected behavior

HTML includes `value="abc-123"` and `placeholder="eg. Jane Doe"` (as in 3.0.1).

### Actual behavior

Input is rendered with `type="search"` and correct `name`, but without `value` / `placeholder`.

### Test output

<details>
<summary>v3.2.2 — FAILS</summary>

```
Running ExUnit with seed: 587516, max_cases: 24
Excluding tags: [:test]
Including tags: [location: {"test/petal/form_test.exs", [654, 673]}]

  1) test search_input forwards value and placeholder (PetalComponents.FormTest)
     test/petal/form_test.exs:654
     Assertion with =~ failed
     code:  assert html =~ ~s(value="abc-123")
     left:  "<form>\n  \n  \n  \n  <input class=\" pc-text-input\" id=\"user_name\" name=\"user[name]\" type=\"search\">\n\n</form>"
     right: "value=\"abc-123\""
     stacktrace:
       test/petal/form_test.exs:669: (test)

.
Finished in 0.3 seconds (0.00s async, 0.3s sync)
37 tests, 1 failure, 35 excluded
```

Note: the second test (`text_input still forwards value`) passes, proving the issue is specific to inputs missing `attr(:rest, ...)`.

</details>

<details>
<summary>v3.0.1 — PASSES</summary>

```
Running ExUnit with seed: 897687, max_cases: 24
Excluding tags: [:test]
Including tags: [location: {"test/petal/form_test.exs", [644, 663]}]

..
Finished in 0.3 seconds (0.00s async, 0.3s sync)
37 tests, 0 failures, 35 excluded
```

Both `search_input forwards value and placeholder` and `text_input still forwards value` pass.

</details>

### Why existing tests did not catch this

`test/petal/form_test.exs` `"search_input"` only checks for `search` type and field name, not forwarded attributes. The `"text_input"` test does check `placeholder` and `itemid`, which is why `text_input` was not broken.

### Summary

`PetalComponents.Form.search_input/1` (and several sibling inputs) no longer forward globals such as `value` and `placeholder` to the rendered `<input>`, although they did in **3.0.1**. This breaks common LiveView patterns (pre-filled search fields, `phx-update="ignore"` on inputs, etc.).

### Affected versions

- **Broken:** 3.0.2+, verified on 3.2.2 and 4.0.6
- **Last known good:** 3.0.1

### Regression commit

https://github.com/petalframework/petal_components/commit/d2f957c1f5df10b8860f199aad7a4a292841ae47 (“Refactor form.ex”)

### Root cause

`render_basic_input/2` correctly merges `assigns[:rest]` into `Form.*_input/3` options, but Phoenix only populates `:rest` when the **immediately preceding** function declares `attr(:rest, :global, include: @form_attrs)`.

After the refactor, only `text_input/1` has that `attr` before a `render_basic_input` caller. `search_input/1`, `email_input/1`, `number_input/1`, `password_input/1`, `telephone_input/1`, `url_input/1`, `datetime_local_input/1`, and `date_input/1` do not — so `@rest` is `%{}` and attributes like `value` are dropped.

In 3.0.1, each of those functions had its own `attr(:rest, :global, include: @form_attrs)` block (see tag `v3.0.1`).

### Steps to reproduce

1. Checkout `v3.2.2` (or current `main`).
2. Add the test below to `test/petal/form_test.exs` (same style as existing `"text_input"` test).
3. Run `mix test test/petal/form_test.exs`.

```elixir
test "search_input forwards value and placeholder" do
  html =
    rendered_to_string(~H"""
    <.form :let={f} for={%{}} as={:user}>
      <.search_input
        form={f}
        field={:name}
        value="abc-123"
        placeholder="eg. Jane Doe"
      />
    </.form>
    """)

  assert html =~ ~s(value="abc-123")
  assert html =~ ~s(placeholder="eg. Jane Doe")
end
```

> This was mostly generated and reproduced by LLM models. I have confirmed it the regression through the above automated test passing on 3.0.1 which is the same I a saw this bug manifesting in internal systems where anything other than `text_input` stop forwarding the `@rest` attributes.

nhobes:
Fixes #488.

## The bug

Since 3.0.2, `email_input`, `number_input`, `password_input`, `search_input`, `telephone_input` and `url_input` silently dropped `value`, `placeholder`, `phx-*` bindings and every other `@form_attrs` global. Only `text_input` still forwarded them.

Root cause (well diagnosed in the issue, thanks @pmargreff): the d2f957c refactor collapsed the per-function `attr` blocks, but Phoenix `attr` declarations only apply to the function *directly below* them. So only `text_input` had `attr :rest, :global` in effect; the other six got no `@rest` at all, and `render_basic_input/2` fell back to `%{}`.

Note: `datetime_local_input` and `date_input` (also named in the issue, which was filed against 3.2.2) already have their `attr` blocks back on current `main`, so this PR covers the remaining six.

## The fix

Restore the `attr` block (`form`, `field`, `label`, `class`, `rest`) on each of the six inputs, same shape as `text_input`.

## Tests

Extended each input's existing test to assert forwarded `value` / `placeholder` (the originals only checked type and name, which is why the regression slipped through).

- Without the lib change: 6 failures — the new assertions catch the regression
- With it: full suite passes, 488 tests, 0 failures

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
 # Changelog
+### Unreleased
+
+#### Fixed
+
+- **`email_input`, `number_input`, `password_input`, `search_input`, `telephone_input` and `url_input` forward global attributes again** (`value`, `placeholder`, `phx-*` bindings, etc.). A 3.0.2 refactor left these six without their own `attr :rest, :global` declarations, and since `attr` only applies to the function directly below it, only `text_input` kept forwarding. Fixes [#488](https://github.com/petalframework/petal_components/issues/488).
+
 ### 4.2.0 - 2026-06-29
 
 A typography pass. The heading and body scale got a proper going-over, and the type set is now complete enough to lay out a full article from components alone. Everything here is additive or a restyle, with no breaking API changes (see Upgrading for the visual shifts).
--- a/lib/petal_components/form.ex
+++ b/lib/petal_components/form.ex
@@ -130,9 +130,7 @@ defmodule PetalComponents.Form do
     ~H"""
     <label class="pc-checkbox-label">
       <.checkbox form={@form} field={@field} {@rest} />
-      <div class={
-        label_classes(%{form: @form, field: @field, type: "checkbox", class: @label_class})
-      }>
+      <div class={label_classes(%{form: @form, field: @field, type: "checkbox", class: @label_class})}>
         {@label}
       </div>
     </label>
@@ -143,9 +141,7 @@ defmodule PetalComponents.Form do
     ~H"""
     <label class="pc-checkbox-label">
       <.switch form={@form} field={@field} {@rest} />
-      <div class={
-        label_classes(%{form: @form, field: @field, type: "checkbox", class: @label_class})
-      }>
+      <div class={label_classes(%{form: @form, field: @field, type: "checkbox", class: @label_class})}>
         {@label}
       </div>
     </label>
@@ -312,11 +308,23 @@ defmodule PetalComponents.Form do
     """
   end
 
+  attr(:form, :any, default: nil, doc: "")
+  attr(:field, :atom, default: nil, doc: "")
+  attr(:label, :string, default: nil, doc: "labels your field")
+  attr(:class, :any, default: nil, doc: "extra classes for the text input")
+  attr(:rest, :global, include: @form_attrs)
+
   def email_input(assigns) do
     ~H"""
     {render_basic_input(&Form.email_input/3, assigns)}
     """
   end
+
+  attr(:form, :any, default: nil, doc: "")
+  attr(:field, :atom, default: nil, doc: "")
+  attr(:label, :string, default: nil, doc: "labels your field")
+  attr(:class, :any, default: nil, doc: "extra classes for the text input")
+  attr(:rest, :global, include: @form_attrs)
 
   def number_input(assigns) do
     ~H"""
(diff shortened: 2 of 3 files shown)


nhobes:
The uncovered line was the `form_field type="switch"` label path — a pre-existing gap that mix format pulled into the diff by collapsing its `label_classes` call onto one line (project coverage was unchanged, 1929 lines / 269 misses before and after). Added a `form_field switch label` test mirroring the checkbox one, so the patch is now fully covered.

PR #498, opened 2026-07-10
Add command - the ⌘K palette

nhobes:
The classic command menu (⌘K), Phoenix-native. Nine composable pieces on the 4.4.0 surface system: `command`, `command_dialog`, `command_input`, `command_list`, `command_empty`, `command_group`, `command_item`, `command_separator`, `command_shortcut`.

## Why it beats a cmdk port
- **Items are real links and buttons.** `navigate`/`patch`/`href` render a `<.link>`; anything else is a button. Enter clicks the highlighted element, so every LiveView binding works with zero adapter code.
- **Native `<dialog>`.** Top layer, focus trap, `::backdrop` and Escape come from the browser. Global Cmd/Ctrl+K (configurable), `pc:command-open`/`close` events, `open_command/1` JS helper, reset-on-close.
- **Client-side filtering, zero deps.** The `PetalCommand` hook scores prefix > word-boundary > substring > fuzzy subsequence, with per-item `keywords` aliases. Items hide but never reorder - the server owns DOM order, so LiveView patches stay safe (re-filters on `updated()`).
- **Full WAI-ARIA combobox pattern.** Focus stays in the input; `aria-activedescendant` carries the virtual highlight; listbox/option/group roles; arrows/Home/End/Enter with optional `loop`; highlight re-homes to the first match on each query change.
- Groups auto-hide when empty, separators hide during a query, `command_empty` shows at zero matches.

## Dogfooded
The playground's topbar search is now real: the button and ⌘K open a `command_dialog` listing every page + theme actions. New Command page with inline + dialog demos.

545 tests green (12 new). Behaviour verified live in both modes: filter, keyboard model, loop, ⌘K, reset-on-close, empty state, Enter-to-navigate.

Ships with the next release (4.5.0) alongside the doctrine-port sweep.


nhobes:
Second commit lands two more ports (`0cc834d`):

**Rating, now interactive.** A real radio group (posts in forms, native arrow keys, focus-visible rings) with pure-CSS hover preview and fill via `:has()` — still zero JavaScript. Three icon sets: star, heart, and `face` — a five-point sentiment scale (Awful→Great) where each position is a distinct drawn expression with its own colour; faces single-select while stars/hearts fill cumulatively. Display mode is 100% back-compat (half-stars, legacy classes untouched).

**Slide_over, sheet-grade.** `description` attr, pinned `:footer` slot, aria-labelledby/describedby wiring, doctrine ghost close button, hairline border on the attached edge. Verification caught `JS.show` flattening the flex column to `display:block` (footer floated mid-panel) — fixed with `display: "flex"`. Public show/hide API unchanged.

Playground: Rating page (live form-bound hero, sentiment demo, fractional display) + Slide over page (edit-profile sheet with origin/width controls, cart with pinned checkout footer). 553 tests green, all 22 pages verified.

nhobes:
Third commit: rating review feedback — fractional display fill via clip overlay (3.5 hearts = half a heart; faces round to the nearest expression and the label carries the precision), labels unified on the muted tier with `label_position` right/bottom, display-mode faces no longer react to hover (selector was missing the interactive-group scope), and custom glyphs via a `:glyph` slot + `--pc-rating-active-color` token (flame spice-level demo in the playground). Interactive stays whole-step by design — it's a radio group posting integers. 555 tests green.

nhobes:
Fourth commit: `precision="half"` — interactive half-steps, still pure CSS. Two radio hit-areas per icon (left = x.5, right = x), so 3.5 is clickable, posts as `3.5`, and arrow keys step by halves natively. Fill overlay driven by `:has()` state rules with matching hover previews. Faces stay whole (ordinal scale) — the playground's new step seg (1 | ½) disables ½ for faces with a note. Verified live: fills 100/100/100/50/0 after clicking the 3.5 hit-area. 557 tests green.

nhobes:
Fifth commit: the high-traffic ports — **tabs** (new `variant="segmented"`: raised white pill on a muted track, the shadcn style; legacy `underline` flag still resolves; aria-selected now a proper string), **table** (sortable headers with `sortable`/`sort_key` on cols + `sort_by`/`sort_dir`/`on_sort`, direction chevrons, `aria-sort`; plus `density="compact"`, `striped`, `sticky_header` — the free presentation half of the data_table split), and **pagination** (discrete quiet buttons, current page rides primary solid — follows the colour dials; pure CSS, API untouched). Three new playground pages under Navigation/Data groups. 561 tests, 25 pages verified.

nhobes:
Sixth commit — review feedback on the ports: segmented number badge now legible in dark (solid-fg token; it was white-on-near-white under the monochrome ramp), pill gaps tightened (gap-8→gap-2) with underline keeping its classic spacing — and the underline container border turned out to have NEVER applied (`.pc-tabs--underlined` vs rendered `pc-tabs--underline`; unified). Playground: number-badge toggle + segmented-with-icons example, pagination sibling/boundary windowing segs, table variant seg (basic/ghost) + empty-state toggle + user_inner_td people-cells example, and the search/filter line drawn explicitly in the hint (data-layer = pro data_table). 561 tests.

nhobes:
Seventh commit: ghost table is now genuinely minimal per review — transparent muted header (no fill), font-medium, px-3/py-3 cells, softer hairlines; basic keeps the framed card-table look. Sort buttons inherit th weight so both variants match. Verified: th bg transparent, weight 500, 12px cells.

nhobes:
Eighth commit — table refinements checked against shadcn's current table.tsx source: universal row hover wash (every tbody row, one step above the stripe tone), ghost headers corrected to full-strength `text-foreground font-medium` (their current recipe — the muted version I'd ported was their old one), and a `:footer` slot rendering a proper tfoot totals row (their TableFooter classes; `<.td colspan={3}>` composition). Plus dead-CSS cleanup: `.pc-table thead` never matched. 562 tests.

nhobes:
Eleventh commit: **skeleton rebuilt** — the old component was eight frozen layouts with zero customisability (no class/rest/sizing). Now: `skeleton/1` the composable brick (size with classes, `variant` block/text/circle on theme radii, `animation` pulse/**shimmer**/none), `skeleton_text/1` (deterministic varied line widths — no re-render dance), and `skeleton_group/1` (the a11y wrapper most libraries skip: `role=status` + `aria-busy` announces once, animation cascades via `:where()` with per-brick override). Shimmer is reduced-motion-safe. Legacy `kind:` layouts still render; the one behaviour change is bare `<.skeleton />` = the brick now. Playground page includes a real simulate-loading demo. 565 tests.

nhobes:
Twelfth commit: **the rest of the free library** — 12 new playground pages (Button group, Loading, Breadcrumbs, Stepper, Avatar, Card, Accordion, Marquee, Spotlight card, Number ticker, Text animation ×4 functions, Confetti) under new Display/expanded groups. The playground now covers all 38 pages. Found in verification: the accordion's JS never updated `aria-expanded` after toggle (fixed in openItem/closeItem). Deferred: menu+container (future Layout page), user_dropdown_menu (dropdown polish), navigation_menu (queued enhancements), chat (finale). 565 tests, full sweep green, interactive demos (stepper jump, ticker bump, confetti, accordion) verified live.

nhobes:
Thirteenth commit — quality pass on the coverage batch per review: **marquee** edge fade rebuilt as a CSS `mask-image` (the old overlay divs were pinned to white/gray-900 and mismatched on zinc-950 — a mask reveals what's behind, correct on any background by construction); **pagination** current page moved from primary solid to the outline surface (shadcn's rule — pagination marks where you *are*, not what you *chose*; solid primary stays for checked form controls); **accordion** gains `size="sm"` compact rows (ghost variant already existed); **loading** page rebuilt around real-world states (loading buttons, inline status, generating panel). 565 tests.

nhobes:
Fourteenth commit — design-audit round two: **stepper** fully onto the doctrine (was gray/green off-system blobs; now muted-outline upcoming → primary-ring active → primary-solid complete with solid-fg check, primary connectors, follows the dials), **card** calmed (shadow-xs, text-sm body, semibold lg heading), **accordion** ghost rhythm bug fixed (ghost had no base padding so compact was *taller* — base py-4 now) + default at shadcn scale + 'multiple' verified working with clearer copy, **breadcrumbs** get the current-page treatment (strong last crumb + `aria-current=page`) and tighter modern separators, **avatar** demos use real local portrait photos (5 × 4KB, dev-only). 566 tests.

nhobes:
Fifteenth commit — the try-harder round, checked against shadcn's actual sources: **accordion** default is now their exact row recipe (border-b rows, py-4 font-medium hover:underline trigger, size-4 nudged chevron) with the boxed style preserved as `variant="bordered"`; **card** gains `card_header` (title/description/:action) and the playground hero is their login-card example verbatim, with variants that finally differ (basic = elevated panel, outline = blending surface, shown on a tinted backdrop); **stepper** geometry root-caused — a legacy pl-4/5/6 hack block below the new rules was fighting alignment, container lacked w-full — now horizontal bars are visible (32px) and vertical is 0.0px off; missing `__title` CSS added; colour stays primary by doctrine (progression = selection signal, not success). Breadcrumb icons already existed — now demoed. 567 tests, 38 pages.

nhobes:
Sixteenth commit — card variants reconciled per review: the transparent 'blending' outline was nonstandard and is gone. The trio is now the industry one: **basic** (border + shadow-xs — shadcn's card, unchanged), **outline** (same panel, no shadow — its honest historical meaning), **muted** (NEW — the tinted borderless well: GitHub boxes / Material filled / Tailwind UI wells). Three distinct jobs, demoed side by side. Media card gets a real 36KB local cover photo (dev-only, like the avatars). 569 tests.

nhobes:
Seventeenth commit — card taxonomy final: **two variants**. basic (the bordered panel — shadcn's one-and-only card) + muted (the tinted well). `outline` was indistinguishable from basic (shadow-xs is imperceptible, doubly so in dark) so it's now a documented **legacy alias of basic**, retired in 5.0 with the legacy button variants — zero visual change for existing apps. A variant you can't tell apart doesn't earn its keep. 569 tests.

nhobes:
Eighteenth commit — accordion by the glance test: **ghost is a legacy alias of default** (it was the same rows with different icons; 5.0 removal, zero API break), and the open-header highlight is now **scoped to bordered only** — the JS was adding it to every button and the unscoped CSS painted a sharp gray bar across the borderless rows (Matt's 'quirk'). Rows open clean; bordered keeps its soft fill as the which-is-open signal, so its content bg stays put. Verified live in dark. 569 tests.

nhobes:
**Avatar upgrade** (`1b4923b`) - closing the gap with shadcn's avatar and then some:

- `status` attr on `<.avatar>`: online / offline / busy / away renders a ringed presence dot pinned bottom-right, sized per avatar (`xs` 6px up to `xl` 14px). `role=status` + `aria-label` for screen readers.
- `max` attr on `<.avatar_group>`: caps the stack and folds the rest into a `+N` bubble (`aria-label="N more"`).
- Fixed a real dark-mode bug the group examples exposed: the stack ring was `dark:ring-gray-100` (near-white halo on dark). Now `dark:ring-zinc-950` matching the page background - same principle as shadcn's `ring-background`. Status dots use the same token.
- `card_media` now declares `alt` (it isn't a Phoenix global, so it was silently dropped before).
- Playground avatar page restructured into labelled sections: sizes, fallback chain, a presence team list, and stacked groups with the +N overflow - real-life examples instead of one flat row.

nhobes:
**Button group rebuilt as a composition wrapper** (`0337f07`) - audited against [shadcn's button-group](https://ui.shadcn.com/docs/components/base/button-group) and matched its model:

Before: a closed `:button` slot API that rendered its own bare buttons - one look, no real `<.button>`s, no inputs, no orientation.

Now the group fuses **any children**:
```heex
<.button_group aria_label="Merge options">
  <.button>Merge pull request</.button>
  <.button_group_separator />
  <.button aria-label="More"><.icon name="hero-chevron-down" class="w-4 h-4" /></.button>
</.button_group>
```

- Fuse mechanics: inner radii collapse, inner borders collapse to a single 1px line, outer corners keep `--pc-radius`. Works with buttons (any color/variant/size), `<.input>`, selects.
- `orientation="vertical"` fuses top-to-bottom.
- `<.button_group_separator />` for between solid buttons (their borders are transparent). It tints from `--pc-button-solid-fg` via `color-mix`, so it's white-on-dark-solid in light mode and dark-on-white-solid in dark mode automatically - shadcn's `bg-input` can't do that.
- `<.button_group_text>` for non-interactive rail segments (`https://` prefix, units, kbd hints).
- Nested groups stop fusing and gap into clusters (`:has`), so toolbars come free.
- Detail work: a focused child rises above its fused neighbours so the ring never clips; `active:scale` is neutralised inside groups so pressing a segment doesn't crack the strip open; inputs flex to the leftover rail width.
- **Back-compat:** the legacy `:button` slot API renders byte-identical behind a `pc-button-group--legacy` modifier. All 5 existing tests pass untouched. 574 total green.

Playground page went from two static rows to: split-button + view-switcher hero, a nested editor toolbar, sizes, mixed rails (text prefix + input + button, search + solid), and a vertical zoom stack.

nhobes:
**Button-group subtlety pass** (`799aaaa`) per Matt's review:

- **Hero split is outline now.** Realisation: an outline split button doesn't need the separator at all - fusing already draws the 1px divider from the buttons' own borders. So `Merge pull request | v` reads as a split button while staying as quiet as the rest of the page.
- **The solid split moved to its own labelled section** ("solid buttons have no borders, the separator is the divider") - that's the separator's actual job, demonstrated at normal scale instead of shouting from the hero. Same reasoning as shadcn, who demo their split on `secondary`, not primary.
- **Separator got smarter:** default is now border-coloured (`gray-300` / `white/15` - works between muted surfaces), and only when it sits next to a primary solid does it switch to the `color-mix(--pc-button-solid-fg 25%)` tint. Both canonical cases work with zero user classes.
- **Search rail:** solid `Search` -> outline magnifying-glass icon button.
- **Square icon buttons everywhere they belong:** the library already had `size="icon"` (square, `p-2`) - the toolbar (undo/redo/bold/italic/link), both split chevrons, search, and the vertical +/- now use it. Measured 34x34 squares; the +/- padding is symmetric as requested.

574 tests green.

nhobes:
**Split button is now interactive** (`b220695`):

The hero Merge pull request split opens a real dropdown - the canonical GitHub trio: **Create a merge commit** / **Squash and merge** / **Rebase and merge**, each with an icon and a one-line description, right-aligned under the chevron. Verified opening in both light and dark (menu surface tracks to zinc-900 in dark).

The interesting bit was fusing our `<.dropdown>` into a group. The component wraps its trigger in two divs (`.pc-dropdown > div > button`), so the group's structural fuse rule (`> :not(:first-child)`) can't reach the actual button, and the button wouldn't stretch to the rail height. Fix is contained: a scoped rule flexes the two wrappers through so the trigger fills the height, and the collapse classes (`rounded-l-none border-l-0`) ride on `trigger_class`. So `<.dropdown>` drops into any button_group as a first-class split-button citizen now - no component changes, just 2 CSS rules. 574 tests green.

nhobes:
**Aurora ported from Pro (rebuilt) + the last six coverage pages** (`e1010e2`) - the playground now covers the entire petal.build components list (44 pages).

**Aurora** is free now, and better than the Pro one:
- `colors` attr builds the gradient from any 3-6 colour palette - `<.aurora colors={["#f97316", "#f43f5e", "#fbbf24"]}>` and done
- Found + fixed a real rendering bug while porting: the Pro version hardcodes the black stripe texture, so on light backgrounds the invert filter hue-shifts every palette (the default blue renders **gold**). The stripe colour and invert must flip together per mode (white+invert on light, black+none on dark) - that's how the original Aceternity effect works. Light mode now shows true blues; custom palettes keep their hues in both modes.
- Wrapper sizes to its content (Pro needed a sized parent), content renders above the glow so text stays crisp, `prefers-reduced-motion` freezes the drift, the `PetalAurora` hook pauses it off-screen, decorative layers are `aria-hidden`, and everything is `pc-`prefixed so it can't collide with the Pro classes (the border_beam lesson).

**New pages:** Links (the routing primitive), Icons (heroicons gallery + variants + class sizing), Menu (grouped sidebar, active state), Navigation menu (flyout verified opening: 3 rich links + CTA footer, aria-expanded, click-away), Container (max-width staircase), Aurora (hero with real copy + palettes + tuning).

577 tests green.

nhobes:
**AI Chat: gap-fill vs shadcn's bubble/marker/message-scroller + a live playground demo** (`e31beea`)

- **`Chat.marker/1`** - the missing piece vs shadcn's Marker: inline system notes ("Searched the web - 8 sources"), labelled separators ("Today"), bordered status rows, and a `loading` state (spinner + `role=status`). The shimmering streaming-status treatment composes the existing `shimmer_text/1` - no duplicate shimmer.
- **Scroller upgrades** in `PetalChatScroll`: conversations now open at the latest message, and **prepend preservation** - loading older history above no longer moves the reader. Measured live: **viewShift 0px** while 155px of history inserted above. Implementation is anchor-tracking (top visible id'd row) rather than added-node inspection, because morphdom recreates un-keyed siblings; that also produced a documented rule on `conversation/1`: give thread rows stable ids (or use streams).
- Turn anchoring already existed (`PetalChatStream.anchorTop` pins your question to the top when the answer starts) - no change needed.
- **The playground got its finale: a genuinely live AI Chat page.** New "AI" nav group. The demo streams word-sized tokens from the LiveView via `push_event` - typing indicator, stop button mid-stream, then the bubble snaps to rendered (sanitised) markdown on commit. Suggestions chips, "Load earlier messages" demonstrating the scroll preservation, a markers section, and tool calls/reasoning/errors. No JS SDK anywhere - it's all the socket.

Deliberately NOT added: reactions/read receipts (social-chat territory, parked in the roadmap, demand-gated). AI-side feedback is already covered by `message_actions`.

580 tests green. The playground now covers **every** component page on petal.build - 45 pages.

nhobes:
**Fix: chat dark mode** (`3fa6d35`) - Matt caught the user bubble rendering white-on-white in dark. The chat theme tokens predate the monochrome ramp: the bubble bg rides `primary-600` (which flips near-white in dark) but `--pc-chat-user-fg` was hardcoded `#ffffff` - and the composer send/stop buttons repeated the same pairing. The assistant bubble had no dark values at all (zinc-100 card in a dark thread).

Fix is one token: `--pc-chat-user-fg` now rides `--pc-button-solid-fg` - the same label-on-solid token buttons/pagination/steppers use - so it flips with the mode and tracks custom primary ramps. Assistant tokens got proper dark values (white/6 surface, zinc-100 text; markdown inherits). Send/stop/stop-icon all inherit the token now. Verified both modes live - dark reads like ChatGPT dark, light unchanged.

nhobes:
**Chat theming follow-ups** (`df9cdc1`) - Matt's dark-mode review found three distinct root causes:

1. **The primary dial wasn't reaching the chat bubble.** The playground's `@theme inline` rewires *utilities* only - it never emits `--color-*` variables - so the chat's `--pc-chat-user-bg: var(--color-primary-600)` stayed pinned at the package default while `--pc-button-solid-fg` (a plain var) tracked the dial. Net effect: pick blue, get white text on a still-monochrome bubble. Fixed with a raw-var bridge in `dev/colors.css` (`:root` + `[data-primary]` so the mapping re-resolves against each dial's overrides). Real apps are unaffected - a standard `@theme` emits the vars. Bubbles + Send now go blue-600/white under a blue primary in both modes.
2. **The "Today" separator line was near-white in dark** - `dark:` variants inside `::before`/`::after` selectors compile to an invalid selector and silently drop. The line colour now lives on the parent as `--pc-chat-marker-line` (gray-200 light / white-10 dark).
3. **Bonus find from sweeping for the same bug class: the Input OTP caret's `dark:bg-gray-100` never compiled** - shipped invisible-in-dark since 4.4. Same fix pattern (`--pc-otp-caret` on the slot).

580 tests green.

nhobes:
**Fix: the chat now follows the live edge while streaming** (`7a59245`) - Matt noticed that sitting at the bottom and sending a message stranded you: the response streamed in below the fold and you had to scroll down manually every time.

Root cause was a broken interaction: `anchorTop` tried to pin your new question to the viewport top, but at send time there's no answer content below it yet, so the scroll just clamps at the bottom edge - then tokens grew the thread below the fold with **no follow** and **no jump button** (it only re-evaluated on scroll/patch events, which token appends don't produce).

New model - the shadcn `autoScroll` / ChatGPT convention Matt described:
- **At the live edge you ride it**: pinned through the append patch and through every streamed token (the edge check runs BEFORE each token lands, so content growth can't disengage a follower).
- **Scroll up and it lets go instantly** - nothing ever yanks you back down. Each token now dispatches a scroll event, so the jump-to-latest arrow appears *during* streaming (it previously couldn't), and clicking it re-engages the follow.
- `anchorTop` is gone - question-to-top anchoring conflicts with edge-riding in long threads.

Verified live across the matrix: slack 0 through patch/stream/commit at the edge; mid-stream scroll-up holds position with the button visible; jump re-engages; history-prepend preservation still 0px shift. 580 tests green.

nhobes:
**Chat UX round from Matt's review** (`c0704c4`):

- **Submit drops to the live edge.** Matt checked Claude/ChatGPT and corrected the spec: your own send should jump to the bottom even if you're scrolled up - it's an intentional act, unlike incoming content which still never yanks you. One listener in the composer hook (pin + re-engage follow before the patch lands). Verified: scrolled to top, submitted, slack 0 instantly and through the whole stream.
- **Message action bar** - the ChatGPT/Claude row under answers: new `action_button` (icon-only with aria-label/tooltip, any phx binding), `copy_button icon` mode (clipboard flips to a check via the hook), and `message_actions visible="hover"` for per-row fade-in (always visible on touch, where hover doesn't exist). Feedback stays app-side by design - thumbs push your phx-click events.
- **Send button is the arrow-up icon by default** now (36px square, matches the field height); `submit_label` restores a text button. The suggestions starter chips were already a component (`Chat.suggestions`) - now actually documented in the demo note.

584 tests green.

nhobes:
**Chat: plain variant is the new default + actions slot** (`d1b1c96`) - Matt noticed some AI apps don't bubble the assistant at all. Convention audit confirmed it's not "some": **ChatGPT, Claude and Gemini all render assistant text plain on the surface** - only the user message gets a bubble, and the action bar sits below the text. Bubbles-both-sides is the messenger convention (iMessage/Intercom), not the AI one.

- `conversation variant="plain"` is the new default: assistant/system text on the surface (colour still rides `--pc-chat-assistant-fg`), user keeps the right-aligned bubble, assistant rows span full width. `variant="bubbles"` is one attr away for messenger-style chat - both audiences served.
- `chat_message` gained an `:actions` slot that renders **below the message, outside the bubble** (new `pc-chat__body` column carries the width cap) - so the bar placement matches the reference apps in both variants, and composing it inside the bubble still works for anyone who prefers that.

The demo now reads like ChatGPT: dark user bubble right, plain full-width answer with code block, icon bar (copy/thumbs/regenerate) under the text. 586 tests green.

nhobes:
**Chat page controls + reasoning fixes** (`97e0ac5`):

- **Playground toggle:** the chat page grew the standard control strip - `variant` (plain | bubbles) flips the thread style live, same seg pattern as the accordion/tabs/table pages. One control, no bloat.
- **Reasoning indent bug** (pre-existing, Matt caught it): the body had `whitespace-pre-wrap`, which faithfully rendered the HEEx template's leading indentation - first line pushed way right, continuation lines flush left. Removed; multi-line content flows normally now.
- **Disclosure affordance:** the reasoning summary has a rotating chevron (> closed, v open) so it reads as clickable - the native details marker was suppressed, which left zero visual cue.

587 tests green.

nhobes:
**Chat page: actions-visibility toggle** (`8c8903f`) - Matt asked whether `visible="hover"` still works now the bar lives outside the bubble. It does, verified with a real mouse hover: the reveal keys on `.pc-chat__row` (the whole message row), not the bubble, so slot placement is irrelevant - and it's `opacity`, not `display`, so the bar always reserves its space (no layout shift on reveal, and moving the cursor onto the bar itself keeps it open). Added `actions (always | hover)` as the second control in the chat page's strip so it's feelable.

nhobes:
**Chat: hover-mode polish** (`0fee483`) - Matt flagged that `visible="hover"` left what looked like a permanent dead gap under the answer. Root cause was demo composition plus a missing convention:

- Space reservation is deliberate (collapsing it causes the classic cursor-chase flicker: bar appears -> layout grows -> content shifts under the cursor). What makes it read as *rhythm* instead of a hole: **every assistant turn carries the bar** (demo now does), and **the newest message keeps its bar always visible even in hover mode** - a one-rule `:last-child` override, same behaviour as ChatGPT (persistent on latest, hover on older).
- Verified live: seed answer's bar visible while it was the latest; after streaming a new turn the new answer holds the persistent bar and the older one fades to hover-reveal.
- Playground control heading renamed to **action bar** (always | hover).

nhobes:
**User-message actions (copy + edit)** (`bcaa5bc`) - Matt asked whether the action bar should extend to user messages too. It already can: the `chat_message :actions` slot is **role-agnostic** - no new component surface, just `copy_button` + `action_button` in the slot. So this is a demo/docs change, not a feature build.

- The demo's user question now carries **copy + edit**, right-aligned under the bubble (verified: bar right edge = bubble right edge), hover-revealed via the same `action bar` toggle. This matches ChatGPT/Claude, which both show copy/edit under user turns.
- **Edit is wired for real** - clicking it loads the message text back into the composer (draft assign + `phx-change` binding, cleared on send), so the demo shows a working edit, not a dead button. The *behaviour* stays app-side by design (like thumbs): we ship the button, the app decides what edit/rewind do.
- Documented the slot as role-agnostic on `chat_message`.

588 tests green.

nhobes:
**Demo: edit forks the thread** (`120350f`) - Matt spotted that editing a message and sending appended a duplicate instead of ChatGPT's replace-and-regenerate. Fixed in the demo:

- Editing a user message + sending now **replaces that message, drops everything after it, and regenerates** - a fork, not an append. Verified live: two full exchanges -> edit the first message -> collapses to a single clean exchange.
- Needed to seed the conversation as data (the opener was hardcoded HEEx, so it couldn't be replaced). Turns now hold the seed Q&A; `chat_edit` records the target index; `chat_start` truncates with `Enum.take(turns, i)` before regenerating. Side benefits: suggestions gate on a `sent` flag, and every committed answer now gets per-code-block copy.
- This is demo behaviour only - the component ships the edit button and the app decides fork-vs-append, same boundary as thumbs. But the demo now teaches the right pattern.

588 tests green.

nhobes:
**prompt_input edit-mode banner** (`c4b197b`) - Matt flagged that our edit silently loaded text into the composer with no "you're editing" signal and no cancel. Evaluated both reference patterns:

- **Claude** edits inline in the message bubble - elegant, but requires every message to become a form with its own textarea/state/focus/submit, which fights the `message = display, composer = input` separation that keeps the LiveView kit clean.
- **ChatGPT** shows an edit banner on the composer with a cancel X - reuses the one input surface.

Went with ChatGPT's pattern as a real `prompt_input` feature: `editing`, `edit_label`, `on_cancel_edit` attrs. When editing, a header banner (pencil + "Editing message" + cancel X) sits above the field. The composer restructured to a column, but the non-editing render is byte-identical (a `-row` wrapper carries the old layout), so existing users see no change. In the demo, edit shows the banner, cancel clears it, and the message being edited gets a subtle primary ring tying it to the composer. Verified live end to end. 591 tests green.

nhobes:
**Fix: composer lost focus while typing** (`eea9f4c`) - Matt hit it editing a message: delete a letter and the field de-focuses, so you have to re-click for every keystroke. Cause was mine - I'd made the textarea a *controlled* input (`value=` + `phx-change`), so every keystroke round-tripped to the server, re-rendered the textarea, and reset the caret. Controlled textareas are simply the wrong pattern for a chat composer.

Fixed at the component level: the `prompt_input` textarea is now `phx-update="ignore"` with a stable id, so LiveView never patches it - typing is pure client-side, focus is never lost. `value` becomes the mount-time initial only; to set it later (edit, quote, clear) you push a `pc-chat-set-input` event to the `PetalChatComposer` hook, which sets the value, focuses, and puts the caret at the end.

Verified live: deleted + typed characters with focus retained through every keystroke (`[true,true,true,true]`), the edited text submits correctly, the composer clears after send, and - bonus - typed-but-unsent text now survives unrelated re-renders (variant toggle) too. This also makes the composer correct for real apps, not just the demo. 592 tests green.

nhobes:
**Fix: 'Yesterday' messages had no action bar** (`78bbf09`) - Matt caught it. Cause: the history messages were hardcoded scaffolding rendered *outside* the turns loop (which is where the action bars live), so they silently lacked them while 'Today' had them.

Unified it - history and the date markers are now real turns (with stable ids so scroll-preservation still works), rendered through the single message loop. Every message, old or new, gets the same action bar. 'Load earlier' just prepends the history turns.

That refactor exposed a genuine scroll-hook gap: it decided edge-follow purely on "was I at the edge before the patch", so loading history while near the bottom jumped you to the bottom instead of holding position. Fixed the precedence: if the anchor row moved, content changed *above* it (a prepend) so hold the reader there; only fall back to stick-to-bottom when nothing moved above (a genuine append). Verified both: prepend shift **0px** (was -178), and follow-at-edge still pins to **0 slack** through a live stream. 592 tests green.

nhobes:
**Stepper wizard + nav/menu polish** (`490b79c`, `81e96bf`) - four things from Matt's review:

1. **Stepper is a real wizard now** (reactbits-style): the indicator on top, per-step content (account form -> workspace -> invites -> review summary), Back/Continue that walk the flow, Complete on the last step, and a done state with Start over. Steps still jump on click. Playground-only.
2. **navigation_menu opens on hover by default.** New `trigger` attr - default `hover` (pure CSS: `:hover`/`:focus-within` with a transparent bridge across the gap so the pointer never drops the panel; no round-trip, keyboard + touch open via focus). `trigger="click"` keeps the old explicit JS toggle for touch-first apps. Clicking to open a top nav felt odd; hover is the convention.
3. **Nav-menu surfaces realigned to the doctrine.** They were the old Tailwind-example grays (`bg-gray-800`, `ring-gray-900/5`, `bg-gray-700` tiles). Now `bg-zinc-900` + `border-white/10` panels like dropdown/command, white-alpha washes on tiles and hovers, `primary-300` current/hover text in dark. Verified the flyout in dark - consistent now.
4. **Vertical menu active text was too dark in dark mode** - it used `primary-400`, which on the monochrome ramp is a dark gray (#52525b) sitting on a `white/10` wash. Now `primary-300` (#d4d4d8, readable), and it tracks a themed primary too.

Heads-up for the changelog: `navigation_menu` default flips click -> hover. Existing users get the better default automatically; anyone wanting the old behavior sets `trigger="click"`. 593 tests green.

nhobes:
**Nav hover reworked into a JS hook** (`18b7ce7`) - Matt hit both limits of the pure-CSS version: crossing the gap from trigger to panel closed it (CSS `:hover` has no close delay), and a wide panel spilled off the right edge (CSS can't reposition). Checked how shadcn/Radix do it - a JS state machine - and built the equivalent:

- **`PetalNavMenu` hook**: opens on hover/focus; on leave it schedules the close ~140ms later, so moving onto the panel fires a fresh enter that cancels it - the panel holds open across the gap. No more accidental close.
- **Viewport nudging**: on open it measures the panel and `translateX`-es it back inside the window if it would overflow an edge. Verified: a panel that was ending at x=849 on an 829px viewport now shifts -28px to sit at 821, fully on-screen (the "Contact sales" footer link is no longer clipped).
- Bonus: `aria-expanded` now tracks open/close (the pure-CSS version couldn't). Escape and focus-out close.

CSS moved from `:hover` to the hook's `--open` class, with `:focus-within` as a keyboard fallback if the hook isn't registered. `trigger="click"` unchanged. 593 tests green.

nhobes:
**Nav hover + click-to-toggle** (`15433ca`) - Matt noticed shadcn lets you click an open trigger to close it, while ours only opened on hover and ignored clicks. Added the toggle:

- Click an open panel -> closes; click again -> reopens. While the pointer stays on the trigger, `pointerenter` doesn't refire, so a click-to-close stays closed until you hover away and back (matches shadcn exactly).
- **Touch gating**: `pointerenter`/`leave` now skip `pointerType === "touch"` - otherwise a tap would fire enter (open) then the click toggle (close) and flash. On touch the tap alone toggles.
- Added an outside-`pointerdown` listener so tapping/clicking away closes any open panel (the touch dismiss, and mouse click-away).

Verified all seven paths live: hover-open, click-close, stays-closed-while-hovering, click-reopen, click-outside-close, touch-no-flash, touch-tap-toggle. 593 tests green.

nhobes:
**Fix: click-to-close actually works now** (`fc05322`) - Matt couldn't reproduce click-to-close on the live playground, and he was right. A real mouse click *focuses* the trigger button, and I'd left a `:focus-within` CSS fallback that kept the panel visible while the item held focus - so the hook dutifully removed the `--open` class but focus-within overrode it and held the panel open. My synthetic `.click()` test never moved focus, so it false-passed. Re-verified with the actual mouse this time.

- Visibility is now driven **only** by the hook's `--open` class, never `:focus-within`, so a click that focuses the trigger still closes the panel.
- Opening is on **activation** (hover / click / Enter-Space's native click), not focus - which is the correct W3C disclosure pattern and also kills a touch wart (tap would fire focus-open then click-close = flash). Focusout still closes when you tab past an open item; Escape closes with focus kept on the trigger.

Verified live with **real clicks + keyboard**: hover-open, click-close (the fix - confirmed the trigger is focused yet the panel closes), click-reopen, hover-away-close, Enter-open, Escape-close-while-focused. 593 tests green.

Lesson logged: use the real input tool for input-driven behavior - synthetic events skip side effects like focus.

nhobes:
**Radius consistency pass** (`cf5cc37`) - Matt's design-system audit found surfaces that ignored the radius dial, plus the checkbox bug:

- **Checkbox went fully round at 14/full** - the derived radius (`--pc-radius - 6px`) had no ceiling, so 8px+ on a 16px box = a circle, which reads as a radio. Now clamped to 4px max: a rounded square at any dial setting. (Radios stay round - they should.)
- **Chat composer input + send + stop** were hardcoded `rounded-lg`; now track `--pc-radius` like every other input/button.
- **Nav flyout panel** was hardcoded `rounded-2xl`; now `min(--pc-radius, 1rem)` like the dropdown/command/popover panels.

The rule I applied: form controls and overlay surfaces follow the radius dial (clamped so small controls never round off); the **chat bubble** keeps its own `--pc-chat-radius` because a bubble's roundness is the metaphor, not an app-wide preference. Verified at radius 14/full: checkbox 4px, chat input/send 14px, nav panel 14px. 593 tests green.

nhobes:
**Sidebar demo (shadcn-grade) + stepper scrollbar** (`7ffed51`):

**Sidebar** - the Menu page is now a full sidebar, and the best part: it's **all composition, zero new component**. Turns out `vertical_menu` already supports collapsible sub-items (an item with a nested `menu_items` renders a toggle + rotating chevron + expand/collapse), which is exactly shadcn's collapsible tree. So the demo assembles:
- **Workspace switcher** (header) - a `dropdown` with a custom trigger (logo tile + name + chevron), opening the workspace list + "Add workspace"
- **Grouped nav** - `vertical_menu` with Platform/Projects groups, icons, and a collapsible **Playground -> History / Starred / Settings** (open because History is the current_page)
- **Account menu** (footer) - a `dropdown` with avatar + name that **opens upward** (`menu_items_wrapper_class="top-auto bottom-full"`) so it doesn't clip off the bottom
- framed by a main content area with a breadcrumb

Verified live in both modes: collapsible toggles, workspace opens down, account opens up, all within the card. This maps to shadcn's sidebar-07 - and it stays honest about the free/paid line: the note points out Petal Pro's SidebarLayout is what wraps this with collapse + a mobile drawer.

**Stepper scrollbar** - the horizontal demo showed an ugly bar in the constrained playground width; hidden it (`scrollbar-width:none` + `::-webkit-scrollbar`) while keeping it scrollable. It's a demo-space artifact - real pages with room don't overflow.

593 tests green.

nhobes:
**Sidebar polish + generic identities** (`93717d4`) - Matt's review:

- **Full-width hover / chevron to the edge**: the trigger button wasn't `w-full`, so the hover highlight and chevron only spanned the content, not the sidebar. Added `trigger_class="w-full"` to both triggers - now the highlight fills the rail and the chevron sits flush right, like shadcn. (Verified: trigger 239px = sidebar 256 minus padding.)
- **Genericised every example identity** across the playground so nothing references the team: Acme Inc -> Northwind, Startup Co -> Vertex Labs, Matt Platts -> Alex Rivera, matt@petal.build -> alex@example.com, plus the slide_over profile form and the table demo. Good catch - these ship in the public repo.

593 tests green.

nhobes:
**Nav radius consistency + trigger toggle** (`3a190cc`) - Matt's design-system question:

The flyout **panel** already tracked `--pc-radius` (my earlier fix), but its **inner** elements didn't - the top-level links/triggers (`rounded-md`), the flyout link rows (`rounded-lg`), and the icon tiles (`rounded-lg`) hardcoded their corners. So at radius 0 the panel squared off but the hover rows and icon tiles stayed rounded - exactly the inconsistency you spotted. All three now derive from the dial via the same clamp the dropdown/menu rows use.

Verified: **radius 0** -> panel/row/tile/trigger all 0px (square). **radius 14** -> panel 14px, inner rows/tiles 10px (the `--radius` - 4 = `--radius-sm` relationship, so the outer surface is rounder than the inner rows - the harmonious look).

Also added a **trigger (hover | click)** toggle to the nav page - the one attr worth feeling - keyed the nav id to the mode so switching fully remounts (avoids a stale inline display from the click-mode JS.toggle). 593 tests green.

nhobes:
**Greptile sweep** (`522e875`): all 27 of Greptile's live comments are the **same P1 re-flagged on each commit** - the `PetalCommandDialog` hook registered five listeners on the dialog element in `mounted()` but `destroyed()` only removed the document keydown shortcut. On a reused dialog node across a hook remount, the stale `close`/`click` handlers could run `reset()`/`close()` against a torn-down hook and dispatch duplicate bubbling `input` events. Valid finding - `destroyed()` now removes every listener `mounted()` adds. No other distinct issues in the queue. 593 tests green.

PR #501, opened 2026-07-27
Deployable playground via PLAYGROUND_DEPLOY (playground.petal.build)

nhobes:
Hosts the component playground publicly on Fly (app `petal-components-demo`, then `playground.petal.build`).

## What
- `dev.exs`: `PLAYGROUND_DEPLOY=true` switches the same playground into deploy mode:
  - **Compiles out `Phoenix.CodeReloader`/`LiveReloader` in `Dev.Endpoint`** - the reloader re-evaluates the entire 8.6k-line file on every request. Measured: **12.7s/request with it, 2ms without**. (`live_reload: false` alone does not help; the plugs were hardcoded.)
  - Tailwind watcher off, `debug_errors` off, `SECRET_KEY_BASE` required from env, LiveView `check_origin` locked to the public hosts.
- `Dockerfile`: MIX_ENV=dev image (every playground dep is `only: :dev`), tailwind binary pre-installed at build, CSS still built at boot as usual.
- `.dockerignore`: `_build`/`deps` excluded (local checkouts have absolute-path symlinks that dangle on builders - the exact failure mode that broke CI on #500).
- `fly.toml`: 1GB shared-cpu-1x, one machine always warm.

## Verified locally
- Deploy mode: `?c=button/slider/toast` all 200 in ~2ms after boot; `/assets/app.css` 200 (1.7MB); `/dev-static` 200; unknown route renders plain "Not Found", no stacktrace.
- Dev mode unchanged: reloader + live-reload iframe still active with the flag unset.

Deploy steps after merge (separate, not in CI): `fly secrets set SECRET_KEY_BASE=...`, `fly deploy --remote-only`, measurement gate, then `fly certs add playground.petal.build`.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,23 @@
+# _build and deps MUST stay out of the build context: in local checkouts and
+# worktrees they can be symlinks to absolute macOS paths (dangling on the
+# builder), and even as real dirs they're hundreds of MB the image rebuilds
+# anyway. The generated CSS is rebuilt at boot.
+_build
+deps
+priv/static/assets
+.git
+.github
+.elixir_ls
+test
+guides
+doc
+cover
+tmp
+screenshot.png
+logo.png
+*.md
+!rules.md
+.DS_Store
+fly.toml
+Dockerfile
+.dockerignore
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,9 +15,14 @@ RUN apt-get update -y && \
 
 WORKDIR /app
 
+# MIX_HOME/HEX_HOME live under /app so the runtime user can read them -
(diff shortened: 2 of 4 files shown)


nhobes:
On the remaining P2 (_build ownership): keeping it writable is deliberate. The entrypoint is `mix run`, and Mix 1.17 takes file locks under the build path on every invocation - a read-only `_build` risks a boot crash. Source (`lib`), `deps`, `config` and the mix files are root-owned read-only; the write surface is generated CSS, generated heroicons, and `_build`. An attacker who can write `_build` beams from inside the app is already executing inside the BEAM, so the extra restriction buys nothing for this no-data demo app. Documented in the Dockerfile.

PR #526, opened 2026-08-02
fix(css): icon sizing drops !important for the doubled-selector contract

nhobes:
## What

Converts every icon-sizing `!important` in `default.css` to the (0,2,0) doubled-selector contract that the colour-scheme controls proved out in 4.9.0, and adjudicates the rest of the file's important declarations.

**The premise was wrong.** The old comments claimed a consumer's heroicons plugin sizes `.hero-*` in the utilities layer. It doesn't: the plugin every phx.new app vendors registers icons via `matchComponents` - the **components layer**, same as our rules, at (0,1,0). So the old `!important`s were both overkill and a lockout: important component declarations beat even a consumer's `w-6!` (importance reverses layer order), making these icons genuinely un-resizable.

**The contract.** Doubling the class gives (0,2,0): outranks the plugin in either import order with no `!important`, while a plain `w-* h-*` utility on the icon still overrides via the utilities layer. Geometry sits in the doubled rule; colour/transition stay at single-class specificity.

Converted (9): pagination prev/next chevrons, accordion plus/minus, vertical-menu chevron, avatar placeholder silhouette, chat reasoning chevron, chat marker icon, chat action icon, chat composer-send icon.

Dropped as redundant (3): carousel slide-mode `position/top/left` - its (0,3,0) selector already beats the (0,1,0) base rule.

Kept, now with honest comments (2): showcase `<pre>` background fights an **inline style** from mdex's `:html_inline` formatter (only `!important` outranks an inline style); reduced-motion `1ms` is the standard a11y pattern.

## Verification

- **Live-browser cascade test** against a worst-case simulated plugin (`@layer components`, single class, injected *after* our stylesheet - later source order): all nine sites hold their intended size, and `w-6` on an icon overrides to 24px. **11/11 cases pass.**
- The test also caught a real bug pre-push: a `*/` inside a comment's `w-*/h-*` text terminated the comment early and the parser's error recovery silently ate the following rule. Comment reworded; rule verified compiled and applied.
- `mix test`: 679 tests, 0 failures, 1 skipped.
- Playground compiles and serves with the new rules.

## Deliberately out of scope

The legacy `has-error` form styling and the scheme-dropdown positioning still carry bang utilities. Those fight *internal* rules (not consumer CSS), are regression-sensitive, and deserve their own audited pass rather than riding along here.


nhobes:
Second commit extends the audit to the rest of the file per Nic's call: the `has-error` family, scheme-dropdown positioning, checkbox label, and stray bang utilities in scanned markup (including a doc comment Tailwind was compiling into a real `!important` utility). Every de-banged declaration verified in headless Chrome against the compiled CSS - 10/10 cases including the focused-input specificity case, plus an icon-contract regression guard. The changelog P2 (absorbed overscroll entry) is fixed in the same commit. `default.css` now contains exactly two `!important`s, both load-bearing and commented: the mdex inline-style fight and reduced-motion.

PR #534, opened 2026-08-04
fix(showcase): preview panel drops z-10 - it trapped overlay examples under host chrome

nhobes:
## What

Removes `z-10` from `.pc-showcase__preview` and documents why the rule must never get a z-index back.

## The bug (live on petal.build now)

A z-index on a positioned element creates a **stacking context**. The modal and slide-over docs examples render inside the preview panel, so their `fixed z-50` overlays were flattened to an effective `z-10` against the page. petal.build's docs sidebar (`fixed z-20`) and navbar therefore painted **over** an open slide over, and the dark overlay skipped them. Born on 2026-07-30 when the marketing docs moved 19 pages onto the showcase registry; invisible in the playground because its only fixed chrome is a mobile-only drawer.

Raising the overlay's z-index can never fix this - z only competes within its own context. Verified: an overlay forced to `z-9999` still loses to the z-20 sidebar while the trap exists.

## Why removal is safe

- The `z-10` had no layering job: preview and code panels are non-overlapping vertical siblings (it was carried over from the original marketing implementation in the registry's Phase-0 commit).
- No negative z-index exists anywhere in the stylesheet that could have depended on the context (swept).
- The command palette was never affected - its native `<dialog>` lives in the top layer, which no ancestor stacking context can trap. (That immunity is the argument for eventually moving modal/slide-over to the top layer too - logged as a follow-up, not this PR.)

## Scope

Docs surfaces only. Consumer apps never render showcase frames (`@source not` even excludes the modules from their builds), and `modal`/`slide_over` markup+CSS are untouched. petal_pro (4.6.2) has zero showcase usage - unaffected now and after upgrading.

## Verification

- Headless-Chrome stacking suite against the real compiled stylesheet, using `elementFromPoint` as the paint oracle: **before** - sidebar over overlay (prod bug reproduced), z-9999 overlay still loses; **after** - overlay covers sidebar and page content, 5/5 pass.
- `mix test`: 720 tests, 0 failures.

Intended release: **4.10.1** (4.10.0 is already on Hex). petal_marketing gets an interim one-line override (`.pc-showcase__preview { z-index: auto; }`) so prod is right before the bump - separate PR there, deletable after it.


nhobes:
Second commit: Nic found post-deploy that the trigger button still jumped 8px left on open/close even after the stacking fix. Different bug, same neighborhood: the `.pc-modal` / `.pc-slide-over` ROOT is an in-flow div, so revealing it makes it a real flex item inside the centered, gapped showcase preview - one extra `gap` re-centers the row. (Ruled out the scrollbar: only flex siblings moved, and the modal page had no scrollbar at all.) Both roots are now `fixed z-50` - out of flow, zero layout contribution, and the overlay/panel were always viewport-fixed so appearance is unchanged. This one IS consumer-visible in the good sense: any app with a trigger + overlay in a gapped flex row had the same jump. Headless-verified against compiled CSS: 8.0px → 0.0px, wrapper placement identical; 720 tests green. Changelog documents the one nuance (custom CSS positioned against the root's old `relative`).

PR #595, opened 2026-08-11
feat: border_plasma effect - pulse and rotate border glow (4.14.0)

nhobes:
A new effect for the family: a glowing border that either breathes in place or sweeps a conic gradient around the ring. Aimed at CTAs, pricing cards and hero panels.

## License / clean-room statement

Two reference projects were named. They got different treatment, on purpose.

- **[border-beam](https://github.com/Jakubantalik/border-beam) - MIT ("Copyright (c) 2026 Jakub Antalik"), verified by fetching `LICENSE` directly.** Permissive, so I read its README and prop surface. What I actually took from it is *design*, not code: the pulse/rotate split, the idea of an intensity dial, and the confirmation that pulse is the one worth defaulting to. It is a React component with its own DOM and its own CSS; none of it was copied. Its "rotate" is a travelling beam, which is what `border_beam` already does here, so our rotate is a genuinely different thing (a conic sweep of the whole ring).
- **[transitions.dev](https://github.com/Jakubantalik/transitions.dev) / organic-shimmer - NO LICENSE. `LICENSE` returns 404 and no license is declared anywhere.** Per the brief I did **not** read a single line of its source. Everything shimmer-adjacent here is clean-room from the one-line visual description on its public detail page ("wavy shimmer with phase-locked edge glow") and nothing else. In practice it contributed almost nothing, since I did not build a shimmer.

The implementation is built on `shine_border`'s existing padding-box mask trick and `border_beam`'s registered-custom-property workaround, both already in this repo.

## Naming

Picked **`border_plasma`**. The other two candidates both collide with something we already ship:

- `border_glow` collides semantically with `border_beam`'s existing `glow` boolean attr. `<.border_beam glow>` next to `<.border_glow>` in the same docs is a support question waiting to happen.
- `border_aura` collides with the `aurora` component. The Effects nav would read "Aurora" and "Border aura" two rows apart.

`border_plasma` has no collision, and "plasma" is honest about a luminous field lighting the whole ring rather than a point of light travelling along it - which is exactly the distinction from `border_beam`.

## Modes

`mode="pulse"` (**default**) - the whole ring is a closed conic loop (first colour repeated at both ends so there is no visible seam) and the brightness breathes. Nothing travels, so it pulls the eye without dragging it across the page. Default duration 4s.

`mode="rotate"` - the gradient becomes an arc, colour across roughly a third of the ring and transparent for the rest, with the whole thing spun by an animated angle. Busier, so it earns its place on one hero panel rather than every card. Default duration 6s.

Both are demoed in the playground with live mode / intensity / duration / width toggles, plus a static side-by-side of the two modes and a `Buy now` CTA at `intensity="strong"`.

## Reduced motion

Handled by the `@media (prefers-reduced-motion: no-preference)` idiom - the animation is never started rather than being switched off after the fact.

The important part is what the static state looks like. `--pc-plasma-t` is registered with `initial-value: 1`, so an unanimated plasma sits at **full brightness**. Reduced-motion users get a deliberate, fully lit gradient ring - not a dead border and not a dimmed one. Verified with `--force-prefers-reduced-motion` (screenshot below).

## CSS-only vs JS

100% CSS. No hook, nothing to add to `hooks:`, nothing exported from `petal_components.js`.

Two registered properties do the work: `--pc-plasma-angle` (`<angle>`) for the sweep, and `--pc-plasma-t` (`<number>`) for the breath. Registering the angle is what makes it animatable at all - an unregistered custom property is untyped, so the browser can only step it discretely and the sweep would judder round in jumps. The intensity maths lives on the elements rather than inside the keyframes, because Chrome will not resolve inherited custom properties inside `@keyframes` - the same constraint `border_beam` already documents.

## Theme colours

This one **diverges from the rest of the effects family** and I want it flagged rather than buried. `border_beam`, `meteors` and `spotlight_card` all default to hard-coded hexes (`#ffaa40`, `#9c40ff`, ...) on the theory that decorative colour is author-chosen. `border_plasma` instead defaults to `--color-primary-500` / `--color-secondary-500`, so it inherits a consumer's brand with no configuration. `color_from` / `color_to` still accept literals.

The component emits **no** colour custom properties unless the author passed one, so the CSS fallback to the theme token is what wins. That is also why one colour can be overridden while the other stays on the theme.

## Testing

- `mix test` - 904 tests, 0 failures (12 new). Includes the showcase registry guard and the escaped-colon CSS guard.
- `npm test` - 157 tests, 0 failures.
- `mix format --check-formatted` - clean.

## Screenshots

Playground at `/c/border-plasma`, headless Chrome at 2x. Pulse on a card in dark mode, light mode, both modes side by side, the strong CTA, and the reduced-motion static state.

## Open questions / taste calls for review

1. **Glow spread.** My first pass masked the bloom to the same hairline as the crisp ring, and blurring a 2px line by 10px diluted it to nothing - it read as a plain gradient border. The bloom now rides a ring twice the border width so there is enough light in the line for the blur to spend. Louder than v1, still restrained. Happy to dial either way.
2. **Default intensity is `medium`.** `strong` is genuinely bright on a dark hero; `subtle` is close to `shine_border` in weight. Is `medium` the right default, or should the default be quieter and let people opt up?
3. **Default `border_width` is `2px`**, not `1px` like `border_beam` and `shine_border`. A 1px plasma is very faint because there is so little ring to light. Inconsistent with its siblings, but I think it is the right call for this effect - worth a second opinion.
4. **Pulse breathes brightness only** - no scale, no blur animation. Scaling a masked ring distorts its thickness and animated blur is expensive, so brightness was the honest lever. It reads as a breath, but it is a subtler breath than the reference's outward-blooming halo.
5. **Version bump to 4.14.0 is included** in this branch (mix.exs + CHANGELOG). Easy to drop if you would rather this rides along with other work.
6. **Not added to `rules.md`** - that file does not currently enumerate the effects components at all, so I left it alone rather than starting a partial list. Say the word if it should go in.

Do not merge on the score - the brief was explicitly that Nic eyeballs the visual first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,29 @@
 # Changelog
+### 4.14.0 - 2026-08-11
+
+#### Added
+
+- **New `border_plasma` effect: a glowing border that breathes in place,
+  or sweeps a conic gradient around the ring.** `border_beam` sends one
+  travelling light around the edge, which reads as motion crossing the
+  screen. Plasma lights the whole ring instead, so it pulls the eye
+  without dragging it - the treatment you want on a pricing card or a
+  hero CTA that sits next to body copy. `mode="pulse"` is the default
+  breath; `mode="rotate"` spins a conic arc for something busier. Three
+  `intensity` steps (subtle, medium, strong) set how hard the ring burns
+  and how far the bloom carries, and `spread` overrides the bloom
+  distance on its own. Unlike the rest of the effects family, the
+  default colours come off the theme rail (`--color-primary-500` and
+  `--color-secondary-500`) rather than hard-coded hexes, so a plasma
+  picks up a consumer's brand without being told - `color_from` and
+  `color_to` still take literals when you want them. Pure CSS: a
+  registered `--pc-plasma-angle` drives the sweep and a registered
+  `--pc-plasma-t` drives the breath, both masked to the border ring with
+  the padding-box trick. No hook, nothing to wire up. Reduced motion is
+  handled by never starting the animation rather than hiding the effect,
+  and the breath rests at full brightness, so those users get a static
+  lit ring instead of a dead border.
+
 ### 4.13.0 - 2026-08-11
 
 #### Fixed
--- a/README.md
+++ b/README.md
@@ -228,6 +228,7 @@ Petal Components stands on other people's work:
 - [Tailwind CSS](https://tailwindcss.com/) - every component is styled with Tailwind v4.
 - [Apache ECharts](https://echarts.apache.org/) - the engine under `<.chart>`.
 - [Heroicons](https://heroicons.com/) - the icon set under `<.icon>`.
+- [border-beam](https://github.com/Jakubantalik/border-beam) by [Jakub Antalik](https://github.com/Jakubantalik) (MIT) - `<.border_plasma>`'s pulse effect ports its gradient geometry, oscillator tuning and layer treatment.
 
 ## License
 
--- a/assets/default.css
+++ b/assets/default.css
@@ -5485,14 +5485,17 @@
        light reads as whispery warping lines, not blobs. Each stop's alpha
        rides its QUADRANT's opacity oscillator, so whole corners of light
        swell and die independently. */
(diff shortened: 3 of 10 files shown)


nhobes:
Reworked after Nic's review: the first pass was structurally shine_border (uniform conic on a masked ring). Rebuilt as the reference actually behaves - four radial light POCKETS parked around the border (primary/secondary + two oklch color-mix in-betweens), breathing on four unequal-duration oscillators so the phases drift, one slow shared hue wander, and a new plus-lighter `__inner` layer spilling light over the panel edge through a radial vignette mask. Rotate keeps its sweeping arc, now carried by all three layers. Still pure CSS, reduced motion still rests fully lit and still. 904 Elixir / 157 JS green. Verified against beam.jakubantalik.com in dark + light, neutral + violet/cyan themes.

PR #625, opened 2026-08-12
feat(hover_card): hover card component

### LINKED ISSUE #602: Component: Hover Card
## Summary

A net-new `<.hover_card>` component: a hover/focus-triggered rich preview panel anchored to a trigger element. Think the GitHub profile-preview pattern - hover a username chip and a card with avatar, stats and a follow button fades in after a short delay. It sits between our existing `<.tooltip>` (text-only, non-interactive) and `<.popover>` (click-triggered): the panel opens on hover or keyboard focus like a tooltip, but its content is fully interactive (links, buttons), so it must stay open while the pointer is over the panel itself.

CSS-first, riding the machinery we already have: the tooltip's `group-hover`/`group-focus-within` reveal system and the popover's floating-panel surface and 12-way placement grid.

## References

- reui.io "Hover Card" (8 examples) - study the delay feel, the profile-card scenarios, and the hover-bridging behaviour
- Radix UI `HoverCard` - study `openDelay`/`closeDelay` defaults, placement API, and the rule that the card is for sighted-pointer enrichment (content must not be the only way to reach the information)
- GitHub's profile preview card - the canonical real-world scenario

We build a native HEEx/CSS equivalent from scratch. Reference their behaviour and variants, never their code.

## Prior art in this repo (study before writing code)

- `lib/petal_components/tooltip.ex` - the pure-CSS hover/focus reveal: `group/pc-tooltip` wrapper, `group-hover:`/`group-focus-within:` visibility, opacity transition with `delay-75`, placement modifier classes. The hover card generalises this pattern to an interactive panel.
- `lib/petal_components/popover.ex` - the 12-placement API (`@placements ~w(top top-start ... right-end)`), trigger/panel structure, and doc-string conventions.
- `assets/default.css` popover section (`.pc-popover__panel` and its 12 `--<placement>` modifiers) - the floating-panel surface (`bg-white dark:bg-gray-900`, real borders both modes, `shadow-lg`, `--pc-radius` consumption) and the exact placement offsets. Reuse the same offset geometry for `.pc-hover-card__panel--*`.
- `assets/default.css` tooltip section (`.pc-tooltip__content`) - the visibility + opacity + `transition-delay` reveal mechanics.

## API sketch

New module `lib/petal_components/hover_card.ex`, `use Phoenix.Component`, exported via `use PetalComponents` (no name collision - nothing exports `hover_card/1` today; verify against `lib/petal_components.ex` and core_components as usual).

```elixir
attr :id, :string, doc: "the panel id; autogenerated if not set"

attr :placement, :string,
  default: "bottom",
  values: ~w(top top-start top-end bottom bottom-start bottom-end
             left left-start left-end right right-start right-end),
  doc: "which side of the trigger the card appears on, with optional -start/-end alignment"

attr :open_delay, :integer,
  default: 350,
  doc: "milliseconds the pointer must rest on the trigger before the card opens"

attr :close_delay, :integer,
  default: 150,
  doc: "milliseconds after the pointer leaves before the card closes"

attr :class, :any, default: nil, doc: "extra classes for the wrapper"
attr :trigger_class, :any, default: nil, doc: "extra classes for the trigger element"
attr :card_class, :any, default: nil, doc: "extra classes for the card panel"
attr :rest, :global

slot :trigger, required: true, doc: "the trigger content (a link, chip, avatar, plain text)"
slot :inner_block, required: true, doc: "the card content"
```

Delays are emitted as CSS custom properties on the wrapper (`style="--pc-hover-card-open-delay: 350ms; --pc-hover-card-close-delay: 150ms;"`) and consumed by `transition-delay` in `default.css`. The trick: on the hidden-to-visible transition apply the open delay, on visible-to-hidden apply the close delay (set `transition-delay` to the close delay at rest, and to the open delay under the `:hover`/`:focus-within` state). Because `visibility` is transitionable in steps, delay both `visibility` and `opacity` so the panel cannot be interacted with mid-fade-out after close.

Unlike popover, the trigger is not a `<button>` - hover cards decorate content that is usually already a link or inline element. Render the trigger slot inside a focusable wrapper only if the slot content is not itself focusable is NOT something CSS can detect, so keep it simple: render the slot as-is inside a `span.pc-hover-card__trigger` and document that the trigger content should be focusable (a link, a button) for the keyboard path to work, mirroring how tooltip documents `aria-describedby` wiring.

### Hover bridging and interactivity

The panel must stay open while the pointer travels from trigger to panel and while it rests on the panel. Structure the DOM like tooltip (panel is a child of the `group` wrapper) so `group-hover` covers both trigger and panel; the placement offsets use margin (`mt-2` etc., as popover does), which keeps a hoverable gap - if the margin gap breaks the hover chain in testing, bridge it with a transparent `::before` strip on the panel covering the offset zone (a known CSS hover-card technique). Do not reach for a hook for this.

A hook is allowed only if CSS genuinely cannot deliver the delay/bridging contract (precedent: `PetalPopover` in `assets/js/petal_components.js`). Start CSS-only; if a hook proves necessary, flag it in the PR with the specific failure it solves.

## Variants & states to build

- [ ] All 12 placements: `top`, `top-start`, `top-end`, `bottom`, `bottom-start`, `bottom-end`, `left`, `left-start`, `left-end`, `right`, `right-start`, `right-end`
- [ ] Closed (rest) state: panel invisible, `pointer-events` disabled, not in the a11y tree beyond its container
- [ ] Open via pointer hover, after `open_delay`
- [ ] Open via keyboard focus (`focus-within` on the wrapper - focus a link inside the trigger and the card opens; Tab into the card's own links keeps it open)
- [ ] Stays open while pointer is over the panel; closes `close_delay` after the pointer leaves both trigger and panel
- [ ] Custom delays via `open_delay`/`close_delay` attrs
- [ ] Panel surface: floating-panel treatment identical to popover (`bg-white dark:bg-gray-900`, border both modes, `shadow-lg`, radius `min(var(--pc-radius, 0.625rem), 1rem)`)
- [ ] Reduced motion: with `prefers-reduced-motion`, drop the fade (instant show/hide) but KEEP the open/close delays - the delay is interaction design, not decoration
- [ ] Dark mode across all of the above

## Playground page

Add a nav entry and page in `dev.exs` following the existing component pages.

- Placement - all 12 values
- Open delay - e.g. 0 / 350 / 700 ms
- Close delay - e.g. 0 / 150 / 500 ms

Example scenarios (shadcn-grade, no lorem ipsum):

1. **User profile chip** - an `@username` chip with a small `<.avatar>`; hovering opens a profile card with avatar, name, bio line, follower/following stats, and a follow `<.button>`. This is the hero example and proves the interactive-content requirement (the button must be clickable).
2. **Link preview** - an inline text link inside a sentence; hovering opens a page-preview card (favicon/title/description/URL), demonstrating the pattern on plain prose.

Demonstrate keyboard use on the page: a caption noting that tabbing to the trigger link opens the card and tabbing onward reaches the card's controls.

## Accessibility

There is no dedicated WAI-ARIA hover-card pattern; follow Radix's model plus the tooltip conventions already in this repo:

- The card content must be supplementary. Document in the moduledoc that the trigger itself must link to the full information (e.g. the profile page) - the hover card is enrichment for pointer/keyboard users, not the only path.
- Keyboard: opening on `focus-within` is mandatory. Focus order: trigger, then the card's interactive children while open. No focus trap (this is not a dialog). Escape support: if achievable without a hook (it is not, in pure CSS), document the limitation honestly in the moduledoc rather than half-shipping it; moving focus away closes the card, which is the primary keyboard close path.
- Touch devices have no hover: the trigger must remain a functioning link/button on tap. Do not intercept taps to open the card.
- The panel gets an autogenerated `id`; decorative layers (arrow, if added later) get `aria-hidden="true"`. Do not use `role="tooltip"` (content is interactive) and do not use `role="dialog"` (it does not take focus modally); a plain container is correct here.
- `prefers-reduced-motion`: rest state complete and legible, per the tooltip/popover precedent (see the reduced-motion block near the end of `assets/default.css`).

## Tests

`test/petal/hover_card_test.exs` using `ComponentCase` + LazyHTML:

- [ ] Renders wrapper, trigger slot and panel with `pc-hover-card`, `pc-hover-card__trigger`, `pc-hover-card__panel` classes
- [ ] Autogenerated panel `id`; explicit `id` respected
- [ ] Each of the 12 `placement` values emits the matching `pc-hover-card__panel--<placement>` class; invalid values raise (attr `values:` enforcement)
- [ ] `open_delay`/`close_delay` emit the CSS custom properties with the given millisecond values; defaults are 350/150
- [ ] `class`, `trigger_class`, `card_class` merge onto the right elements
- [ ] `rest` passthrough lands on the wrapper
- [ ] Interactive content inside `inner_block` (a link) renders inside the panel
- [ ] Panel does not carry `role="tooltip"` or `role="dialog"`

## Non-goals

- No top-layer/`popover`-attribute mode in v1 (popover's `top_layer` escape hatch exists if a user needs unclipped rendering; a hover-card top-layer mode can follow if demand shows up)
- No collision detection / auto-flip
- No arrow in v1 (tooltip has one; add later if wanted)
- No controlled open state (`open` attr, server-driven show/hide)
- No Alpine.js variant - LiveView.JS + CSS only
- Not a replacement for tooltip (text hints) or popover (click-to-open panels); the moduledoc should include a short "which one do I want" comparison

## Process

1. Build on a feature branch: `lib/petal_components/hover_card.ex` + a Hover Card section in `assets/default.css` + tests + playground page (nav entry, dials, both example scenarios). CSS lives in this repo - no external packages, no new hex deps.
2. Verify in the playground (`mix run dev.exs`, http://localhost:4000): every placement dial, both delay dials, light + dark, keyboard-only (Tab to open, Tab through card content), touch emulation (tap must follow the trigger link, not trap), and `prefers-reduced-motion`.
3. Run the suites: `mix test` (and `npm test` only if a hook ended up being necessary). All green before PR.

## Definition of done

- `<.hover_card>` component with the API above (deviations flagged in the PR)
- `default.css` section following the popover surface + placement conventions
- Tests green, covering every attr, placement and the delay custom properties
- Playground page with placement + delay dials and the profile-chip and link-preview scenarios
- PR screenshots, light + dark
- Docs-ready: every attr has a `doc:` string; moduledoc includes usage examples, the tooltip/popover/hover-card comparison, the keyboard/touch behaviour notes, and the "trigger must link to the full information" guidance

mplatts:
Closes #602

## Summary

`<.hover_card>` - an interactive preview panel that opens when the pointer rests on its trigger, or when keyboard focus lands inside it. It fills the gap between the two components we already ship: `tooltip` is a text hint that opens on hover, `popover` is an interactive panel that opens on click. The hover card is an interactive panel that opens on hover or focus, after a delay.

Pure CSS. No hook, no timers, nothing to clean up on unmount.

```heex
<.hover_card placement="bottom" open_delay={350} close_delay={150}>
  <:trigger>
    <.link navigate={~p"/users/jane"} class="font-medium underline">@jane</.link>
  </:trigger>
  <%!-- avatar, bio, stats, a follow button --%>
</.hover_card>
```

**None.** Every attr, default and slot matches the sketch exactly: `id`, `placement` (12 values, default `bottom`), `open_delay` (350), `close_delay` (150), `class`, `trigger_class`, `card_class`, `rest`, plus the required `:trigger` and `:inner_block` slots. The delays are emitted as `--pc-hover-card-open-delay` / `--pc-hover-card-close-delay` on the wrapper and consumed by `transition-delay`, as specified.

One implementation detail worth flagging, because it is not in the sketch: a caller-supplied `style` in `@rest` is **merged after** ours rather than emitted as a second `style` attribute (a browser would silently ignore the duplicate). So `style="--pc-radius: 0px"` still works, and still wins over anything we set. There is a test for it.

## Anchoring: reused, not reinvented

The hover card reuses both existing mechanisms rather than adding a third:

- **Positioning** is the popover's. `.pc-hover-card__panel--<placement>` is a copy of `.pc-popover__panel--<placement>`, offset for offset, for all 12 placements. A card and a popover anchored the same way land in the same place, and the CSS says so in a comment.
- **Reveal** is the tooltip's. `group/pc-hover-card` on the wrapper, `group-hover:` / `group-focus-within:` flipping `visibility` and `opacity` on the panel.
- **Surface** is the popover's: `bg-white dark:bg-gray-900`, real borders in both modes, `shadow-lg`, `min(var(--pc-radius, 0.625rem), 1rem)`.

The `PetalPopover` hook and its viewport/flip/collision machinery were deliberately **not** pulled in. That hook exists for top-layer mode - escaping `overflow: hidden` ancestors and tracking the visual viewport for a panel painted outside the page. This component has no top-layer mode (a #602 non-goal), and no collision detection or auto-flip (also a non-goal), so there is nothing for it to do here. If a top-layer hover card is ever wanted, the hook is right there and this component can grow a `top_layer` attr the same way popover has one.

## Delays, bridging, keyboard, touch

**Delays.** `transition-delay` at rest is the close delay; under `:hover` / `:focus-within` it becomes the open delay. That single swap is the whole timing mechanism. `visibility` is in the transition property list alongside `opacity`: because it interpolates in steps, on the way out the panel stays `visible` for the full delay *and* the fade and then flips hidden in one go, so it can never be half-faded and still clickable.

**Pointer bridging.** The panel is a child of the group wrapper (the tooltip structure), so hovering the panel keeps the whole thing open - that is what makes the content genuinely interactive. The placement offsets (`mt-2` etc.) leave a real 8px gap, and crossing it did drop the hover in testing, so the panel carries a transparent `-inset-2` `::before` behind itself that spans the gap. It inherits the panel's visibility, so it is never a hit target while the card is closed.

**Keyboard.** Opens on `focus-within`. Tabbing to the trigger link opens the card, tabbing again reaches the card's own controls, tabbing past it closes. No focus trap - this is not a dialog. **Escape is not wired**, because in pure CSS it cannot be. The moduledoc says that plainly and points at `popover` if you need it, rather than half-shipping it.

**Touch.** Nothing intercepts the tap. The trigger stays the link or button it is, which is the whole reason the moduledoc leads with "the card is enrichment, never the only path to the information" - the trigger must reach the full content.

**Screen readers.** At rest the panel is `visibility: hidden`, so it is out of the accessibility tree entirely. It carries no `role`: not `tooltip` (the content is interactive), not `dialog` (it never takes focus modally). Tested.

**Reduced motion.** The fade goes, the delays stay. The delay is interaction design; without it every glancing pointer fires a card.

## One gotcha, documented

The wrapper, trigger and panel are all `<span>`s so a card drops into a sentence cleanly - but the HTML parser closes an open `<p>` the instant it meets a `<div>`, so block-level card content inside a paragraph gets hoisted out and renders unstyled. This bit the link-preview showcase example during browser verification. Fixed there, and written up in the moduledoc with a broken/fine pair.

## Tests

| suite | before | after |
|---|---|---|
| `mix test` | 913 passing, 1 skipped | **929 passing, 1 skipped** |
| `npm test` | 157 passing | **157 passing** (unchanged - no hook shipped, so no new JS test) |

`mix format --check-formatted` clean. `mix credo` unchanged from baseline (same exit code, and the only new entry is the repo-wide `@moduledoc` convention that `tooltip`, `popover` and `input_group` all share).

16 new component tests cover: wrapper/trigger/panel classes and nesting, autogenerated and explicit `id`, all 12 placement modifier classes, the `values:` constraint and every attr carrying a `doc:`, default and custom delay custom properties, caller-`style` merging, the three class attrs landing on the right elements, `rest` passthrough, interactive content inside the panel, absence of `role="tooltip"` / `role="dialog"`, and absence of any JS binding.

## Browser verification

Playground at `/c/hover-card`, 1280x1000, light and dark:

- opens after the open delay, not before (a short dwell genuinely does not fire it)
- pointer parked in the 8px gap: trigger no longer hovered, card still open - the bridge holds
- pointer on the panel body: still open
- pointer away: closes
- focus the trigger link: opens; Tab once: focus lands on the Follow button inside the card and it stays open
- `left` placement dial: card re-anchors mid-LiveView-patch without closing
- dark mode: panel reads correctly against the page

Screenshots to follow on the PR.

## Files

- `lib/petal_components/hover_card.ex` (new)
- `lib/petal_components/showcase/hover_card.ex` (new) - profile preview, link preview in prose, placement/timing
- `test/petal/hover_card_test.exs` (new)
- `assets/default.css` - appended Hover Card section
- `dev.exs` - nav entry, state, `ctl_hover_card` handler, snippet, `/c/hover-card` page with placement + both delay dials
- `lib/petal_components.ex`, `lib/petal_components/showcase/registry.ex`, `CHANGELOG.md`


mplatts:
## Screenshots

Playground page at `/c/hover-card`, captured at 1280px wide.

**Card open** — pointer rested on the `@jane` trigger past the 350ms open delay. The gap between trigger and panel is the bridged zone.

Hover card open, light mode

**Light, at rest**

Hover card at rest, light mode

**Dark**

Hover card, dark mode

<sub>Images live on the `pr-assets` branch, which exists only to host PR screenshots — it never merges to `main` and ships in no Hex release.</sub>

nhobes:
Eye-test fix: the demo was clipping top-* placements — the preview frame's `overflow-hidden` was doing its job correctly, the preview was just 15px too short for a top-placed card (220px card + 8px gap vs 212px of headroom). New commit:

- Preview grows `min-h-[28rem]` → `min-h-[34rem]` (33px slack each side at desktop); below `md` the trigger parks on the side the card grows away from instead of centering (a centered trigger needs ~753px for side placements); demo card is `w-56 sm:w-64` so it fits a 375px frame.
- Measured with the panel actually open per placement: **clipped placements 12→0 at 375 / 640 / 768 / 1280**, tightest margin 14.8px (375, left-*).
- Bonus real find: this was the only playground page with a pannable main pane at 375px (`main.scrollWidth` 442 vs 375) — the `:placement` showcase example's `right-start` card reached x=442. Fixed (`w-full` on the row; `justify-between` alone was a no-op on a shrink-to-fit flex item). Page scrollWidth was and stays 375 — #648's fence holds.
- Moduledoc gained the one honest sentence it lacked: placement is static, pick a side with room. (Auto-flip/collision remains the separately-tracked opt-in enhancement shared with nav_menu.)

One taste call deliberately left open (details in the eye-test notes): the `link_preview` showcase example still loses ~70px off its **left** edge at 375 — its `w-72` card is `placement="top"` on a trigger near the left edge in wrapped prose. Left overflow creates no scroll so the fence holds; the two one-line fixes (`top-start`, or `w-36 sm:w-72`) both change how the example renders on petal.build, so that's the maintainer's pick.

`mix test` 929/0 exit 0; format clean.

PR #626, opened 2026-08-12
feat(kbd,separator,collapsible): micro primitives for everyday parity

### LINKED ISSUE #604: Component: Kbd, Separator & Collapsible (micro primitives)
## Summary

Three tiny primitives that finish everyday parity with shadcn/reui, shipped together in one PR because each is under a day of work: `<.kbd>` (keyboard-key chip), `<.separator>` (semantic divider with optional label), and `<.collapsible>` (a single disclosure region, distinct from the grouped accordion). All three are pure HEEx + CSS + `Phoenix.LiveView.JS` with no hooks and no new dependencies. The library already has partial versions of the first two hiding in other components (a `.pc-kbd` CSS class, `hr/1` in typography); this issue promotes them to real, documented components.

## References

Study behaviour and variants only. We build native HEEx/LiveView equivalents from scratch.

- Kbd: https://reui.io/components/kbd (6 examples), https://ui.shadcn.com/docs/components/kbd
- Separator: https://reui.io/components/separator (6 examples), https://www.radix-ui.com/primitives/docs/components/separator
- Collapsible: https://reui.io/components/collapsible (10 examples), https://www.radix-ui.com/primitives/docs/components/collapsible

Existing in-repo prior art to study before writing anything:

- `lib/petal_components/typography.ex` — the inline-code chip family (`inline_code/1`, line ~213) that `<.kbd>` must visually align with, and `hr/1` (line ~256) which `<.separator>` supersedes for app UI without breaking it.
- `assets/default.css` line ~1475 — a `.pc-kbd` rule already exists (`:is(.pc-kbd, .pc-input-group__addon kbd)`), used by the command palette trigger (`lib/petal_components/command.ex` line ~189, the `kbd` attr on the trigger) and input-group addons. The new component adopts and extends this class rather than inventing a parallel one.
- `lib/petal_components/accordion.ex` — the JS toggle wiring (`js_attributes/8`, `aria-expanded`/`aria-controls`, `compose_js/2` for user-supplied `on_toggle` JS) and its a11y fix history in git log. Collapsible follows the same patterns for a single region.
- `lib/petal_components/helpers.ex` — `uniq_id/2` for generated ids, `compose_js/2` for merging user JS.

## API sketch

Starting proposal; refine in the PR and flag deviations.

### 1. `<.kbd>` — `lib/petal_components/kbd.ex`

```elixir
attr :class, :any, default: nil, doc: "CSS class"
attr :size, :string, default: "md", values: ["sm", "md"], doc: "chip size"
attr :keys, :list, default: nil,
  doc: ~s|renders a key sequence with separator glyphs, e.g. keys={["cmd", "K"]}. Known names ("cmd", "shift", "alt", "ctrl", "enter", "esc", "tab", "backspace", arrows) map to their symbols; unknown strings render verbatim|
attr :separator, :string, default: "+", doc: "glyph between keys in a sequence"
attr :rest, :global
slot :inner_block, required: false, doc: "single-key content when keys is not used"
```

```heex
<.kbd>K</.kbd>
<.kbd keys={["cmd", "K"]} />
<.kbd keys={["ctrl", "shift", "P"]} separator="then" size="sm" />
```

- Renders semantic `<kbd>` elements. A sequence is a wrapping span containing one `<kbd>` per key with the separator glyph between them (separator is `aria-hidden="true"`).
- Styling: mono/`font-sans` per the existing `.pc-kbd` rule, key-cap surface with subtle border and a bottom edge (slightly heavier bottom border or inset shadow so it reads as a physical key). Extend the existing `.pc-kbd` rule in `assets/default.css` so command trigger and input-group chips pick up the improved cap treatment for free; verify neither regresses.
- Symbol map lives in a module attribute, private, documented in the moduledoc.

### 2. `<.separator>` — `lib/petal_components/separator.ex`

```elixir
attr :class, :any, default: nil, doc: "CSS class"
attr :orientation, :string, default: "horizontal", values: ["horizontal", "vertical"]
attr :label, :string, default: nil, doc: "optional inline label, e.g. \"OR\" or a date"
attr :label_position, :string, default: "center", values: ["start", "center", "end"]
attr :decorative, :boolean, default: true,
  doc: "true renders aria-hidden (purely visual); false renders role=\"separator\" with aria-orientation for semantically meaningful dividers"
attr :rest, :global
slot :inner_block, required: false, doc: "rich label content; wins over the label attr"
```

```heex
<.separator />
<.separator label="OR" />
<.separator orientation="vertical" class="h-6" />
<.separator decorative={false} />
```

- Horizontal without label: a single hairline div on the gray ramp (match `pc-hr` colours: `border-gray-200 dark:border-gray-800`, but without `pc-hr`'s prose-oriented `my-8`; app-UI separators carry no default margin, callers add spacing).
- With label: flex row, hairlines either side via pseudo-elements or flanking divs (see `.pc-chat__marker--separator` in `assets/default.css` line ~4805 for the existing in-repo pattern), label in muted small text. `label_position` shifts the label by collapsing one flank.
- Vertical: `self-stretch` hairline; caller controls height. Labelled vertical is out of scope.
- `decorative` default true matches Radix. When false: `role="separator"` plus `aria-orientation="vertical"` when vertical (horizontal is the ARIA default, omit it).
- Do not touch `hr/1` in typography.ex; it stays for prose. Note the relationship in both moduledocs.

### 3. `<.collapsible>` — `lib/petal_components/collapsible.ex`

```elixir
attr :id, :string, default: nil, doc: "container id; autogenerated when omitted"
attr :class, :any, default: nil, doc: "CSS class"
attr :open, :boolean, default: false, doc: "render expanded; server-controlled open state"
attr :on_toggle, JS, default: %JS{},
  doc: "additional JS commands composed onto the toggle (see compose_js/2 in helpers.ex)"
attr :rest, :global

slot :trigger, required: true, doc: "the always-visible row; rendered inside a button" do
  attr :class, :any
end
slot :inner_block, required: true, doc: "the collapsible content"
```

```heex
<.collapsible open={@show_details}>
  <:trigger>API keys (3)</:trigger>
  <ul>...</ul>
</.collapsible>
```

- Trigger renders as `<button type="button">` with a trailing chevron (`hero-chevron-down-solid`, rotating 180deg when open, matching accordion's icon treatment).
- Toggle is client-side via `JS.toggle_attribute`/class toggle on the container plus `JS.set_attribute` for `aria-expanded`; compose the user's `on_toggle` with `compose_js/2`. `open` attr sets the initial/server-driven state, so LiveView can also control it by re-rendering with a changed assign.
- Animation: CSS `grid-template-rows: 0fr -> 1fr` on a grid wrapper with `overflow: hidden` on the inner row (the modern height-auto animation). Use `interpolate-size: allow-keywords` / `height: auto` transitions where supported as progressive enhancement; the grid-rows approach is the baseline and must degrade to instant show/hide under `prefers-reduced-motion`.
- A11y: `aria-expanded` on the trigger, `aria-controls` pointing at the content region id, content gets an id derived from container id (follow `content_panel_id/2` naming in accordion.ex). Study accordion's a11y fix history (`git log -p lib/petal_components/accordion.ex`) before wiring this; don't repeat bugs already fixed there.
- Explicitly distinct from accordion: one region, no exclusivity, no entries list. Do not refactor accordion in this PR.

All three modules: `use Phoenix.Component`, exported via `lib/petal_components.ex`. Check for collisions before naming — nothing may clash with existing `use PetalComponents` exports or `Phoenix.Component`/core_components names (`kbd`, `separator`, `collapsible` are clear at time of writing; re-verify).

## Variants & states to build

Kbd:
- [ ] Sizes `sm` / `md`
- [ ] Single key (slot content) and sequence (`keys` list) forms
- [ ] Symbol mapping for known modifier/named keys; verbatim fallback
- [ ] Custom `separator` glyph
- [ ] Light + dark, and visual parity with `pc-inline-code` chips and existing `.pc-kbd` consumers (command trigger, input-group addon)

Separator:
- [ ] Horizontal, plain
- [ ] Horizontal with label at `start` / `center` / `end`
- [ ] Vertical (unlabelled)
- [ ] `decorative` true (aria-hidden) and false (`role="separator"`, `aria-orientation`)
- [ ] Label via attr and via slot

Collapsible:
- [ ] Closed and open initial states (`open` attr)
- [ ] Chevron rotation synced to state
- [ ] Disabled trigger
- [ ] Animated expand/collapse (grid-rows baseline; `interpolate-size` enhancement)
- [ ] Reduced-motion: instant toggle, rest states fully legible
- [ ] `on_toggle` composition

## Playground page

One "Primitives" playground page (nav entry) with a section per component, following the existing dial pattern (segmented controls via `handle_event`) in `dev.exs`:

- Kbd section: size dial, separator glyph dial; examples: a command palette hint row (`⌘K`), a shortcuts cheat-sheet table (dogfood `<.kbd>` in the command palette trigger docs example), a menu item with trailing shortcut.
- Separator section: orientation dial, label-position dial, decorative toggle; examples: an "OR" divider between a login form and OAuth buttons, a date separator in an activity feed, a vertical separator in a toolbar between button groups.
- Collapsible section: open toggle, disabled toggle; examples: a "Show advanced options" block in a settings form, a changelog entry with collapsed details, a sidebar "Starred projects" section. Demonstrate keyboard toggling in the copy.

Standard applies: shadcn-grade demos, real product moments, no lorem ipsum. Exercise light + dark.

## Accessibility

- Kbd: semantic `<kbd>` elements; sequence separators `aria-hidden="true"`. No interactive behaviour.
- Separator: WAI-ARIA separator pattern (static, non-focusable variant). `decorative` true = `aria-hidden="true"` and no role; false = `role="separator"` plus `aria-orientation="vertical"` when vertical.
- Collapsible: WAI-ARIA disclosure pattern. Trigger is a real `<button>` (Enter/Space free), `aria-expanded` reflects state on every toggle path (client JS and server re-render), `aria-controls` references the content id. `disabled` uses the button's native disabled attribute. Focus stays on the trigger across toggles. `prefers-reduced-motion` disables the height animation.

## Tests

`test/petal/kbd_test.exs`, `test/petal/separator_test.exs`, `test/petal/collapsible_test.exs` using `ComponentCase` + LazyHTML:

- [ ] Kbd: renders `<kbd>` with `pc-kbd`; both sizes emit their class; `keys` list renders one `<kbd>` per key with `aria-hidden` separators between; symbol map (cmd -> ⌘ etc.) and verbatim fallback; custom separator; `class`/`rest` passthrough.
- [ ] Separator: horizontal and vertical classes; label attr and slot render; each `label_position` class; `decorative={true}` emits `aria-hidden="true"` and no role; `decorative={false}` emits `role="separator"`, and `aria-orientation="vertical"` only when vertical.
- [ ] Collapsible: trigger is `<button type="button">`; `aria-expanded="false"`/`"true"` matches `open`; `aria-controls` matches the content region id; content id derived from container id; autogenerated id when none passed; disabled renders a disabled button; chevron present with rotation class when open; `on_toggle` JS composed into `phx-click`.
- [ ] Every attr/variant/slot has at least one rendering assertion; a11y attributes asserted explicitly.

No JS hook, so no `npm test` additions expected.

## Non-goals

- No Alpine.js support (LiveView.JS + CSS only).
- No refactor of `accordion.ex` and no removal or restyling of typography's `hr/1` (separator supersedes it for app UI; hr stays for prose).
- No labelled vertical separators.
- No nested/exclusive collapsible groups (that is the accordion).
- No JS hook; if the animation genuinely cannot be done in CSS, stop and flag it on the issue rather than adding one.
- No changes to command palette or input-group markup beyond whatever the shared `.pc-kbd` rule improvement gives them for free.

## Process

1. Build on a feature branch: three component modules + `default.css` sections + tests + the Primitives playground page (nav entry, dials, examples). CSS lives in this repo; no external packages.
2. Verify in the playground (`mix run dev.exs`, http://localhost:4000): every dial, light + dark, keyboard-only operation of the collapsible, and `prefers-reduced-motion`. Also re-check the command palette trigger and input-group kbd addon after touching `.pc-kbd`.
3. Run `mix test`. All green before PR.
4. Open a PR against `main` with light + dark screenshots of the playground page. The automated review must pass.
6. Release (Hex publish, docs, changelog, MCP schema sync) is handled by maintainers after merge and is not part of this contribution.

## Definition of done

- `lib/petal_components/kbd.ex`, `separator.ex`, `collapsible.ex` exported from `lib/petal_components.ex` with no name collisions.
- `assets/default.css` sections for each (kbd extends the existing `.pc-kbd` rule; existing consumers verified unregressed).
- All three test files green; every attr documented with a `doc:` string; each moduledoc includes usage examples (docs-ready).
- Primitives playground page with dials and the example scenarios above, screenshotted light + dark on the PR.
- Collapsible passes a keyboard-only and reduced-motion walkthrough.

mplatts:
Closes #604

Three micro primitives in one PR, as the issue asked: `<.kbd>`, `<.separator>`, `<.collapsible>`. All pure HEEx + CSS + `Phoenix.LiveView.JS`. No hooks, no new dependencies, no Alpine.

### `<.kbd>` - `lib/petal_components/kbd.ex`

Semantic `<kbd>` chips. Single key through the default slot, or a sequence through `keys={["cmd", "K"]}` which renders one `<kbd>` per key with `aria-hidden` separator glyphs between them. A private symbol map folds known names to their glyph, case-insensitively, and anything unknown renders verbatim. Two sizes, `sm` and `md`.

The existing shared `.pc-kbd` rule was extended rather than forked, so the command palette trigger and input group addons pick up the key cap treatment for free. The treatment is an inset `box-shadow` border with a heavier bottom edge rather than a real `border`, specifically so existing consumers get the new look without their box size changing - the palette trigger and the search field addon both sit inside tight chrome where 2px would show.

### `<.separator>` - `lib/petal_components/separator.ex`

A hairline with no margin of its own. Optional label via the attr or the slot (slot wins), positioned `start` / `center` / `end` by collapsing one flank to a stub. Vertical mode is a `w-px self-stretch` rule; the caller sets the height. `decorative` defaults to `true` (aria-hidden, no role) to match Radix, and `decorative={false}` renders `role="separator"` plus `aria-orientation="vertical"` only when vertical.

`Typography.hr/1` is untouched. Both moduledocs now say which one to reach for.

### `<.collapsible>` - `lib/petal_components/collapsible.ex`

One disclosure region. The trigger is a real `<button type="button">` carrying `aria-expanded` and `aria-controls`; the panel is `role="region"`, labelled by the trigger, and `inert` while collapsed so it is out of the tab order and out of the a11y tree instead of being an invisible trap. `disabled` uses the native attribute. Chevron rotation is synced to state on both the server render and the client toggle.

Toggling is `JS.toggle_attribute` on `data-state` / `aria-expanded` / `inert` plus `JS.toggle_class` on the chevron, composed with the caller's `on_toggle` through `compose_js/2` (user commands run first). The `open` attr is the server-rendered state, so LiveView can drive it by re-rendering.

The height animation is `grid-template-rows: 0fr -> 1fr` on a grid wrapper with the inner row clipping, and it is dropped entirely under `prefers-reduced-motion: reduce`.

Four, all small:

1. **`esc` maps to `Esc`, not `⎋`.** Every other name in the map folds to a glyph. `⎋` is obscure enough that most people read it as a rendering failure, and shadcn/reui both print `Esc`. Same reasoning for `pageup` / `pagedown` -> `PgUp` / `PgDn`.
2. **No `interpolate-size: allow-keywords` progressive enhancement on the collapsible.** The issue floated it as an optional layer on top of the grid-rows baseline. Layering a `height: auto` transition over a `grid-template-rows` transition means two mechanisms animating the same box, and they fight on browsers that support both. The grid-rows approach already animates content of any height with nothing measured in JS, so the enhancement buys nothing and risks a double transition. Happy to add it if you disagree.
3. **The collapsible's JS command is public as `toggle_collapsible/2`, not `toggle/2`.** These modules are imported unqualified by `use PetalComponents`, and exporting `toggle/1,2` into every consumer's namespace is asking for a collision. The name is uglier; the collision isn't worth it.
4. **`<.separator>` ignores `label` when `orientation="vertical"`** rather than rendering a broken flex row. Labelled vertical separators are out of scope per the issue, so this makes the out-of-scope combination degrade to a plain vertical rule. Asserted in the tests.

## Implementation decisions worth a look

- **`inert` on the collapsed panel.** The alternative was `display: none`, which cannot animate, or leaving it visible-but-clipped, which leaves focusable content reachable inside a zero-height box. `inert` toggles instantly on click while the visual collapse animates, which is the right tradeoff.
- **`:id` is a declared attr defaulting to `nil`, so `assign_new/3` never fires.** The component tests the value instead. Worth knowing if you copy the pattern.
- **The new CSS section is wrapped in its own `@layer components { }`.** The tail of `assets/default.css` after line ~7015 is currently unlayered, which means those rules beat consumer utilities passed through `class`. Appending the new section inside a layer keeps `class` overrides working the way callers expect. Nothing existing was reflowed. The pre-existing unlayered tail is left alone - flagging it as a separate observation, not fixing it here.
- **Client toggle state survives server re-renders that do not change `open`.** LiveView only diffs changed dynamics, so flipping an unrelated assign will not stomp a client-side toggle. Documented in the moduledoc. It is also why the playground's `disabled` dial does not close an already-open region.

## Playground

New nav entry under Display: **Primitives** (`/c/primitives`). One page, a section per primitive, following the existing dial pattern.

- Kbd: size dial, separator glyph dial (`+`, `·`, `then`, none). Examples: a command palette hint row, the command trigger dogfooding the same chip, a shortcuts cheat sheet, a menu with trailing shortcuts.
- Separator: orientation dial, label position dial, decorative toggle. Examples: the OR divider between a login form and OAuth, a date row in an activity feed, a vertical rule in a toolbar.
- Collapsible: open and disabled dials. Examples: an advanced options block in a settings form, a changelog entry, an API keys list.

## Verification

Ran in the playground at `mix run dev.exs`, both themes:

- Every dial on all three sections.
- Collapsible toggled by mouse and by keyboard only (Tab to the trigger, Enter) - focus ring lands on the trigger, focus stays there across toggles, chevron rotates, content expands and collapses.
- Command palette trigger and input group `kbd` addon re-checked after the shared `.pc-kbd` change: both render at the same size with the new cap treatment, in light and dark.

Screenshots were captured locally during the walkthrough but are not attached here - happy to drop light/dark captures in a comment if you want them on the PR before review.

## Tests

| Suite | Before | After |
| --- | --- | --- |
| `mix test` | 913 tests, 0 failures, 1 skipped | **949 tests, 0 failures, 1 skipped** |
| `npm test` | 157 passing | **157 passing** (no JS hook, so nothing to add) |

36 new component tests across `test/petal/kbd_test.exs`, `separator_test.exs` and `collapsible_test.exs`. Every attr, variant and slot has at least one rendering assertion, and the a11y attributes are asserted explicitly.

`mix format --check-formatted` is clean. `mix credo` reports the same issues as `main` (verified by stashing) - nothing new in the added files.


mplatts:
## Screenshots

All three primitives live on one playground page, `/c/primitives`. Captured at 1280px wide; each pair below is that page scrolled to the relevant section.

### Kbd
Kbd, light
Kbd, dark

### Separator
Separator, light
Separator, dark

### Collapsible
`Advanced options` expanded by click; `Open on render` expanded by default.

Collapsible, light
Collapsible, dark

---

Two things surfaced while capturing these, both worth a decision before merge:

**1. The vertical separator is faint.** Two independent reviewers looked at the `Bold | Italic | Link | Code` toolbar and read the vertical rule the same way: legible, but a single-pixel gray-200 hairline against white that could be mistaken for whitespace on a quick scan. The horizontal variant gets away with gray-200 because its label and spacing reinforce it; the vertical one has nothing to lean on. Suggest bumping the vertical variant to gray-300.

**2. One combined page breaks the playground's 1:1 convention.** `main` has exactly 60 nav slugs and 60 `render_page/1` clauses - one page per component. This PR adds a single `primitives` slug covering three components, so `Kbd`, `Separator` and `Collapsible` don't appear in the playground nav by name and can't be linked to individually. The showcase registry does register all three separately, so petal.build docs will have three pages while the playground has one. Splitting into three slugs would restore the convention.

Neither is a correctness problem and both are quick to change - flagging rather than assuming.

mplatts:
## Both review points addressed

**1. Vertical separator darkened** (`98a16a0`) — gray-200 → gray-300 in light, gray-800 → gray-700 in dark, vertical variant only. The horizontal rule keeps gray-200: it sits in open layout with a label and whitespace reinforcing it, while the vertical one is a bare hairline wedged between toolbar buttons.

A second reviewer checked the result at 4× and made a point worth recording: the divider now sits *darker than the toolbar's own container border*, which is the relationship you want. At gray-200 the divider and the frame were competing at equal weight and the eye read the whole control as one undifferentiated pill. Their recommendation was to stop here rather than go louder, since a separator more prominent than the labels beside it would be worse.

**2. Split into one page per component** (`e8ef49b`) — the combined `primitives` slug is gone, replaced by `collapsible`, `kbd` and `separator`. Playground is back to a strict 1:1 mapping: **63 slugs, 63 `render_page/1` clauses**. Shared `@prim` state became per-component `kbd:` / `separator:` / `collapsible:` assigns and `ctl_prim` became `ctl_kbd` / `ctl_separator` / `ctl_collapsible`, matching the shape neighbouring single-component pages already use. Every example and dial carried over; nothing was trimmed.

All three pages were confirmed to render their own component rather than the silent `button` fallback, and each `ctl_*` handler was exercised rather than assumed: separator `orientation` flips to the vertical toolbar, `decorative` off puts `role="separator"` on the element, collapsible `open` flips `aria-expanded`, kbd `size`/`separator` round-trip.

### Verified independently

| Check | Result |
|---|---|
| `mix test` | 949 tests, 0 failures, 1 skipped |
| `npm test` | 157 passing |
| slugs : clauses | 63 : 63 |

## Screenshots (regenerated after the split)

### Kbd
Kbd, light
Kbd, dark

### Separator
Vertical rule at the new gray-300, in the `Bold | Italic | Link | Code` toolbar.

Separator, light
Separator, dark

### Collapsible
`Advanced options` expanded in both schemes.

Collapsible, light
Collapsible, dark

nhobes:
Polish pass done. Deviations for your sign-off when reviewing (all deliberate, none blocking):

- **Three pages instead of the brief's single /c/primitives page** - Matt made this call explicitly (head commit message says so). It matches the sidebar's one-page-per-family taxonomy, so I'd bless it; the PR description should just say so.
- **The brief's 'Starred projects sidebar' collapsible example** isn't in the playground page - fine to omit or add later; note it when closing #604.
- **Vertical separator renders gray-300/700**, one step darker than the brief's 'match pc-hr' - commented as deliberate in the CSS, just needs a line in the PR description's deviations list.
- Light + dark screenshots still pending on the PR body per process step 4.

PR #629, opened 2026-08-12
feat(number_field): a real spinbutton for quantities, prices and percentages

### LINKED ISSUE #607: Component: Number Field
## Summary

A net-new `<.number_field>` component: a real numeric spinbutton to replace bare `<input type="number">` in product UI. The native number input has inconsistent spinners across browsers, no styling hooks, and no support for big-step or hold-to-repeat behaviour. This component composes the input-group field surface with decrement/increment buttons, min/max/step clamping mirrored to ARIA, keyboard and wheel support, and light display formatting. It is the standard control for quantities, prices and percentages.

## References

Study these for scope and interaction design only. We build a native HEEx/LiveView equivalent from scratch; reference their behaviour and variants, never their code.

- reui.io "Number Field" (6 examples: default, stacked buttons, split buttons, min/max, precision, disabled)
- Base UI NumberField (https://base-ui.com/react/components/number-field) for the clamping, wheel and hold-to-repeat semantics
- WAI-ARIA spinbutton pattern: https://www.w3.org/WAI/ARIA/apg/patterns/spinbutton/

In this repo, study before building:

- `lib/petal_components/input_group.ex` - the anatomy to reuse: the group carries the field surface (border, radius, focus ring) and any petal input inside is stripped of its own surface. The number field is an input-group composition, not a new surface.
- `lib/petal_components/input_otp.ex` - the house precedent for a small hook wrapping one real input (`PetalInputOTP` in `assets/js/petal_components.js`). Follow this shape for the number-field hook.
- `lib/petal_components/field.ex` - attr naming and form integration conventions (`field`, `id`, `name`, `value`, `errors`, auto-generation from `%Phoenix.HTML.FormField{}`).
- `assets/default.css` - the `pc-input-group` section (around line 1415) for the surface/addon CSS this component builds on.

## API sketch

New module `lib/petal_components/number_field.ex`. Starting proposal; refine as needed and flag deviations in the PR.

```elixir
attr :id, :any, default: nil, doc: "input id; generated from the field or name if not passed"
attr :name, :any, doc: "input name; generated from the field if not passed"
attr :value, :any, doc: "current value; generated from the field if not passed"

attr :field, Phoenix.HTML.FormField,
  doc: "a form field struct, e.g. @form[:quantity]; sets id/name/value/errors like other inputs"

attr :min, :any, default: nil, doc: "lower bound; values are clamped and mirrored to aria-valuemin"
attr :max, :any, default: nil, doc: "upper bound; clamped and mirrored to aria-valuemax"
attr :step, :any, default: 1, doc: "increment for buttons, arrow keys and wheel"
attr :big_step, :any, default: nil, doc: "increment for shift+arrow; defaults to step * 10"

attr :precision, :integer, default: nil,
  doc: "decimal places shown on blur; raw value while editing. nil = no formatting"

attr :variant, :string, default: "stacked", values: ~w(stacked split plain),
  doc: "stacked: both buttons stacked at the inline end; split: minus at the start, plus at the end; plain: no buttons, keyboard/wheel only"

attr :size, :string, default: "md", values: ~w(sm md lg)
attr :rest, :global, include: ~w(autocomplete form placeholder readonly required inputmode)
```

Notes on the anatomy:

- One real `<input>` (type `text` with `inputmode="decimal"`, per the Base UI approach; `type="number"` fights custom formatting) inside the input-group surface. The buttons are the group's leading/trailing addons. Reuse the `pc-input-group` CSS contracts where possible; component-specific classes are `pc-number-field`, `pc-number-field__button`, etc.
- One small hook, `PetalNumberField`, exported from `assets/js/petal_components.js` alongside the others. It owns: clamping, stepping, hold-to-repeat on the buttons (pointer events with an initial delay then accelerating repeat), wheel handling while focused, keyboard map, `aria-valuenow` updates, and blur-time formatting. This clears the "CSS + LiveView.JS genuinely cannot do it" bar the same way `PetalInputOTP` does.
- Formatting stays light and dependency-free: `precision` rounds/pads on blur. The moduledoc should also document the `Intl.NumberFormat` pattern for currency/percent display (format the visible text on blur, keep the raw numeric value in the input while editing and in the posted form value). Do not add any dep, hex or npm, for formatting.
- Form integration: the `field` attr works exactly like other inputs, so `<.number_field field={@form[:qty]} />` works standalone. Additionally register a `"number-field"` type in `lib/petal_components/field.ex` (`@types` values list plus a clause) so `<.field type="number-field" field={@form[:qty]} min={1} />` gets label + errors for free, matching how other custom inputs are surfaced there.
- Check `lib/petal_components.ex` for name collisions before exporting `number_field/1`.

## Variants & states to build

- [ ] Variant `stacked`: chevron-up/chevron-down buttons stacked vertically at the inline end
- [ ] Variant `split`: minus button at the inline start, plus button at the inline end
- [ ] Variant `plain`: no buttons; keyboard, wheel and typing only
- [ ] Sizes `sm`, `md`, `lg` (input height, button hit areas, text size)
- [ ] Disabled (whole control, buttons non-interactive, surface uses the existing input-group disabled treatment)
- [ ] Buttons individually disabled at the bounds (value at min disables decrement, at max disables increment) with `aria-disabled`
- [ ] `precision` formatting on blur, raw while editing
- [ ] Error state via `<.field>` integration (rides `pc-form-field-wrapper--error .pc-input-group`, see `assets/default.css`)
- [ ] Dark mode for every state; focus ring on the group surface, `focus-visible` primary only

## Playground page

Add a nav entry and page in `dev.exs` following the existing component pages.

- variant: stacked / split / plain
- size: sm / md / lg
- min/max/step preset: e.g. "1 to 99, step 1" / "0 to 100, step 5" / "unbounded, step 0.5"
- disabled: on / off

1. Cart quantity: `split` variant, `min={1}`, integer steps, sitting in a line-item row with product name and price
2. Price input: `precision={2}`, a `$` leading addon, demonstrating the format-on-blur behaviour
3. Percentage allocation: `0..100`, `step={5}`, `%` trailing addon, shift+arrow jumping by 25

The page should demonstrate keyboard-only operation (arrows, shift+arrow, Home/End) in a caption or hint.

## Accessibility

Follow the WAI-ARIA spinbutton pattern:

- The input carries `role="spinbutton"`, `aria-valuemin`, `aria-valuemax` (when bounds are set) and `aria-valuenow`, kept in sync by the hook on every change
- Increment/decrement buttons are real `<button type="button">` elements with `aria-label` ("Increase value" / "Decrease value") and `tabindex="-1"` so the input remains the single tab stop
- Keyboard map: ArrowUp/ArrowDown step by `step`; Shift+Arrow steps by `big_step`; Home jumps to `min`, End to `max` (when set); typed input is clamped on blur, never mid-keystroke
- Wheel increments/decrements only while the input is focused, and calls `preventDefault` so the page does not scroll underneath
- Hold-to-repeat must also stop on `pointerleave`/`pointercancel`, not just `pointerup`
- Disabled state uses the native `disabled` attribute on input and buttons

## Tests

`test/petal/number_field_test.exs` using `ComponentCase` with LazyHTML assertions:

- [ ] Renders each variant with the expected structure (buttons present/absent, order for split)
- [ ] Each size and `disabled` render the expected classes/attributes
- [ ] `min`/`max` are mirrored to `aria-valuemin`/`aria-valuemax`; `value` to `aria-valuenow`; `role="spinbutton"` present
- [ ] Buttons have `aria-label` and `tabindex="-1"`
- [ ] `field` integration: name/id/value derived from a `%Phoenix.HTML.FormField{}`; the posted form name is correct
- [ ] `<.field type="number-field">` renders label and errors
- [ ] Data attributes the hook consumes (`min`, `max`, `step`, `big_step`, `precision`) are emitted correctly

JS hook tests in `npm test` for the pure logic (clamp, step, big step, precision rounding) if the hook factors those into testable functions, following the existing hook test layout.

## Non-goals

- No currency/locale formatting engine or masking library; `precision` plus the documented Intl pattern is the whole formatting story
- No scrubbing (drag-to-change) interaction
- No mobile-specific stepper redesign; the same anatomy works at touch sizes
- No changes to the existing `<.field type="number">` native input; it stays as-is
- No Alpine.js support; LiveView.JS + the hook only

## Process

1. Build on a feature branch: `lib/petal_components/number_field.ex` + `default.css` section + `PetalNumberField` hook in `assets/js/petal_components.js` + tests + playground page (nav entry, dials, the three scenarios). CSS/JS live in this repo; no external packages.
2. Verify in the playground (`mix run dev.exs`, http://localhost:4000): exercise every dial, light + dark, keyboard-only, wheel, hold-to-repeat, and `prefers-reduced-motion`.
6. Release (Hex publish, docs, changelog, MCP schema sync) is handled by maintainers after merge; not part of the contribution.

## Definition of done

- `<.number_field>` component with all three variants, three sizes, clamping, keyboard/wheel/hold-to-repeat, and ARIA wiring
- `pc-number-field` section in `assets/default.css` riding the input-group surface and the gray ramp
- `PetalNumberField` hook exported from the JS bundle
- `<.field type="number-field">` integration
- Tests green (`mix test`, `npm test`)
- Playground page with dials and the cart/price/percentage scenarios
- PR screenshots light + dark
- Docs-ready attr docs: every attr has a `doc:` string; the moduledoc includes usage examples and the Intl currency/percent pattern

Milestone: 4.16.0

mplatts:
Closes #607

## Summary

A net-new `<.number_field>`: a real spinbutton for quantities, prices and percentages, composed on the existing `pc-input-group` surface rather than standing up a new one. Three variants (`stacked`, `split`, `plain`), three sizes, clamping mirrored to ARIA, a full keyboard map, wheel-while-focused, and hold-to-repeat.

- `lib/petal_components/number_field.ex` - the component. `@moduledoc` carries the docs (why not `type="number"`, variants, keyboard table, a11y contract, the `Intl.NumberFormat` pattern for currency/percent); the public function's `@doc` points at it.
- `assets/js/petal_components.js` - `PetalNumberField` hook plus `numberFieldMath` (parse / clamp / decimal-safe add / format), exported so the specs can pin the rules without a DOM.
- `assets/default.css` - a `pc-number-field` section appended inside its own `@layer components { }` block, so it doesn't inherit the unlayered tail's habit of beating consumer utilities.
- `lib/petal_components/field.ex` - a `"number-field"` type: label, help text and the error tone for free.
- `dev.exs` - `/c/number-field` with live dials and the three scenarios from the issue.
- Showcase module, registry entry, `PetalComponents` import, changelog.

## The spinbutton vs native-input decision

**`<input type="text" inputmode="decimal">` carrying `role="spinbutton"`**, per the issue's sketch and the Base UI approach. Not `type="number"`. The reasons, in order of weight:

1. **`precision` is impossible on `type="number"`.** Writing `"24.50"` back into a number input is a value the browser normalises to `24.5` on read and re-renders however it likes. Format-on-blur is the component's whole formatting story, and the native type fights it.
2. **A half-typed value survives.** `type="number"` reports `value === ""` for `"12."` or `"1e"`, so any hook reading the input mid-keystroke sees an empty field. That makes "clamp on blur, never mid-keystroke" unimplementable.
3. **Spinners.** Unstyleable and different in Chrome, Firefox and Safari - the thing this component exists to replace.
4. `inputmode="decimal"` still gets the numeric keypad on mobile, which is most of what `type="number"` was buying.

What it costs: **native constraint validation**. `min`/`max` no longer participate in `:invalid` or in the browser's own form validation. The moduledoc says so in as many words and tells you to validate the bound server-side. `aria-valuemin` / `aria-valuemax` / `aria-valuenow` carry the semantics to assistive tech, which is what the APG spinbutton pattern asks for, and the hook enforces the clamp.

Bound buttons take `aria-disabled`, not `disabled` - a disabled button leaves the a11y tree, and a screen reader user tabbing the value up to 99 should still hear that an increment exists and is spent. A `disabled` *control* uses the native attribute everywhere, since then it genuinely isn't operable.

- **`:leading` / `:trailing` slots added.** The sketch had no slots, but two of its three required scenarios need a `$` and a `%` addon. They render into the input-group's own addon rows alongside the buttons.
- **`decrement_label` / `increment_label` attrs added.** The sketch hard-coded "Increase value" / "Decrease value". Those are the defaults; the cart scenario wants "Remove one seat".
- **`<.field type="number-field">` variant rides `number_variant`, not `variant`.** `variant` is already taken on `<.field>` by radio-card. This follows the existing `combo_variant` precedent rather than inventing a third convention.
- **No `errors` attr on `number_field/1`.** The sketch mentioned it. The error ring is painted by `.pc-form-field-wrapper--error .pc-input-group`, which is the wrapper's job - so `<.field>` renders the messages and the standalone component stays a control, matching `<.input>`.
- **No identity is a raise, not a generated id.** `<.number_field />` with no `field`, `id` or `name` raises `ArgumentError`. The hook needs a stable element id and the label's `for=` needs to name the input; minting a constant would collide on the second instance. Same call `combo_box` already makes.
- **`step` defaults to `1` and `big_step` to `step * 10`, as sketched** - but `big_step` also drives `PageUp`/`PageDown`, not only shift+arrow.
- **Wrapper id is `"#{id}-field"`.** The hook lives on the group, the input keeps the clean id so `for=` and any `aria-describedby` still work.

## Tests

| Suite | Before | After |
| --- | --- | --- |
| `mix test` | 913 tests, 0 failures, 1 skipped | 943 tests, 0 failures, 1 skipped |
| `npm test` | 157 passing | 195 passing |

30 new Elixir tests (anatomy per variant, sizes, hook data attrs, id derivation, ARIA, bounds, disabled/readonly, precision, `%Phoenix.HTML.FormField{}` integration, `<.field type="number-field">` including errors and size mapping) and 38 new JS tests (the pure maths, the keyboard map, wheel focused vs not, blur clamping and formatting, button presses, hold-to-repeat with fake timers including every stop path, ARIA sync across a patch).

`mix format --check-formatted` clean, `mix compile --force --warnings-as-errors` clean, `mix credo` identical to `main` (14 refactoring opportunities, 31 readability issues on both).

## Verified in the playground

Live on `/c/number-field`, light and dark: every dial, keyboard-only stepping (arrow, shift+arrow to 23, End to 99 with `aria-valuenow` following), the wheel, hold-to-repeat, `precision` turning `7.5` into `7.50` on tab-out, and the buttons still stepping after a variant switch rewires them mid-patch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,35 @@
 # Changelog
+### Unreleased
+
+#### Added
+
+- **New `number_field` component: a real spinbutton for quantities,
+  prices and percentages, in place of a bare
+  `<input type="number">`.** The native number input has spinners you
+  can't style, that differ in every browser, and a value the browser
+  sanitises out from under you the moment you try to format it. This one
+  is a text input carrying `role="spinbutton"` on the existing
+  `input_group` surface, so it inherits the border, radius, focus ring
+  and error tone the rest of the form family already uses, and the
+  steppers are ours to draw. Three variants: `stacked` chevrons at the
+  inline end, `split` minus and plus with the value centred - the cart
+  quantity look - and `plain` for keyboard only. `min`, `max` and `step`
+  clamp and mirror to `aria-valuemin` / `aria-valuemax` /
+  `aria-valuenow`; `big_step` (defaulting to ten steps) rides
+  shift+arrow and page up/down; Home and End jump to the bounds.
+  `precision` rounds and pads on blur while the raw text stands while
+  you type, so `7.5` becomes `7.50` only once you're done - and the
+  moduledoc documents the `Intl.NumberFormat` pattern for currency and
+  percent display, because no formatting dependency ships with this. The
+  `PetalNumberField` hook owns the stepping, the clamping, wheel while
+  focused, and hold-to-repeat that accelerates and stops on
+  `pointerleave` as well as `pointerup`. The input is the single tab
+  stop, the buttons are labelled at `tabindex="-1"`, and a button at its
+  bound takes `aria-disabled` rather than `disabled` so it keeps its
+  name for a screen reader. `<.field type="number-field">` wires it into
+  the label/help/error machinery, and `<.field type="number">` is
+  untouched.
+
 ### 4.14.0 - 2026-08-11
 
 #### Added
--- a/assets/default.css
+++ b/assets/default.css
@@ -7586,3 +7586,85 @@
       transition-duration: 1ms;
     }
   }
+
+/* Number field
+ *
+ * A spinbutton riding the input-group surface: the group keeps the border,
+ * radius and focus ring, so this section only paints the input density and
+ * the step buttons. Inside @layer components so a consumer's own utilities
+ * still outrank it.
+ */
+@layer components {
+  .pc-number-field .pc-number-field__input {
+    /* text-base below sm, like pc-text-input: iOS Safari zooms the viewport
+       when focusing an input under 16px, and a re-layout mid-interaction
+       makes the next tap land on stale coordinates. */
+    @apply w-full min-w-0 px-3 py-2 text-base sm:text-sm text-gray-900 tabular-nums placeholder:text-gray-400 focus:outline-none disabled:cursor-not-allowed dark:text-gray-100 dark:placeholder:text-gray-500;
+  }
+  .pc-number-field--sm .pc-number-field__input {
+    @apply px-2.5 py-1.5 sm:text-xs;
+  }
+  .pc-number-field--lg .pc-number-field__input {
+    @apply py-2.5 sm:text-base;
+  }
+  /* Split sits the value between two buttons, so centring it is what stops
+     the field reading as lopsided. */
+  .pc-number-field--split .pc-number-field__input {
+    @apply text-center;
+  }
+
+  /* Addons holding buttons shed the text inset and wear the same 2px inset
+     frame as an input-group button chip. */
+  .pc-number-field .pc-input-group__addon:has(.pc-number-field__button) {
+    @apply gap-1 p-0.5;
+  }
+  .pc-number-field .pc-input-group__addon--leading:has(.pc-number-field__button) {
+    @apply pl-0.5;
+  }
+  .pc-number-field .pc-input-group__addon--trailing:has(.pc-number-field__button) {
+    @apply pr-0.5;
+  }
+
+  .pc-number-field__button {
+    @apply flex items-center justify-center shrink-0 w-7 text-gray-500 cursor-pointer select-none transition-colors duration-150 hover:bg-gray-100 hover:text-gray-900 focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-primary-500/50 disabled:cursor-not-allowed dark:text-gray-400 dark:hover:bg-gray-400/17 dark:hover:text-gray-100;
+    border-radius: max(calc(var(--pc-radius, 0.625rem) - 0.125rem), 0px);
+  }
+  .pc-number-field--sm .pc-number-field__button {
+    @apply w-6;
+  }
+  .pc-number-field--lg .pc-number-field__button {
+    @apply w-8;
+  }
+  /* At a bound the button keeps its place and its label (aria-disabled, not
+     disabled) but reads as spent and stops answering the pointer. */
+  .pc-number-field__button[aria-disabled="true"] {
+    @apply text-gray-300 pointer-events-none dark:text-gray-600;
+  }
+  /* Doubled selector: the icon sizing contract, same as the combobox chevron. */
+  .pc-number-field__icon.pc-number-field__icon {
+    @apply w-4! h-4!;
+  }
+  .pc-number-field--sm .pc-number-field__icon.pc-number-field__icon {
+    @apply w-3.5! h-3.5!;
+  }
+
+  /* Stacked: two half-height buttons splitting the trailing addon's height,
+     the classic spinner. */
+  .pc-number-field__stack {
+    @apply flex flex-col self-stretch gap-px;
+  }
+  .pc-number-field__stack .pc-number-field__button {
+    @apply flex-1 min-h-0;
+  }
+  .pc-number-field__stack .pc-number-field__button--inc {
+    border-radius: max(calc(var(--pc-radius, 0.625rem) - 0.125rem), 0px)
+      max(calc(var(--pc-radius, 0.625rem) - 0.125rem), 0px) 0 0;
+  }
+  .pc-number-field__stack .pc-number-field__button--dec {
+    border-radius: 0 0 max(calc(var(--pc-radius, 0.625rem) - 0.125rem), 0px)
+      max(calc(var(--pc-radius, 0.625rem) - 0.125rem), 0px);
+  }
+  .pc-number-field--split .pc-number-field__button {
+    @apply self-stretch;
+  }
+}
--- a/assets/js/petal_components.js
+++ b/assets/js/petal_components.js
@@ -1435,6 +1435,263 @@ export const PetalInputOTP = {
   },
 };
 
+// Number field maths, kept pure and exported so the specs can pin the rules
+// without a DOM: parsing, clamping, decimal-safe stepping and blur formatting.
+export const numberFieldMath = {
+  // "" and "abc" are both "no number yet" - a half-typed field must not
+  // resolve to 0, or every keystroke would fight the user.
+  parse(text) {
+    if (text === null || text === undefined) return null;
+    const trimmed = String(text).trim();
+    if (trimmed === "") return null;
+    const n = Number(trimmed);
+    return Number.isFinite(n) ? n : null;
+  },
+
+  clamp(n, min, max) {
+    if (n === null) return null;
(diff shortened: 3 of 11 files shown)


mplatts:
## Screenshots

Playground page at `/c/number-field`, 1280px wide, full page.

**Light**

Number field, light mode

**Dark**

Number field, dark mode

### Verified independently

Re-ran every check on a clean worktree off the pushed branch:

| Check | Result |
|---|---|
| `mix test` | 943 tests, 0 failures, 1 skipped (+30) |
| `npm test` | 195 passing (+38) |
| `mix credo` | zero new entries vs `main` (45 before, 45 after; the `field.ex` complexity finding just shifts line 249 → 255 as the new clause pushes it down) |
| new dependencies | none — `mix.exs` and `package.json` untouched |

**The `field.ex` change is worth a careful look** since that file is on every form in every consumer app. It is strictly additive: `"number-field"` joins the allowed `type` list, a `number_variant` attr is added, the `:global` include list gains `big_step precision decrement_label increment_label inputmode`, and a new `def field(%{type: "number-field"})` clause is appended. No existing clause or type is modified.

The `type="text" inputmode="decimal"` + `role="spinbutton"` decision (rather than native `type="number"`) is well argued in the PR body — native normalises `"24.50"` away, making `precision` unimplementable, and reports `value === ""` for half-typed input like `"12."`, making "clamp on blur, never mid-keystroke" unimplementable. The tradeoff is losing native constraint validation, which the moduledoc calls out.

PR #630, opened 2026-08-12
feat(scroll_area): themed native scrollbars, one div, zero JS

### LINKED ISSUE #609: Component: Scroll Area
**Milestone: 4.16.0**

## Summary

A net-new `<.scroll_area>` component: a themed scroll container that makes overflow regions look intentional instead of inheriting whatever the OS default scrollbar happens to be. It is CSS-first and zero JS - modern `scrollbar-width` + `scrollbar-color` (Chrome 121+, Firefox) with `::-webkit-scrollbar` fallback styling, a thin themed thumb riding the gray ramp in both light and dark mode, an optional stable gutter (`scrollbar-gutter: stable`), and optional fade-edge masks that hint at clipped content.

The component itself is small. The real value is consistency: the library already has at least eight hand-rolled scroll regions (command palette list, combo box list, dropdown menu, slide-over content, chat thread, chat code blocks, nav menu panel, multi-select) each styling overflow slightly differently. `<.scroll_area>` gives them one set of `pc-scroll-area` classes to adopt, so every scrollable surface in a Petal app looks like it came from the same design system. The issue includes an adoption checklist as follow-through.

## References

Study these for scope and interaction design only - we build a native HEEx/CSS equivalent from scratch, never port their code:

- reui.io "Scroll Area" (5 examples: vertical, horizontal, both, themed thumb, in-panel usage)
- Radix UI ScrollArea (https://www.radix-ui.com/primitives/docs/components/scroll-area) - note that Radix reimplements scrollbars in JS with custom thumb elements; we deliberately do NOT. We theme the native scrollbar with CSS and accept the platform's behaviour. That trade is the point: zero JS, native momentum/keyboard/AT behaviour for free.
- shadcn/ui scroll-area (wraps Radix; same caveat)

## API sketch

New module `lib/petal_components/scroll_area.ex`, `use Phoenix.Component`, exported via `use PetalComponents` (check `lib/petal_components.ex` for name collisions first - `scroll_area` is currently free).

```elixir
attr :orientation, :string,
  default: "vertical",
  values: ["vertical", "horizontal", "both"],
  doc: "which axis scrolls: vertical (overflow-y), horizontal (overflow-x), or both"

attr :fade_edges, :boolean,
  default: false,
  doc:
    "fade content out at the scroll edges with a mask-image gradient, hinting that more content exists beyond the clip. Masks apply on the scrolling axis/axes only"

attr :gutter_stable, :boolean,
  default: false,
  doc:
    "reserve scrollbar space with scrollbar-gutter: stable so content does not shift when the scrollbar appears or disappears (classic vertical scrollbars only; overlay scrollbars have no gutter)"

attr :visibility, :string,
  default: "auto",
  values: ["auto", "always"],
  doc:
    "auto follows the platform (overlay scrollbars appear on scroll on macOS); always requests a persistently visible scrollbar where the platform allows it - see the platform-truth note in the moduledoc"

attr :class, :any,
  default: nil,
  doc: "size the viewport here, e.g. class=\"max-h-72\" or class=\"max-w-full\" - sizing is deliberately class-driven, not attr-driven"

attr :rest, :global

slot :inner_block, required: true
```

Rendered markup is a single element - no wrapper soup:

```heex
<div
  class={[
    "pc-scroll-area",
    "pc-scroll-area--#{@orientation}",
    @fade_edges && "pc-scroll-area--fade",
    @gutter_stable && "pc-scroll-area--gutter-stable",
    @visibility == "always" && "pc-scroll-area--always",
    @class
  ]}
  tabindex="0"
  {@rest}
>
  {render_slot(@inner_block)}
</div>
```

Follow the attr/doc/class-list conventions in `lib/petal_components/skeleton.ex` (compact attrs, every attr has a `doc:` string, `@class` last in the list, `@rest` global).

### CSS sketch (`assets/default.css`, new section)

- `.pc-scroll-area` base: `scrollbar-width: thin` plus `scrollbar-color: <thumb> transparent` where the thumb rides the gray ramp (something like `--color-gray-300` light / a `gray-400` alpha dark, tuned by eye against the existing panel surfaces). `overscroll-behavior: contain`.
- `::-webkit-scrollbar` fallback block for engines that expose it: thin track (transparent), rounded thumb on the same gray values, hover state one step darker. Note: Chrome 121+ honours `scrollbar-color` and then ignores `::-webkit-scrollbar-*`, so the two mechanisms never fight - the fallback covers older Chromium and current Safari.
- `--vertical` / `--horizontal` / `--both` modifiers set `overflow-y: auto` / `overflow-x: auto` / `overflow: auto` (the off axis gets `hidden`).
- `--gutter-stable` adds `scrollbar-gutter: stable` (consider `stable both-edges` only if the playground shows asymmetry; default to plain `stable`).
- `--fade` applies `mask-image: linear-gradient(...)` on the scroll axis (both gradients composited via `mask-composite`/comma-separated masks when orientation is `both`). Static masks are acceptable for v1 - the fade shows at both edges regardless of scroll position. Scroll-position-aware fading (via `animation-timeline: scroll()` where supported, progressive enhancement) is a nice-to-have, not required; if attempted it must degrade to the static mask.
- `--always`: where the platform allows, keep the scrollbar persistently visible. On WebKit, styling `::-webkit-scrollbar` with an explicit size already opts out of overlay behaviour. On Firefox there is no force-visible mechanism. Document this honestly (below).
- Dark mode via `dark:` values on the same gray ramp - never invent new grays.

There is prior art in this file to stay consistent with: `.pc-command__list` (`scrollbar-width: thin`, ~line 2063) and `select[multiple]` (~line 1390) already do the thin-scrollbar treatment ad hoc. The new section becomes the canonical version they migrate to.

### The platform-truth caveat (must appear in the moduledoc)

Overlay scrollbars are OS-controlled. On macOS with "Show scroll bars: Automatically", the scrollbar only appears during scroll and ignores most theming; the same is true on most mobile browsers. `<.scroll_area>` themes what the platform exposes - it does not fight the OS. `visibility="always"` is a request, honoured where the engine allows it (WebKit via the `::-webkit-scrollbar` opt-out), silently ignored elsewhere. Write this in plain language in the moduledoc so users are not surprised, and repeat it as a note on the playground page.

## Variants & states to build

- [ ] Orientation: `vertical`, `horizontal`, `both`
- [ ] `fade_edges` on each orientation (single-axis mask for vertical/horizontal, composited for both)
- [ ] `gutter_stable` on and off (vertical; document that it is a no-op for overlay scrollbars)
- [ ] `visibility`: `auto` and `always`
- [ ] Light and dark thumb/track theming on the gray ramp
- [ ] Thumb hover state (one gray step darker)
- [ ] No-overflow state: content shorter than the viewport renders cleanly with no scrollbar, no gutter jump (unless `gutter_stable`), and no fade masking content that is fully visible is acceptable for v1 (static masks - see CSS sketch)

## Playground page

Nav entry + hero + dials, per the house pattern (segmented controls wired via `handle_event`).

Dials:
- Orientation (vertical / horizontal / both)
- Fade edges (on / off)
- Gutter (auto / stable)
- Visibility (auto / always)

1. **Tag list** - a horizontal scroll_area of `<.badge>` tags in a card, fade edges on, showing the "more tags this way" hint.
2. **Chat transcript** - a vertical scroll_area with fade edges wrapping a realistic message thread (borrow content style from the chat component's playground page).
3. **Code block** - `orientation="both"` around a wide, tall `<pre>` snippet, gutter stable, demonstrating two-axis theming.

Include the platform-truth note visibly on the page (macOS overlay scrollbars behave differently from Windows/Linux classic scrollbars).

## Accessibility

There is no WAI-ARIA pattern for a scroll container; the relevant guidance is keyboard operability of scrollable regions:

- The scroll area must be keyboard scrollable: `tabindex="0"` on the container so keyboard users can focus it and use arrow keys / PageUp / PageDown / Home / End (native behaviour once focusable - no JS).
- Because it is focusable, it gets the standard `focus-visible` ring in primary (never a persistent `:focus` fill). Consider making `tabindex` overridable via `@rest` for cases where an inner element is already focusable and a second tab stop would be noise - if so, document when to remove it.
- Give it an accessible name when it is a standalone region: document in the moduledoc that users should pass `aria-label` (through `@rest`) when the scroll area is a primary content region, e.g. `aria-label="Chat messages"`.
- Fade masks are decorative-only (pure CSS mask, no extra DOM) - nothing to hide from AT.
- No `prefers-reduced-motion` concerns for v1 (no animation). If the scroll-position-aware fade enhancement is attempted, it must not introduce motion beyond the user's own scrolling.

## Tests

`test/petal/scroll_area_test.exs` using `ComponentCase` + LazyHTML:

- [ ] Base render emits `pc-scroll-area` and default `pc-scroll-area--vertical`
- [ ] Each `orientation` value emits its modifier class (and only that one)
- [ ] `fade_edges` toggles `pc-scroll-area--fade`
- [ ] `gutter_stable` toggles `pc-scroll-area--gutter-stable`
- [ ] `visibility="always"` emits `pc-scroll-area--always`; `"auto"` emits nothing extra
- [ ] `class` merges (e.g. `max-h-72` present alongside `pc-*` classes)
- [ ] `tabindex="0"` present; `aria-label` passed through `@rest` lands on the container
- [ ] Inner block renders inside the container
- [ ] Invalid `orientation`/`visibility` values raise at compile/render time (attr `values` enforcement)

## Adoption checklist (follow-through)

The consistency payoff. After the component lands, migrate the existing ad-hoc scroll styling in `assets/default.css` to the shared `pc-scroll-area` treatment - either by applying the same custom-property/class recipe inside each component's CSS, or by extracting the thumb/track values into shared variables both consume. Each is a small, separately reviewable change (they can ride in the same PR as tick-box commits or follow-up PRs, builder's call with maintainer sign-off):

- [ ] `.pc-command__list` (command palette, ~line 2063) - already `scrollbar-width: thin`; align thumb colours
- [ ] `.pc-combo-box__list` (~line 2215)
- [ ] `.pc-dropdown__menu-items-wrapper` (~line 1962)
- [ ] `.pc-data-table__filter-options` (~line 2727)
- [ ] `.pc-slideover__content` (~line 4043)
- [ ] `.pc-chat__thread` (~line 4585) and `.pc-chat__markdown pre` (~line 4724)
- [ ] `.pc-nav-menu__panel` (~line 5062)
- [ ] `select[multiple]` in the input styles (~line 1390)

(Line numbers are current-main approximations; grep for the class names.)

## Non-goals

- No JS-reimplemented scrollbars (Radix-style custom thumb elements). Native scrollbars, themed - that is the whole design.
- No scroll-linked behaviours (scroll spy, infinite scroll, scroll restoration) - out of scope.
- No shadow-based overflow indicators as an alternative to masks - one mechanism (`fade_edges`) for v1.
- No forced overlay scrollbars where the OS shows classic ones, or vice versa - we do not fight the platform.
- No sizing attrs (`max_height` etc.) - sizing stays class-driven.

## Process

1. Build on a feature branch: `lib/petal_components/scroll_area.ex` + `assets/default.css` section + tests + playground page (nav entry + dials + the three examples). CSS lives in this repo - no external packages, no JS hook (this component is CSS-only by design).
2. Verify in the playground (`mix run dev.exs`, http://localhost:4000) - exercise every dial, light + dark, keyboard-only scrolling (Tab to the area, arrows/PageDown), and check on both a classic-scrollbar setup and macOS overlay scrollbars if available.
3. Run the suites: `mix test` (no `npm test` needed - no hook). All green before PR.
4. Open a PR against `main` with screenshots (light + dark) of the playground page; a Windows-or-Linux classic-scrollbar screenshot alongside a macOS one is a bonus since they render differently. The automated review must pass.

## Definition of done

- [ ] `scroll_area/1` component with every attr carrying a `doc:` string and a moduledoc with usage examples plus the platform-truth caveat
- [ ] `pc-scroll-area` section in `assets/default.css` (modern properties + webkit fallback, dark mode on the gray ramp)
- [ ] Tests green (`mix test`)
- [ ] Playground page with the four dials and three example scenarios, platform note included
- [ ] Adoption checklist addressed or explicitly deferred with maintainer sign-off
- [ ] PR screenshots (light + dark)

mplatts:
Closes #609

## What this is

`<.scroll_area>` - a themed scroll container, so an overflow region looks like it belongs to the design system instead of inheriting whatever the OS decided. One `<div>`, zero JavaScript.

- `scrollbar-width` + `scrollbar-color` do the work on Chrome 121+ and Firefox; a `::-webkit-scrollbar` block covers Safari and older Chromium. The two never fight - an engine that honours the modern properties ignores the webkit pseudo-elements outright.
- `orientation` is `vertical` / `horizontal` / `both`, with the off axis clipped rather than left scrollable by accident.
- `fade_edges` masks the scrolling axis with a `mask-image` gradient (both gradients composited via `mask-composite: intersect` when `orientation="both"`).
- `gutter_stable` adds `scrollbar-gutter: stable`.
- `visibility="always"` gives `::-webkit-scrollbar` an explicit size, which is the WebKit opt-out from overlay scrollbars.
- Thumb rides the gray ramp in light and dark, one step toward the foreground on hover. No new grays invented.
- Sizing stays class-driven. No `max_height` attr, no second sizing vocabulary next to Tailwind's.

## CSS vs a JS hook

**No hook.** Radix (and shadcn on top of it) reimplements the scrollbar in JS with custom thumb elements. We deliberately do not: native scrollbars mean native momentum scrolling, native keyboard handling and native assistive-tech behaviour, all for free and all of it hard to get right by hand.

The trade is that we theme what the platform exposes rather than overriding it, and that trade is written out in plain language in the moduledoc and repeated as a visible note on the playground page:

- `visibility="always"` is a request. WebKit honours it, Firefox has no force-visible mechanism, and no browser overrides an OS set to hide scrollbars.
- `gutter_stable` is a no-op wherever scrollbars are overlays (the macOS default), because an overlay has no gutter to reserve.
- `scrollbar-color` has no hover state, so the hover step only shows on engines taking the `::-webkit-scrollbar` path.

Nothing here needed JS to work, so nothing here ships JS. No `test/js/` file, because there is no hook to test.

1. **`role="region"` is conditional, not absent and not unconditional.** The sketch rendered `tabindex="0"` and no role; the issue's a11y section asked for the call to be made deliberately. The container renders `role="region"` only when `aria-label` or `aria-labelledby` is present. An unnamed region is an entry in the landmark list that says nothing, and an explicit `role` you pass always wins. Documented in the moduledoc, covered by four tests.
2. **`tabindex` is a defaulted global, not hard-coded.** The sketch hard-coded `tabindex="0"` on the element, which would have produced a duplicate attribute for anyone passing `tabindex="-1"` through `@rest`. It is `Map.put_new/3` on `@rest` instead, so the caller can remove the tab stop when everything inside is already focusable - the case the issue asked to consider.
3. **Scroll-position-aware fading not attempted.** The issue lists it as a nice-to-have. Static masks only, as the issue's own CSS sketch permits for v1. Getting it right needs `@property` registration plus `animation-timeline: scroll()` keyframes and a Firefox fallback path, and a half-tested progressive enhancement is worse than none.
4. **New caveat found and documented:** `mask-image` masks the whole element, border and background included, so a bordered scroll area with `fade_edges` watches its own border fade out. The moduledoc says to put the border on a wrapper, and both playground examples that combine the two do exactly that.
5. **Adoption checklist deferred.** The eight existing ad-hoc scroll regions are untouched here. They are each a separately reviewable change and the issue leaves the call to the maintainer - happy to do them in this PR or a follow-up, whichever @nhobes prefers.

## Playground

`/c/scroll-area`, in the Display group. Four live dials (orientation, fade edges, gutter, visibility) wired via `handle_event`, the platform-truth note, and three scenarios: a horizontal tag rail with faded edges, a chat transcript, and a two-axis deploy log with the gutter reserved.

Showcase module carries four engine-run examples for petal.build.

## Verification

| | before | after |
|---|---|---|
| `mix test` | 913 tests, 0 failures, 1 skipped | **938 tests, 0 failures, 1 skipped** |
| `npm test` | 157 passing | **157 passing** (no JS shipped) |
| `mix credo --all` | 45 entries | **45 entries** (zero new) |

`mix format --check-formatted` clean. `mix compile --force --warnings-as-errors` clean.

Screenshots of the playground page in light and dark, captured in headless Chromium launched with native scrollbars visible, to be attached by hand - they are not committed to the repo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,34 @@
 # Changelog
+### Unreleased
+
+#### Added
+
+- **New `scroll_area` component: one themed treatment for every overflow
+  region, in one div and zero JavaScript.** The library already had eight
+  hand-rolled scroll regions, each styling overflow slightly differently -
+  this is the canonical version they can all adopt. `scrollbar-width` and
+  `scrollbar-color` do the work on Chrome 121+ and Firefox, a
+  `::-webkit-scrollbar` block covers Safari and older Chromium, and the two
+  never fight because an engine that honours the modern properties ignores
+  the webkit pseudo-elements outright. The thumb rides the gray ramp in
+  light and dark and steps one shade toward the foreground on hover.
+  `orientation` picks the axis (vertical, horizontal or both, with the off
+  axis clipped rather than left scrollable by accident), `fade_edges` masks
+  the scrolling axis so content dissolves at the clip instead of being cut,
+  and `gutter_stable` reserves the scrollbar's space so content does not
+  shift the moment it appears. Sizing is deliberately class-driven, so
+  there is no second sizing vocabulary sitting next to Tailwind's. Unlike
+  the Radix-style approach we do not reimplement the scrollbar in JS: you
+  keep native momentum scrolling, native keyboard handling and native
+  assistive-tech behaviour for free, and the trade is that we theme what
+  the platform exposes rather than overriding it - `visibility="always"` is
+  a request WebKit honours and Firefox cannot, and `gutter_stable` is a
+  no-op wherever scrollbars are overlays. The moduledoc says so in plain
+  language. The container is focusable by default (`tabindex="0"`), which
+  is what makes arrow keys and Page Up/Down work at all, and it only takes
+  `role="region"` when you give it a name - an unnamed landmark is noise,
+  not a service.
+
 ### 4.14.0 - 2026-08-11
 
 #### Added
--- a/assets/default.css
+++ b/assets/default.css
@@ -7586,3 +7586,143 @@
       transition-duration: 1ms;
     }
   }
+
+/* Scroll area - one themed treatment for every overflow region.
+ *
+ * CSS only, by design. `scrollbar-width` + `scrollbar-color` do the work on
+ * Chrome 121+ and Firefox; the ::-webkit-scrollbar block below covers older
+ * Chromium and current Safari. The two never fight: an engine that honours
+ * `scrollbar-color` ignores the webkit pseudo-elements outright.
+ *
+ * Note the base deliberately sets NO ::-webkit-scrollbar width/height. Giving
+ * one is what opts a WebKit element out of overlay scrollbars, so it belongs
+ * to --always, not to every scroll area on the page.
+ *
+ * Layered on purpose - these are component defaults and a consumer's own
+ * utilities must be able to beat them. */
+@layer components {
+  .pc-scroll-area {
+    --pc-scroll-area-thumb: var(--color-gray-300);
+    --pc-scroll-area-thumb-hover: var(--color-gray-400);
+    --pc-scroll-area-size: 8px;
+    --pc-scroll-area-fade: 1.5rem;
+
+    scrollbar-width: thin;
+    scrollbar-color: var(--pc-scroll-area-thumb) transparent;
+    overscroll-behavior: contain;
+    @apply focus:outline-hidden focus-visible:ring-2 focus-visible:ring-primary-500/50;
+  }
+
+  .pc-scroll-area:where(.dark, .dark *) {
+    --pc-scroll-area-thumb: var(--color-gray-600);
+    --pc-scroll-area-thumb-hover: var(--color-gray-500);
+  }
+
+  /* Fallback path: Safari and Chromium < 121 */
+  .pc-scroll-area::-webkit-scrollbar-track,
+  .pc-scroll-area::-webkit-scrollbar-corner {
+    background: transparent;
+  }
+  .pc-scroll-area::-webkit-scrollbar-thumb {
+    background-color: var(--pc-scroll-area-thumb);
+    background-clip: padding-box;
+    border: 2px solid transparent;
+    border-radius: 9999px;
+  }
+  .pc-scroll-area::-webkit-scrollbar-thumb:hover {
+    background-color: var(--pc-scroll-area-thumb-hover);
+  }
+
+  /* Orientation - the off axis is clipped, never scrollable by accident */
+  .pc-scroll-area--vertical {
+    @apply overflow-x-hidden overflow-y-auto;
+  }
+  .pc-scroll-area--horizontal {
+    @apply overflow-x-auto overflow-y-hidden;
+  }
+  .pc-scroll-area--both {
+    @apply overflow-auto;
+  }
+
+  .pc-scroll-area--gutter-stable {
+    scrollbar-gutter: stable;
+  }
+
+  /* An explicit track size is the WebKit opt-out from overlay scrollbars.
+     Firefox has no equivalent, so there this is a no-op - documented. */
+  .pc-scroll-area--always::-webkit-scrollbar {
+    width: var(--pc-scroll-area-size);
+    height: var(--pc-scroll-area-size);
+  }
+
+  /* Fade edges - static masks on the scrolling axis. The mask is anchored to
+     the element's box, not to the content, so it stays put while you scroll. */
+  .pc-scroll-area--fade.pc-scroll-area--vertical {
+    -webkit-mask-image: linear-gradient(
+      to bottom,
+      transparent 0,
+      #000 var(--pc-scroll-area-fade),
+      #000 calc(100% - var(--pc-scroll-area-fade)),
+      transparent 100%
+    );
+    mask-image: linear-gradient(
+      to bottom,
+      transparent 0,
+      #000 var(--pc-scroll-area-fade),
+      #000 calc(100% - var(--pc-scroll-area-fade)),
+      transparent 100%
+    );
+  }
+
+  .pc-scroll-area--fade.pc-scroll-area--horizontal {
+    -webkit-mask-image: linear-gradient(
+      to right,
+      transparent 0,
+      #000 var(--pc-scroll-area-fade),
+      #000 calc(100% - var(--pc-scroll-area-fade)),
+      transparent 100%
+    );
+    mask-image: linear-gradient(
+      to right,
+      transparent 0,
+      #000 var(--pc-scroll-area-fade),
+      #000 calc(100% - var(--pc-scroll-area-fade)),
+      transparent 100%
+    );
+  }
+
+  .pc-scroll-area--fade.pc-scroll-area--both {
+    -webkit-mask-image:
+      linear-gradient(
+        to bottom,
+        transparent 0,
+        #000 var(--pc-scroll-area-fade),
+        #000 calc(100% - var(--pc-scroll-area-fade)),
+        transparent 100%
+      ),
+      linear-gradient(
+        to right,
+        transparent 0,
+        #000 var(--pc-scroll-area-fade),
+        #000 calc(100% - var(--pc-scroll-area-fade)),
+        transparent 100%
+      );
+    mask-image:
+      linear-gradient(
+        to bottom,
+        transparent 0,
+        #000 var(--pc-scroll-area-fade),
+        #000 calc(100% - var(--pc-scroll-area-fade)),
+        transparent 100%
+      ),
+      linear-gradient(
+        to right,
+        transparent 0,
+        #000 var(--pc-scroll-area-fade),
+        #000 calc(100% - var(--pc-scroll-area-fade)),
+        transparent 100%
+      );
+    -webkit-mask-composite: source-in;
+    mask-composite: intersect;
+  }
+}
--- a/dev.exs
+++ b/dev.exs
@@ -224,6 +224,18 @@ defmodule Dev.PlaygroundLive do
     }
   ]
 
+  @scroll_log """
+  09:14:01.118 [info]  release v4.15.0 building on builder-04 (elixir 1.19.1 / otp 28)
+  09:14:04.902 [info]  ==> deps compiled in 3.7s, 214 modules
+  09:14:11.470 [info]  ==> assets: tailwind 4.1.3 wrote priv/static/assets/app.css in 812ms
+  09:14:12.006 [info]  ==> digest: 46 files, 1.9MB total, gzip 412KB
+  09:14:19.331 [info]  image pushed: registry.fly.io/acme-prod:deployment-01K2 (sha256:9f21c4ab)
+  09:14:26.884 [info]  machine 5683d9 updating in syd, waiting for health checks
+  09:14:38.019 [info]  machine 5683d9 healthy after 11.1s, 1 of 2 machines updated
+  09:14:51.744 [info]  machine 90e8ff healthy after 9.4s, 2 of 2 machines updated
+  09:14:52.100 [info]  deployment complete, 0 failed, rollback not required\
+  """
+
   @nav [
     %{
       group: "Foundations",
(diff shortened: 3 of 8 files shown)


mplatts:
## Screenshots

Playground page at `/c/scroll-area`, 1280px wide, full page. Captured with `--hide-scrollbars false` so the themed scrollbars are actually visible rather than hidden by headless Chromium.

**Light**

Scroll area, light mode

**Dark**

Scroll area, dark mode

### Verified independently

| Check | Result |
|---|---|
| `mix test` | 938 tests, 0 failures, 1 skipped (+25) |
| `npm test` | 157 passing, unchanged (no JS shipped) |
| new dependencies | none |

Shipping zero JS here is the right call against the house rule that CSS and `Phoenix.LiveView.JS` come before a hook.

The `mask-image` finding is a good catch and worth keeping in the docs: the mask applies to the element's own border, so a bordered scroll area with `fade_edges` fades its own border away. Both playground demos that combine the two now put the border on a wrapper.

Noted for @nhobes: `visibility="auto"` vs `"always"` is indistinguishable in headless Chromium, which honours `scrollbar-color` and ignores the webkit path — so the `always` opt-out is verified by construction, not by screenshot. It needs a real Safari.

nhobes:
Polish pass done. Two calls are yours on review:

1. **The adoption checklist is deferred** - the issue's definition of done lists 8 ad-hoc scroll regions (command list, combo-box list, dropdown wrapper, data-table filter options, slideover content, chat thread, chat markdown pre, nav-menu panel, select[multiple]) that should adopt pc-scroll-area classes 'or be explicitly deferred with maintainer sign-off'. Recommend: follow-up PR after this merges (the adoption touches 8 shipped components and deserves its own diff). Say the word and I'll do the sweep.

2. **Dark thumb tone eye-test** - the thumb is solid gray-600/hover gray-500 where the house dark surfaces mostly ride gray-400 alphas. Check the deploy-log panel in dark mode; if it reads muddy, the knobs are the --pc-scroll-area-thumb dark values.

nhobes:
Two additions since the polish pass:

1. **Merged `main` in** (no rebase, so no history rewrite) — the branch now carries #648–#651: the mobile x-fence, both slide-over keyframe fixes, and the command palette scroll lock. The only conflict was both sides adding a CHANGELOG section under Unreleased; resolved by keeping both (Added, then Fixed).

2. **The playground now dogfoods the component.** The three playground chrome regions that actually scroll vertically — the mobile menu list, the desktop sidebar nav, and the main content pane — were still raw `overflow-y-auto` divs and now run through `<.scroll_area>`. Details worth an eye during review:
   - The two nav regions pass `tabindex="-1"` (everything inside is already a button, per the component's own moduledoc pattern), with the `<nav>` landmark moved inside so it survives.
   - The main pane passes `role="main"` to keep the landmark the `<main>` element carried, and #648's x-axis fence comes free since `pc-scroll-area--vertical` is `overflow-x-hidden overflow-y-auto` — re-measured at 375px on /c/border-plasma: main scrollWidth 436px, page scrollWidth 375px, unchanged.
   - Deliberately left alone: the demo rails and dial strip (scrollbars hidden on purpose), the View-code panels and table wrappers (horizontal-only), and scroll regions inside shipped components.

Suite: 938 tests, 0 failures. `mix format --check-formatted` now passes repo-wide on this branch.

PR #631, opened 2026-08-12
feat(scrollspy): the "on this page" rail

### LINKED ISSUE #613: Component: Scrollspy
# Component: Scrollspy

Target milestone: **4.17.0**

## Summary

A net-new `<.scrollspy>` component: scroll-position-aware navigation highlighting, the classic "On this page" rail you see beside long docs pages. A small `PetalScrollspy` hook watches sections via IntersectionObserver and toggles the active state on the matching nav link as the reader scrolls. The component ships two things and both must be documented: a ready-made vertical nav renderer with an animated active indicator bar, and a plain data-attribute contract so users can wire the hook to any markup of their own.

This belongs in the library because every docs/blog/marketing site rebuilds this by hand, it pairs naturally with `<.prose>`/typography content, and doing the edge cases correctly (topmost-wins, bottom snap, hash navigation) is exactly the fiddly work a component should absorb.

## References

- reui.io "Scrollspy" (2 examples) — study scope and interaction, not code
- shadcn-adjacent docs sites ("On this page" sidebars: Tailwind docs, Radix docs) for the visual pattern and the animated indicator bar
- Bootstrap Scrollspy docs for the behavioural edge cases (multiple visible sections, bottom of page)

We build a native HEEx/LiveView equivalent from scratch. Reference behaviour and variants only.

## API sketch

New module `lib/petal_components/scrollspy.ex`, `use Phoenix.Component`. Check `lib/petal_components.ex` for name collisions before exporting (`scrollspy` is currently free).

```elixir
attr :items, :list, required: true,
  doc: """
  The nav entries. Each item is a map: %{label: "Install", target: "install"}.
  `target` is the id of the section element (no leading #). One level of
  nesting is supported for h2/h3 structure via a `children` key:
  %{label: "Usage", target: "usage", children: [%{label: "Options", target: "options"}]}
  """
attr :offset, :string, default: "6rem",
  doc: "scroll-margin-top applied to targets on click-scroll; set to clear a fixed header"
attr :threshold, :string, default: nil,
  doc: "optional override for the observer rootMargin (activation line), e.g. \"-20% 0px -70% 0px\". Defaults to a reading-position tuned value baked into the hook"
attr :indicator, :string, default: "bar", values: ["bar", "none"],
  doc: "the animated active indicator: a bar sliding along the link list, or none"
attr :heading, :string, default: nil, doc: "optional small heading above the list, e.g. \"On this page\""
attr :class, :any, default: nil, doc: "extra classes for the nav element"
attr :rest, :global
```

Renders roughly:

```heex
<nav id={@id} class={["pc-scrollspy", @class]} phx-hook="PetalScrollspy"
     data-offset={@offset} data-threshold={@threshold} aria-label="On this page" {@rest}>
  <span :if={@heading} class="pc-scrollspy__heading">{@heading}</span>
  <div class="pc-scrollspy__indicator" aria-hidden="true"></div>
  <ul class="pc-scrollspy__list">
    <li>
      <a href="#install" class="pc-scrollspy-link" data-scrollspy-target="install">Install</a>
      <!-- nested children get pc-scrollspy-link--nested -->
    </li>
  </ul>
</nav>
```

Follow the attr style in `lib/petal_components/navigation_menu.ex` (which also has the `aria-current` precedent) for doc strings and defaults.

### Data-attribute contract (the second, markup-free API)

Users can put `phx-hook="PetalScrollspy"` on any container whose descendant links carry `data-scrollspy-target="<section-id>"`. The hook:

- observes each element whose id matches a `data-scrollspy-target`
- sets `aria-current="location"` and adds `pc-scrollspy-link--active` on the active link, removing both from the rest
- reads `data-offset` (scroll-margin-top value) and `data-threshold` (rootMargin override) off the hook element

Document this contract in the moduledoc as a first-class usage mode, with a bare-markup example.

### Hook: `PetalScrollspy`

Lives in `assets/js/petal_components.js` alongside the other exports (`PetalPopover`, `PetalInputOTP` are the precedents for "CSS genuinely cannot do this" hooks — an IntersectionObserver clearly qualifies). Behavioural spec:

- **IntersectionObserver** over the target sections, `rootMargin` tuned so activation happens at the reading position (roughly the top 25–30% of the viewport), overridable via `data-threshold`.
- **Topmost wins**: when multiple sections intersect, the one closest to the top of the viewport is active.
- **Bottom snap**: when the page is scrolled to the very bottom, the last section activates even if it never crosses the activation line (short final sections).
- **Hash navigation**: on mount and on `hashchange`, if `location.hash` matches a target, set it active immediately rather than waiting for observer callbacks.
- **Smooth scroll on click**: CSS `scroll-behavior: smooth` on the scroll container plus `scroll-margin-top` (from `data-offset`) applied to targets so fixed headers do not cover the heading. Respect `prefers-reduced-motion` (no smooth behaviour).
- **Indicator**: when the nav uses the built-in renderer with `indicator="bar"`, the hook positions `.pc-scrollspy__indicator` against the active link via `transform: translateY(...)` and a height match, transitioned in CSS. With reduced motion the bar still moves, just without the transition.
- Clean up the observer and listeners in `destroyed()`; re-scan targets in `updated()` (LiveView patches can replace content).

## Variants & states to build

- [ ] Flat item list (single level)
- [ ] Nested items (one level, indented, `pc-scrollspy-link--nested`)
- [ ] `indicator="bar"` — animated sliding bar
- [ ] `indicator="none"` — text/colour-only active state
- [ ] Optional `heading`
- [ ] Active link state: `pc-scrollspy-link--active` + `aria-current="location"`
- [ ] Hover state on links
- [ ] Dark mode for all of the above (ride the gray ramp, active link in primary; no invented grays)
- [ ] Bare-markup mode (hook on user's own list) working with zero `pc-scrollspy` structure

New `pc-scrollspy` section in `assets/default.css`. Links are text-first (muted gray, primary when active); `focus-visible` ring in primary only.

## Playground page

New playground page in `dev.exs` with a nav entry. The page itself is the demo: a long, doc-like article (real headings and prose, not lorem ipsum — write it as a mini guide) with the spy rail fixed alongside.

- [ ] `offset` — a few values so you can see the fixed-header clearance change
- [ ] nesting — flat vs h2/h3 nested items
- [ ] `indicator` — bar / none

Scenarios:

1. Hero: the docs "On this page" rail beside a long article, with the playground's own sticky header proving the offset works.
2. A bare-markup example wired purely through `data-scrollspy-target` attributes, to prove the contract.

Verify by scrolling top to bottom (topmost-wins in the overlap zones, last item active at the very bottom), clicking every link, and loading the page with a `#hash` in the URL.

## Accessibility

- The rail is a `<nav>` with `aria-label` (default "On this page").
- Active link gets `aria-current="location"` (the correct value for within-page position; `navigation_menu.ex` uses `aria-current` similarly for its active item).
- Links are plain `<a href="#id">` anchors, so keyboard operation is native: Tab through links, Enter follows the anchor. No roving tabindex, no custom keys.
- The indicator bar is decorative: `aria-hidden="true"`.
- `prefers-reduced-motion`: no smooth scrolling, no indicator transition; the rest state is complete and legible.

## Tests

`test/petal/scrollspy_test.exs` using `ComponentCase` + LazyHTML:

- [ ] Renders a `nav` with `phx-hook="PetalScrollspy"`, default `aria-label`, and the id
- [ ] Every item renders an anchor with `href="#<target>"` and `data-scrollspy-target="<target>"`
- [ ] No link has `aria-current` or the active class in the initial render (the hook owns active state)
- [ ] `data-offset` and `data-threshold` land on the hook element; `threshold` omitted when nil
- [ ] Nested children render with the nested class, one level only
- [ ] `indicator="bar"` renders the `aria-hidden` indicator element; `indicator="none"` omits it
- [ ] `heading` renders when given, absent otherwise

JS suite (`test/js/scrollspy.test.js`, vitest + jsdom like the existing `test/js/popover.test.js`): unit-test the observer selection logic, factored so it is testable without a real IntersectionObserver:

- [ ] Topmost-wins when multiple entries intersect
- [ ] Bottom-of-page snap activates the last section
- [ ] Hash on mount sets the matching link active immediately
- [ ] Active toggling sets/removes `aria-current="location"` and the active class exactly once across the link set

`mix test` and `npm test` both green.

## Non-goals

- No horizontal scrollspy (the reui horizontal example is out of scope for v1).
- No automatic TOC extraction from the DOM (users pass `items` or wire their own markup); could be a follow-up.
- No scroll container other than the window/nearest scrollable ancestor detection the observer gives us for free — no explicit `container` attr in v1.
- No more than one level of nesting.
- No URL hash rewriting while scrolling (history spam); clicking a link updates the hash natively, scrolling does not.

## Process

1. Build on a feature branch: `lib/petal_components/scrollspy.ex` + `pc-scrollspy` section in `assets/default.css` + `PetalScrollspy` in `assets/js/petal_components.js` + tests + playground page (nav entry + dials + examples). CSS/JS live in this repo — no external packages, no new hex deps.
2. Verify in the playground (`mix run dev.exs`, http://localhost:4000): exercise every dial, light + dark, keyboard-only, `prefers-reduced-motion`, hash-load, and the bottom-of-page case.
5. **@nhobes is the gatekeeper**: final design/quality checks happen on the PR; he merges when happy. Do not self-merge.

## Definition of done

- [ ] `<.scrollspy>` component with every attr carrying a `doc:` string; moduledoc includes usage examples for both the built-in renderer and the bare data-attribute contract
- [ ] `PetalScrollspy` hook handling topmost-wins, bottom snap, hash navigation, configurable activation line, smooth scroll with reduced-motion fallback
- [ ] `pc-scrollspy` CSS section, light + dark, indicator animation
- [ ] ExUnit + vitest suites green
- [ ] Playground page with dials (offset, nesting, indicator) and both scenarios
- [ ] PR with light + dark screenshots

mplatts:
Closes #613

## Summary

A new `<.scrollspy>` component plus a `PetalScrollspy` hook: the docs "on this page" rail that follows the reader down a long article.

Two APIs, both first-class in the moduledoc:

1. **The built-in renderer.** Pass `items` (label + target id, optionally one level of `children`), get a `<nav>` with an animated indicator bar. `offset`, `threshold`, `indicator` (`bar` / `none`), `heading`, `aria_label`, `class`, `rest`.
2. **The bare data-attribute contract.** `phx-hook="PetalScrollspy"` on any container whose links carry `data-scrollspy-target`, and the hook drives your own markup. It sets `aria-current="location"` and `pc-scrollspy-link--active` on the active link and nothing else. Zero `pc-scrollspy` classes needed.

The hook uses an `IntersectionObserver` over a reading band near the top of the viewport, but only as a *trigger* - every decision is recomputed from live rects. That keeps the rail honest when a section grows, the window resizes, or a LiveView patch swaps the content, and it makes the selection logic a pure function.

### Files

| File | What |
|---|---|
| `lib/petal_components/scrollspy.ex` | new component |
| `lib/petal_components/showcase/scrollspy.ex` | 3 registry examples (basic, nested, bare markup) |
| `assets/js/petal_components.js` | `PetalScrollspy` hook + exported `scrollspyActive/2` |
| `assets/default.css` | `pc-scrollspy` section, inside `@layer components` |
| `dev.exs` | `/c/scrollspy` playground: a real mini guide with offset / nesting / indicator dials |
| `test/petal/scrollspy_test.exs` | 19 component tests |
| `test/js/scrollspy.test.js` | 22 vitest specs |
| `lib/petal_components.ex`, `showcase/registry.ex`, `CHANGELOG.md` | registration |

- **Added an `aria_label` attr** (default `"On this page"`). The sketch hardcoded `aria-label` next to `{@rest}`; since `rest` is a global, a consumer passing their own `aria-label` would have produced a duplicate attribute on the element. An explicit attr is the only way to let them override it.
- **`"topmost wins"` is implemented as "the last section to have reached the activation line".** Read literally, "closest to the top of the viewport" is ambiguous when three short sections are all crowded above the line. The rule shipped is: among the sections that have reached the line, the one furthest down the page wins - i.e. the section that *owns* the reading position. It agrees with the issue's example cases and is unambiguous in the crowded one. Both are covered in the spec.
- **The nav carries a `pc-scrollspy--bar` / `pc-scrollspy--none` modifier.** The rail line is drawn on the list only in bar mode, so `indicator="none"` is genuinely a plain text list rather than a railed one with the bar removed.
- **Nested children render in a `pc-scrollspy__sublist` `<ul>`** inside the parent `<li>`. The sketch only named the link class; the sublist element is what makes the nesting real markup rather than indentation.
- **The selection logic is a named export, `scrollspyActive(sections, {line, atBottom})`**, alongside the hook object. That is what lets the vitest suite test the behaviour directly rather than through a fake observer.
- **Bottom snap only fires on a container that actually scrolls.** Without that guard a short page whose content fits entirely on screen would highlight its last entry forever, which is not a bottom any reader would recognise.
- **Default `rootMargin` is `-25% 0px -70% 0px`** (activation line at 25% of the viewport), overridable via `data-threshold` exactly as specified.

## Observer cleanup

`destroyed()` disconnects the observer, removes the `hashchange` and `resize` listeners from `window`, detaches the `scroll` listener from the scroll root, cancels any pending `requestAnimationFrame`, and restores every style the hook borrowed - the container's previous `scroll-behavior` and each target's previous inline `scroll-margin-top`. `scan()` disconnects the old observer before creating a new one, and the scroll listener *moves* with the scroll root rather than being left on the old one when a patch relocates the sections. There is a spec that destroys a hook and then fires a `hashchange` to prove the stale hook no longer drives the page.

The scroll listener exists because bottom snap is invisible to the observer: a closing section too short to reach the activation line never generates a callback. It is `passive` and coalesced to one pass per animation frame.

## Reduced motion and smooth scroll

Smooth scrolling is a property of the scroll container, not of the click, so the hook sets `scroll-behavior: smooth` on the scroll root at mount and puts the previous value back on destroy. Under `prefers-reduced-motion: reduce` it never sets it at all, so jumps are instant. The indicator bar is *positioned* by JS (transform + height) and *animated* by CSS, which means the reduced-motion opt-out for the transition is one media query in `default.css` - the bar still tracks the active link, it just arrives without the glide. Scrolling never moves focus in either case.

## Tests

| Suite | Before | After |
|---|---|---|
| `mix test` | 913 tests, 0 failures, 1 skipped | **932 tests, 0 failures, 1 skipped** |
| `npm test` | 157 passing | **179 passing** |

`mix format --check-formatted`, `mix compile --force --warnings-as-errors` and `mix credo` are all clean - credo reports the same 14 refactoring opportunities / 31 readability issues as `main`, so zero new entries.

## Screenshots

Captured and eyeballed locally at 1280x1800 on `/c/scrollspy`: light, dark, and a scrolled shot where a non-first entry is active (a scrollspy screenshot with the first item highlighted proves nothing). Verified in both schemes that the active link reads stronger than the rest, the indicator sits against it, the nested example indents correctly, and the bare-markup example highlights with no `pc-scrollspy` classes on it. Images are not committed - happy to attach them to this thread on request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,28 @@
 # Changelog
+### Unreleased
+
+#### Added
+
+- **New `scrollspy` component: the "on this page" rail that follows the
+  reader down a long article.** Pass the sections as `items` and the
+  `PetalScrollspy` hook takes it from there, watching them with an
+  IntersectionObserver and sliding an indicator bar to whichever one is
+  at reading position. The fiddly parts are the point: when several
+  sections are on screen the one that most recently reached the
+  activation line wins, a closing section too short to ever reach that
+  line still highlights once you hit the bottom of the scroll, and a
+  `#hash` in the URL activates immediately on load instead of waiting
+  for the first observer callback. `offset` sets `scroll-margin-top` on
+  the targets so a fixed site header stops covering the heading you just
+  jumped to, `threshold` moves the activation line, `indicator="none"`
+  drops the bar for a colour-only active state, and one level of
+  nesting handles h2/h3 structure. The hook is not tied to the renderer:
+  put `phx-hook="PetalScrollspy"` on any container whose links carry
+  `data-scrollspy-target` and it drives your own markup, setting
+  `aria-current="location"` and `pc-scrollspy-link--active` on the active
+  link. Smooth scrolling and the indicator transition both drop out under
+  `prefers-reduced-motion`, and scrolling never moves focus.
+
 ### 4.14.0 - 2026-08-11
 
 #### Added
--- a/assets/default.css
+++ b/assets/default.css
@@ -7586,3 +7586,67 @@
       transition-duration: 1ms;
     }
   }
+
+/* Scrollspy - the "on this page" rail.
+   Layered, unlike the tail above it: unlayered rules outrank every Tailwind
+   utility, which would stop a consumer overriding these with a class. */
+@layer components {
+  .pc-scrollspy {
+    @apply relative text-sm;
+  }
+
+  .pc-scrollspy__heading {
+    @apply block mb-3 text-xs font-semibold tracking-wide uppercase text-gray-900 dark:text-white;
+  }
+
+  .pc-scrollspy__list {
+    @apply flex flex-col;
+  }
+
+  /* The rail the bar slides along. Only the bar variant draws it - with
+     indicator="none" the active state is carried by the link alone. */
+  .pc-scrollspy--bar .pc-scrollspy__list {
+    @apply border-l border-gray-200 dark:border-gray-800;
+  }
+
+  .pc-scrollspy__sublist {
+    @apply flex flex-col;
+  }
+
+  .pc-scrollspy-link {
+    @apply block py-1.5 pl-4 pr-2 text-gray-500 transition-colors dark:text-gray-400;
+    @apply hover:text-gray-900 dark:hover:text-white;
+    @apply focus-visible:outline-hidden focus-visible:ring-2 focus-visible:ring-primary-500 focus-visible:rounded-sm;
+  }
+
+  .pc-scrollspy-link--nested {
+    @apply pl-8 text-[0.8125rem];
+  }
+
+  .pc-scrollspy-link--active {
+    @apply font-medium text-primary-600 dark:text-primary-300;
+  }
+
+  .pc-scrollspy-link--active:hover {
+    @apply text-primary-700 dark:text-primary-200;
+  }
+
+  /* Positioned by the hook (transform + height); the glide is CSS so the
+     hook never owns timing. Hidden until it has a link to sit against. */
+  .pc-scrollspy__indicator {
+    @apply absolute left-0 top-0 w-0.5 bg-primary-600 dark:bg-primary-300;
+    transition:
+      transform 200ms ease,
+      height 200ms ease;
+  }
+
+  @media (prefers-reduced-motion: reduce) {
+    .pc-scrollspy__indicator {
+      transition: none;
+    }
+
+    .pc-scrollspy-link {
+      transition: none;
+    }
+  }
+}
--- a/assets/js/petal_components.js
+++ b/assets/js/petal_components.js
@@ -5424,6 +5424,272 @@ export const PetalDataTable = {
   },
 };
 
+// Scrollspy: highlight the nav entry for the section being read.
+//
+// The observer is only a cheap trigger - every decision is made from live
+// rects, so a section that grew, a window resize, or a LiveView patch can
+// never leave the rail describing a page that no longer exists.
+//
+// Band, not line, for the observer: rootMargin crops the viewport down to a
+// reading strip near the top, so callbacks fire when a heading arrives there
+// rather than on every pixel of scroll.
+const SCROLLSPY_ROOT_MARGIN = "-25% 0px -70% 0px";
+
+// Which section owns the activation line.
+//
+// `sections` is [{ id, top }] in document order, `top` in viewport
+// coordinates; `line` is the activation line, also in viewport coordinates.
+// The last section that has reached the line wins, which is the one nearest
+// the top of the viewport that the reader has actually got to - a new heading
+// takes over as it arrives, not when the previous section finally leaves.
+//
+// Exported for the spec: this is the whole behaviour of the hook expressed as
+// a pure function, so it can be tested without an IntersectionObserver.
+export function scrollspyActive(sections, { line = 0, atBottom = false } = {}) {
+  if (!sections.length) return null;
+  // At the very bottom there is no scroll left to bring a short final section
+  // to the line, so nothing below would ever be reachable. Snap to the last.
+  if (atBottom) return sections[sections.length - 1].id;
+
+  let active = sections[0].id;
+  // 1px of slack: sub-pixel layout means a section resting exactly on the
+  // line can measure a hair below it and flicker back to its predecessor.
+  for (const section of sections) {
+    if (section.top - line <= 1) active = section.id;
+  }
+  return active;
+}
+
+// The scrollable box the sections live in. Sections inside an overflow pane
+// (an app shell with its own scroller) still measure against the viewport for
+// the observer, but "am I at the bottom" is a question about their scroller.
+function scrollspyScrollRoot(el) {
+  let node = el?.parentElement;
+  while (node && node !== document.body) {
+    const { overflowY } = getComputedStyle(node);
+    if (
+      (overflowY === "auto" || overflowY === "scroll") &&
+      node.scrollHeight > node.clientHeight + 1
+    ) {
+      return node;
+    }
+    node = node.parentElement;
+  }
+  return document.scrollingElement || document.documentElement;
+}
+
+// rootMargin's top component, resolved to viewport pixels. A negative top
+// margin crops the root downward, so the cropped edge - the activation line -
+// sits that far below the top of the viewport.
+function scrollspyLine(rootMargin, height) {
+  const top = String(rootMargin).trim().split(/\s+/)[0] || "0px";
+  const value = parseFloat(top);
+  if (Number.isNaN(value)) return 0;
+  return top.endsWith("%") ? (-value / 100) * height : -value;
+}
+
+export const PetalScrollspy = {
+  mounted() {
+    this.active = null;
+    this.frame = null;
+    this.targetStyles = new Map();
+
+    this.onHashChange = () => this.applyHash();
+    // Coalesce to one pass per frame: scroll outruns paint, and each pass
+    // reads layout.
+    this.onScroll = () => {
+      if (this.frame) return;
+      this.frame = requestAnimationFrame(() => {
+        this.frame = null;
+        this.sync();
+      });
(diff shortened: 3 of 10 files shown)


mplatts:
## Screenshots

**Tracking a non-first section** — scrolled so "Wiring it up" owns the reading position. The indicator bar sits beside it, and the nested example below tracks independently.

Scrollspy active on item 2

**Light**

Scrollspy, light mode

**Dark**

Scrollspy, dark mode

### Verified independently

| Check | Result |
|---|---|
| `mix test` | 932 tests, 0 failures, 1 skipped (+19) |
| `npm test` | 179 passing (+22) |
| new dependencies | none |
| slugs : `render_page` clauses | 61 : 61 |

The `dark:text-primary-400` → `primary-300` fix is a good catch and the kind of thing that only shows up if you actually look: with the playground's default *neutral* primary, 400 lands at the same lightness as `gray-400`, so the active state was invisible in dark mode. 300 is also the more common convention in `default.css` (10 uses vs 7).

Two decisions worth a maintainer eye, both flagged in the PR body: "topmost wins" was ambiguous in the issue and shipped as "the last section to reach the activation line"; and a passive scroll listener rides alongside the observer because bottom-snap is invisible to `IntersectionObserver` by construction.

Gaps the author was straight about: keyboard-only, `prefers-reduced-motion` and hash-on-load are covered by unit specs but were not clicked through in a real browser.

PR #632, opened 2026-08-12
feat(qr_code): server-rendered QR codes as inline SVG

### LINKED ISSUE #606: Component: QR Code
## Summary

A server-rendered `<.qr_code>` function component that encodes a string into a QR code and emits it as inline SVG. Zero JavaScript, crisp at any size, themeable via `currentColor`, and it prints. Reui does not ship one, but the wider shadcn ecosystem does (OriginUI and others), and the utility is high in exactly the apps Phoenix people build: TOTP enrolment screens (`otpauth://` URIs), payment/checkout links, "open on your phone" and share-this-page moments, WiFi credentials on kiosk screens.

## Critical design decision: the encoder (flag in the PR)

QR encoding (byte-mode segmentation, Reed-Solomon error correction, mask evaluation) is real algorithmic work and it has to live somewhere. Two options; **(a) is the recommendation**, but this needs explicit maintainer sign-off because it would be the library's first optional hex dependency:

**(a) Recommended: optional dependency on `eqrcode`** (MIT, pure Elixir, zero transitive deps). Declare it `optional: true` in `mix.exs`. If the user renders `<.qr_code>` without `eqrcode` installed, fail loudly and helpfully: raise at render time (guarded by `Code.ensure_loaded?(EQRCode)`) with a message that says exactly what to add to `mix.exs` (`{:eqrcode, "~> 0.2"}`) and why. No silent blank output. Every other component keeps working with no dep installed. Note: use `eqrcode` for the encoding matrix only — rendering stays ours (see below), so we control the SVG, theming, and rounding.

**(b) Fallback: vendor a minimal byte-mode encoder** (~300 lines: byte-mode data encoding, Reed-Solomon over GF(256), the 8 mask patterns + penalty scoring, format/version info) under `lib/petal_components/qr_code/`. Zero deps, but it is crypto-adjacent algorithmic code we then maintain forever, and versions/EC levels multiply the test surface.

Present the choice in the PR description with whichever was built. If (a) is vetoed on review, (b) is the pivot.

## References

- OriginUI QR code (shadcn ecosystem) — scope and visual reference: rounded-module look, centre logo
- `eqrcode` on Hex (https://hex.pm/packages/eqrcode) — candidate encoder, MIT, pure Elixir
- Study in-repo before writing any code:
  - `lib/petal_components/sparkline.ex` — the house pattern for a pure server-rendered SVG component: `currentColor` inheritance, class-driven sizing, `viewBox` + `aria-hidden` handling, private geometry helpers, `fmt/1` for stable float output
  - `test/petal/sparkline_test.exs` — the test style for SVG components (LazyHTML/string assertions on rendered markup)
  - `lib/petal_components.ex` — confirm `qr_code` does not collide with anything already exported by `use PetalComponents`

## API sketch

`lib/petal_components/qr_code.ex`, `use Phoenix.Component`. Starting proposal; refine and flag deviations in the PR.

```elixir
attr :value, :string, required: true, doc: "the string to encode (URL, otpauth:// URI, WiFi string, arbitrary text)"

attr :size, :any, default: nil,
  doc: "optional pixel size (integer, sets width/height attrs); omit and drive size with classes instead, e.g. class=\"size-48\""

attr :color, :string, default: "currentColor",
  doc: "module (dark) colour; currentColor by default so it themes with text-* classes and prints"

attr :background, :string, default: "transparent",
  doc: "background fill behind the modules; transparent by default. Set explicitly (e.g. \"white\") when placing on dark or busy surfaces"

attr :error_correction, :atom, default: :m, values: [:l, :m, :q, :h],
  doc: "QR error correction level; automatically bumped to :h when the logo slot is used"

attr :rounded, :any, default: 0,
  doc: "0..1 module corner rounding for the modern look; 0 = square modules, 1 = fully round dots"

attr :class, :any, default: nil, doc: "size + colour, e.g. \"size-48 text-gray-900 dark:text-white\""
attr :rest, :global

slot :logo,
  doc: "optional centre logo. Content is placed in a knocked-out quiet zone in the middle of the code; using this slot forces error_correction to :h so the code stays scannable"
```

Rendering rules:

- Inline `<svg>` with a `viewBox` matching the module grid plus the mandatory 4-module quiet zone. Crisp at any raster size.
- All dark modules emitted as a **single `<path>`** (one `d` string of module rects/rounded rects), not one element per module. A version-5 code is ~800 modules; one path keeps the payload and the DOM sane.
- `rounded` maps to per-module corner radius within the path data (arc segments), clamped to half the module size at `rounded={1}`.
- `shape-rendering="crispEdges"` when `rounded == 0` so square modules do not anti-alias into grey seams.
- Root gets `class={["pc-qr-code", @class]}`, `role="img"`, and a sensible label: accept the standard `aria-label` via `@rest`, defaulting to `"QR code"` when none is supplied. Never dump `@value` into the label (it can be a secret, e.g. a TOTP URI).
- Logo slot: render the slot content inside a centred `<foreignObject>` (or nested `<svg>`/`<image>` for image content) over a knockout rectangle filled with `@background`; knockout capped at ~25% of code width, which EC level :h tolerates.

`assets/default.css` gets a small `pc-qr-code` section (display, `max-width: 100%`, sensible default block behaviour). Colour theming rides `currentColor`, so keep CSS minimal.

## Variants & states to build

- [ ] Default render: square modules, `currentColor` on transparent, class-driven size
- [ ] Explicit `size` (px) render path
- [ ] `error_correction` all four levels `:l` `:m` `:q` `:h`
- [ ] `rounded` at 0, mid (0.5), and 1
- [ ] Explicit `color` / `background` overrides
- [ ] Logo slot: knockout + automatic EC bump to `:h`
- [ ] Long-value handling: pick the smallest QR version that fits, error clearly (raise with a helpful message) if the value exceeds the max capacity for the chosen EC level
- [ ] Missing-encoder state (option (a) only): clear compile/render-time error with install instructions

## Playground page

Nav entry + hero + dials, per the house playground pattern (`dev.exs`, segmented controls wired via `handle_event`):

- **Content preset dial**: URL (`https://petal.build`), TOTP (`otpauth://totp/Petal:user@example.com?secret=JBSWY3DPEHPK3PXP&issuer=Petal`), WiFi (`WIFI:T:WPA;S:MyNetwork;P:secret;;`) — plus a free-text input
- **Size dial**: S / M / L / XL
- **Error correction dial**: L / M / Q / H
- **Rounding dial**: square / soft / dots (0 / 0.5 / 1)
- **Logo toggle**: on/off, demonstrating the automatic EC bump

Example scenarios (real product moments, not lorem ipsum):

1. **TOTP enrolment card** — "Scan with your authenticator app" panel with the QR, the manual-entry secret beneath, and a confirm-code input. This is the number-one real use.
2. **Share this page** — small QR in a popover/card footer next to a copy-link button.
3. **Dark mode surface** — QR on a `bg-gray-900` panel demonstrating the correct inversion (see Accessibility).

The playground page must include a **scannability check in both modes**: actually scan the light and dark renders with a phone before opening the PR.

## Dark mode and the inversion rule (document this in the moduledoc)

QR scanners expect dark modules on a light background. Rendering light modules on a dark surface (inverted) is only reliably scannable when: (1) the quiet zone is also the light... no — the correct rule is: on a dark surface, either (a) keep the code dark-on-light by giving it an explicit light `background` (a white rounded card behind the code), or (b) invert to light modules on the dark surface only when the quiet zone is rendered in the dark surface colour and contrast between modules and surface is high (most modern scanners handle inversion, some older ones do not). The moduledoc and playground must document and demonstrate rule (a) as the safe default, with (b) shown as the "looks nicer, test your audience's scanners" option. Enforce contrast in the examples; never ship a grey-on-grey demo.

## Accessibility

- `role="img"` on the `<svg>` with a default `aria-label="QR code"`, overridable via the label attr and never derived from `@value`
- Decorative internals (`<path>`, knockout rect) need no individual labelling
- The component is non-interactive; no keyboard map
- Pair every playground QR with the same information in text form (the URL, the manual TOTP secret) — a QR code must never be the only path to the content

## Tests

`test/petal/qr_code_test.exs` using `ComponentCase` + LazyHTML:

- [ ] **Known-vector encoding assertion**: encode `"HELLO"` at EC `:m`, extract the module matrix (or the emitted path `d` string), and assert a stable hash of it. This pins the encoder output; any encoder change that alters the matrix must be deliberate.
- [ ] Renders `<svg>` with `pc-qr-code`, `role="img"`, default `aria-label="QR code"`; custom label overrides the default; `@value` never appears in the label
- [ ] Modules render as a single `<path>` (assert exactly one module path element)
- [ ] `color` / `background` attrs appear in the output; defaults are `currentColor` / `transparent`
- [ ] `size` sets width/height attrs; omitted `size` leaves sizing to classes
- [ ] `rounded={0}` emits no arc commands; `rounded={1}` does; `shape-rendering="crispEdges"` present only at 0
- [ ] Each `error_correction` level produces a distinct matrix for the same value
- [ ] Logo slot renders slot content, emits the knockout, and forces EC `:h` (assert the matrix matches an explicit `:h` render of the same value)
- [ ] Value exceeding capacity raises with a helpful message
- [ ] Option (a) only: rendering without `eqrcode` available raises the install-instructions error (test via the guard function if the dep is present in the test env)
- [ ] `class` and `rest` passthrough

## Non-goals

- No animated or scannable-video QR, no gradient fills, no per-module colour art styles
- No client-side generation, no JS hook, no canvas
- No barcode formats other than QR (no Code128, no Data Matrix)
- No download-as-PNG button in the component itself (playground may demo right-click-save of the SVG, nothing more)
- No automatic contrast checking at runtime; the inversion rule is documentation + playground demonstration

## Process

1. Build on a feature branch: `lib/petal_components/qr_code.ex` + the `default.css` section + `test/petal/qr_code_test.exs` + playground page (nav entry, dials, examples). If option (a): add `eqrcode` as `optional: true` in `mix.exs` and call the decision out in the PR description for sign-off.
2. Verify in the playground (`mix run dev.exs`, http://localhost:4000): every dial, light + dark, and physically scan both light and dark renders with a phone.
3. Run the suites: `mix test` (no JS hook, so no `npm test` additions). All green before PR.
4. Open a PR against `main` with screenshots (light + dark) of the playground page, plus a phone-camera confirmation that the rendered codes scan. The automated review must pass.
6. Release (Hex publish, docs, changelog, MCP schema sync) is handled by maintainers after merge; not part of the contribution.

## Definition of done

- `<.qr_code>` component with the API above; encoder decision (a)/(b) made, implemented, and flagged in the PR
- Single-path SVG output, `currentColor` theming, quiet zone, `rounded` support, logo slot with automatic EC `:h` bump
- `pc-qr-code` section in `assets/default.css`
- Tests green, including the known-vector matrix hash assertion
- Playground page with content presets, size/EC/rounding dials, logo toggle, TOTP enrolment + share + dark-surface scenarios
- Moduledoc documents the dark-surface inversion rule with the safe default; every attr has a `doc:` string and the moduledoc includes usage examples
- PR screenshots (light + dark) and confirmed real-world scans

mplatts:
Closes #606

`<.qr_code>` encodes a string and emits it as inline SVG. No JavaScript, no canvas, crisp at any size, and it prints.

- Every dark module lands in a **single `<path>`**. Square modules merge into horizontal runs, which halves the path data and kills the hairline seams you get from one rect per module; rounded modules stay individual, since that is the whole point of the look.
- The spec's **4-module quiet zone**, laid down by us after stripping the encoder's own 2-module padding, so the `viewBox` is exactly `modules + 8`.
- `currentColor` by default, so `class="size-40 text-gray-900 dark:text-white"` themes it. Explicit `color` / `background` when you need them.
- `size` for pixel `width`/`height`; omit it and classes drive the size.
- `rounded` 0..1: square -> soft -> full dots. `shape-rendering="crispEdges"` only at 0.
- `:logo` slot knocks a **real hole** in the matrix (modules omitted, not covered over) and forces error correction to `:h`. Slot content is laid out in a 100x100 box scaled into the hole.
- `role="img"` with a `label` that is **never** derived from `@value` - a TOTP enrolment URI contains the shared secret. `aria-hidden="true"` drops `role` and the label for the case where the same data sits beside it as text.
- A value past the level's byte ceiling raises with the real numbers and the way out, rather than failing deep inside the encoder.

## ⚠️ The encoding decision - needs your sign-off

**This adds `eqrcode` to `mix.exs` as `optional: true`.** That is the issue's recommended option (a), and the issue says it needs explicit maintainer sign-off, so flagging it loudly here rather than burying it.

What `optional: true` actually means for consumers:

- It does **not** enter anyone's dependency tree. Apps that never render a QR code carry nothing new. `mix.lock` changes here, consumers' do not.
- Apps that do want `<.qr_code>` add `{:eqrcode, "~> 0.2"}` themselves. It is MIT, pure Elixir, and has zero transitive deps of its own.
- Render `<.qr_code>` without it and you get a `raise` naming the exact mix.exs line, guarded by `Code.ensure_loaded?(EQRCode)`. Never a silent blank box.

Only the **matrix** comes from `eqrcode`. The SVG, quiet zone, run merging, rounding, logo knockout, colour and a11y are all ours, so we keep control of the output.

**The honest cost:** every other component in this library is drop-in with zero setup, and this one is not. `mdex` and `lumis` are optional but degrade gracefully; a QR code cannot. Worth weighing against option (b), a ~400-line vendored Reed-Solomon encoder we would then own forever. I went with (a) because the failure mode of a hand-rolled encoder is the bad kind - codes that scan on the phone you tested with and fail on a cheap scanner - and because a draft PR is the right place to surface the decision rather than pre-empt it. **If you veto it, (b) is the pivot and nothing else in this PR changes.**

Worth adding the dep note to the MCP schema and rules.md if this lands, so agents emitting `<.qr_code>` know to add it.

## Scannability

I cannot hold a phone up to CI, so I did the machine equivalent - rasterised the real output and decoded it with OpenCV's QR detector:

**From the emitted path data** (9/9 exact round-trips): `HELLO` at `:m`; `https://petal.build` at all four EC levels; an `otpauth://totp/...` URI; a `WIFI:T:WPA;...` string; a 240-byte payload at `:q`; and the logo-knockout render.

**From actual browser-rendered pixels** on the playground (5/5 exact): `square`, `soft`, `dots`, `dots + logo`, and `dots + logo inverted on a dark surface`.

The known-vector test pins that `HELLO` matrix by SHA-256 of the path, so any encoder change that moves a module fails the suite loudly.

## Dark mode and the inversion rule

Documented in the moduledoc and demonstrated side by side on the playground:

- **(a) safe** - keep the code dark-on-light with an explicit light `background` on a light card. Scans on anything.
- **(b) inverted** - light modules on the dark surface, quiet zone in the surface colour. Looks better, most modern phone cameras cope, some older and embedded scanners do not.

Because `color` defaults to `currentColor`, `text-gray-900 dark:text-white` gives you (b) automatically in dark mode. That is deliberate and called out as such.

- **Added a `label` attr** (default `"QR code"`). The sketch only described `aria-label` via `@rest`. Both work - `aria-label` in `@rest` wins over `label`, and neither is emitted when `aria-hidden` is present. A first-class attr is more discoverable and shows up in the props table.
- **The logo knockout omits modules rather than covering them with a `@background`-filled rect.** The sketch's approach breaks at the default `background="transparent"`, where a transparent rect knocks nothing out. Omitting the modules is a true knockout at any background.
- **Slot content goes in a nested `<svg viewBox="0 0 100 100">` wrapping the `foreignObject`**, not a bare `foreignObject`. A bare one would size the slot's HTML in module units (~8px), so any sane font size would overflow. The 100x100 box is a usable authoring scale, documented on the slot.
- **Square modules merge into horizontal runs.** The sketch says one path, which this is; run merging is an extra that roughly halves the `d` string and removes anti-aliasing seams.
- **Capacity is checked up front** against the version-40 byte ceiling per level, so the error is ours and useful. `eqrcode` otherwise fails with a `MatchError` on `{:error, :no_version_found}`.
- **Playground has no free-text "custom" segment in the content dial** - the value input below the hero is always live and editable, and typing in it flips the dial to custom. One text field instead of two.

## Test counts

| | before | after |
|---|---|---|
| Elixir | 913 tests, 0 failures, 1 skipped | **938 tests, 0 failures, 1 skipped** |
| JS | 157 passing | **157 passing** (no hook, so no JS added) |
| credo | 14 refactoring, 31 readability | **14 refactoring, 31 readability** (zero new) |

## Screenshots

Playground at `/c/qr-code`, light and dark, full page - hero with all six dials, the TOTP enrolment card, the share card, the inversion-rule panel and the props table. Attaching separately.

## Not done

- No physical phone scan. The OpenCV round-trips above are the substitute; if you want a real phone pass before merge, that is your call at review.


mplatts:
## Screenshots

Playground page at `/c/qr-code`, 1280px wide, full page — hero, dials, TOTP card, share card, inversion panel, props table.

**Light**

QR code, light mode

**Dark**

QR code, dark mode

### Verified independently

| Check | Result |
|---|---|
| `mix test` | 938 tests, 0 failures, 1 skipped (+25) |
| `npm test` | 157 passing, unchanged |
| `compile-without-optional-deps` (CI) | **pass** — the library still builds with `eqrcode` absent |

## One correction to the issue's framing, relevant to the sign-off

Issue #606 says option (a) "would be the library's first optional hex dependency". That isn't the case — `mix.exs` on `main` already carries two runtime optional deps:

```elixir
{:mdex, "~> 0.12", optional: true},
{:lumis, "~> 0.6", optional: true},
```

(plus `a11y_audit` and `wallaby`, both `only: :test`). So `eqrcode` follows an established pattern rather than setting a new precedent, and `chat` already does the same "degrade when the optional dep is absent" dance that `qr_code` does here. That doesn't decide the question — it's still @nhobes's call, and the "every other component is drop-in, this one isn't" cost stands — but the decision is less novel than the brief suggests.

The guard is implemented as the issue specified: `Code.ensure_loaded?(EQRCode)` at render time, raising with the exact `mix.exs` line rather than emitting silent blank output.

## On scannability

This is the part worth trusting, and it's better evidence than a screenshot. Rather than eyeballing it, the encoder output was round-tripped through OpenCV's QR detector: 9/9 exact decodes from the emitted path data (including `otpauth://`, a `WIFI:` string, a 240-byte payload and all four EC levels), and 5/5 from real browser-rendered pixels across the square, soft, dots, dots+logo, and inverted-on-dark renders. A known-vector test also pins the `HELLO` matrix by SHA-256, so any encoder change that moves a module fails loudly.

Still outstanding: no physical phone scan (issue step 2 asks for one). The OpenCV round-trips are a substitute, not a replacement.

If (a) is vetoed, (b) — vendoring a ~300-line byte-mode encoder — is a clean pivot, and only the matrix source changes since the SVG, theming, rounding and logo knockout are all native already.

**If this lands**, the optional-dep note needs to reach `rules.md` and the MCP schema too, since `<.qr_code>` won't work drop-in the way every other component does.

nhobes:
Polish pass done. The one decision gate the issue reserved for you: **eqrcode as the library's first component-critical optional hex dep** (mix.exs, MIT, pure Elixir, zero transitive deps, optional: true so non-QR apps never carry it). The issue named this as needing explicit maintainer sign-off before merge; the stated pivot if vetoed is vendoring a minimal encoder. If you bless it, remember the post-merge cascade the PR notes: rules.md + MCP schema note about the optional dep. (Also: I left the CSS section as its own @layer block rather than folding it into the previous one - the #629 audit explicitly blessed own-block as matching the neighbours, so the fold suggestion here would have contradicted that.)

PR #633, opened 2026-08-12
feat(context_menu): right-click menu anchored at the pointer

### LINKED ISSUE #601: Component: Context Menu
## Summary

A net-new `<.context_menu>` component: a right-click (and long-press on touch) menu attached to an arbitrary region of the page. Right-clicking anywhere inside the trigger region opens a floating menu panel at the cursor; keyboard users open it with Shift+F10 or the Menu key, positioned at the trigger. It completes the menu family alongside `<.dropdown>` — same panel surface, same item anatomy, different invocation model. This is fundamentally an interaction component: the panel and items already have a visual language in the library, and the work here is the open/position/dismiss behaviour done properly.

Target milestone: **4.15.0**.

## References

- reui.io "Context Menu" (10 examples) — study the range of item treatments: icons, keyboard shortcut hints, destructive items, separators, disabled items, labels.
- Radix UI ContextMenu — study open/close behaviour, collision handling, long-press on touch, and the keyboard model. Reference behaviour and variants only; we build a native HEEx/LiveView equivalent from scratch.
- shadcn/ui Context Menu — the visual bar for the panel and items.
- WAI-ARIA APG "Menu and Menubar" pattern — the authoritative spec for roles and the keyboard map.

Existing in-repo prior art to study before writing anything:

- `lib/petal_components/dropdown.ex` — the menu item anatomy this component reuses: `dropdown_menu_item/1`, `dropdown_menu_label/1`, `dropdown_menu_separator/1`, and the `pc-dropdown__menu-items-wrapper` floating-panel surface in `assets/default.css` (around line 1962). Decide whether the context menu items share these CSS classes directly or get `pc-context-menu__*` classes that `@apply`/compose the same styles — either is acceptable, but do not fork the visual treatment.
- `lib/petal_components/popover.ex` and the `PetalPopover` hook in `assets/js/petal_components.js` (around line 1446) — the precedent for a top-layer panel positioned by a hook, including the native `popover` attribute for browser-managed light dismiss.
- `assets/js/petal_components.js` — all hooks live in this single file and are exported from the bottom index. `PetalContextMenu` goes here too.

## Why a hook is needed

CSS cannot read cursor coordinates. Opening a menu at the pointer position, clamping it inside the viewport (flip/shift when the click is near the right or bottom edge), and repositioning on subsequent right-clicks while already open all require JavaScript. This clears the house bar for a hook (precedents: `PetalInputOTP`, `PetalPopover`). The hook is `PetalContextMenu`, added to `assets/js/petal_components.js` and exported from the hooks index. No external packages; no npm dependencies.

Recommended mechanics (builder may refine): render the panel as a native `popover="manual"` element in the top layer, as `PetalPopover` does for its `top_layer` mode. The hook listens for `contextmenu` on the trigger region (calling `preventDefault`), long-press (pointer down held ~500ms without movement) on touch, and Shift+F10 / the `ContextMenu` key when focus is inside the region. On open it shows the popover, positions it at the cursor (or at the focused element for the keyboard path), clamps to the viewport, and moves focus into the menu. Dismiss on Escape, click outside, and item activation; focus returns to the trigger region on close.

## API sketch

Starting proposal — refine as needed and flag deviations in the PR.

```elixir
attr :id, :string, doc: "the menu panel id; autogenerated if not set"
attr :disabled, :boolean, default: false,
  doc: "disables the context menu entirely; right-click falls through to the browser default"
attr :class, :any, default: nil, doc: "extra classes for the trigger region wrapper"
attr :menu_class, :any, default: nil, doc: "extra classes for the menu panel"
attr :rest, :global

slot :trigger, required: true,
  doc: "the right-clickable region. Rendered as a block wrapper; put anything inside - a card, a table row, a text area"

slot :inner_block, required: true, doc: "the menu items"

def context_menu(assigns)
```

Menu items: reuse the dropdown item components (`<.dropdown_menu_item>`, `<.dropdown_menu_label>`, `<.dropdown_menu_separator>`) inside the panel, or mirror them as `context_menu_item` / `context_menu_label` / `context_menu_separator` if the role wiring (see Accessibility) makes reuse awkward. Whichever way it lands, items must support:

```elixir
attr :to, :string, default: nil
attr :label, :string
attr :link_type, :string, default: "button", values: ["a", "live_patch", "live_redirect", "button"]
attr :variant, :string, default: "default", values: ["default", "danger"],
  doc: "danger renders the destructive treatment (danger ramp text, danger hover surface)"
attr :kbd, :string, default: nil, doc: "right-aligned keyboard shortcut hint, e.g. \"⌘D\""
attr :rest, :global
slot :inner_block
```

Icons work the way they already do in dropdown items: an `<.icon>` placed in the inner block. Follow the attr naming, `doc:` string style, and `Link.a` delegation already used in `lib/petal_components/dropdown.ex`.

Check `lib/petal_components.ex` before naming anything: no collisions with existing exports or with `Phoenix.Component`/core_components names.

## Variants & states to build

- [ ] Open at cursor on right-click, with viewport clamping (flip up when near the bottom edge, shift left near the right edge)
- [ ] Reposition when right-clicking elsewhere in the region while already open
- [ ] Long-press open on touch devices
- [ ] Keyboard open (Shift+F10 and the Menu key) positioned at the trigger
- [ ] Close on Escape, outside click, and item activation; focus restored on close
- [ ] `disabled` on the component: native browser context menu is not suppressed
- [ ] Items: plain, with icon, with `kbd` hint, `variant="danger"`, `disabled`, all four `link_type` values
- [ ] Labels and separators between item groups
- [ ] Open/close transition consistent with dropdown/popover, complete rest state under `prefers-reduced-motion`
- [ ] Light and dark mode (floating-panel surface: `bg-white dark:bg-gray-900` with real borders in both modes)
- [ ] Panel radius consumes the `--pc-radius` token per the small-panel rule; item radius uses the nested-child formula

## Playground page

Nav entry plus:

- **Hero / example 1 — file-card grid**: a grid of file cards (name, icon, modified date). Each card is a `<.context_menu>` trigger region; right-click gives Open, Rename, Duplicate, a separator, then Delete as `variant="danger"` with a `kbd` hint (e.g. ⌘⌫). One item disabled to show the state.
- **Example 2 — text-selection area**: a paragraph block whose context menu offers Copy, Highlight, Share, Search the web — the "right-click on content" moment rather than "right-click on an object".
- **Dials**: just a `disabled` toggle wired via `handle_event` — this component is about the interaction, not visual variants. A short hint line on the page telling keyboard users to focus the region and press Shift+F10.

Demos to the usual standard: real product moments, no lorem ipsum. Verify in the playground with `mix run dev.exs` at http://localhost:4000.

## Accessibility

Follow the WAI-ARIA APG Menu pattern:

- Panel: `role="menu"`, `aria-orientation="vertical"`. Items: `role="menuitem"`; separators `role="separator"`; labels are non-interactive (`role="none"`/presentation as appropriate).
- Focus management inside the open menu via **roving tabindex or `aria-activedescendant`** — builder's choice, but one of the two, implemented in the `PetalContextMenu` hook. First item receives focus on open (per the pattern, keyboard-invoked opens focus the first item; pointer opens may focus the panel and move to items on first arrow press).
- Keyboard map:
  - Shift+F10 / Menu key (focus inside trigger region): open at trigger
  - Down/Up arrows: move through items, skipping disabled ones, wrapping at the ends
  - Home/End: first/last item
  - Enter/Space: activate the focused item, then close
  - Escape: close and return focus to the trigger region
  - Tab: closes the menu (focus leaves the widget)
- Focus returns to where it was in the trigger region on any close path.
- `focus-visible` ring in primary only; no persistent `:focus` fills.
- Decorative layers (if any) get `aria-hidden`.

The playground page should be fully operable keyboard-only, and the demo should say so.

## Tests

`test/petal/context_menu_test.exs` using `ComponentCase` with LazyHTML assertions:

- [ ] Renders the trigger region wrapper with the hook attached (`phx-hook="PetalContextMenu"`) and an autogenerated id when none is given; explicit `id` is honoured
- [ ] Panel renders with `role="menu"`, `aria-orientation="vertical"`, and the `popover` attribute (or hidden state) so it is not visible at rest
- [ ] `disabled` on the component omits the hook wiring (or sets the data attribute the hook checks)
- [ ] Items render `role="menuitem"`, each `link_type` produces the right element, `disabled` items get the disabled class and attribute
- [ ] `variant="danger"` emits the danger classes; `kbd` renders the right-aligned hint
- [ ] Separator renders `role="separator"`; label renders non-interactive
- [ ] `class`, `menu_class`, and `rest` pass through
- [ ] Every attr, variant, and slot has at least one rendering assertion; a11y attributes asserted explicitly

`npm test` for the hook: open on `contextmenu` with `preventDefault`, viewport clamping maths, keyboard open, roving focus/activedescendant movement (arrows, Home/End, wrap, disabled skip), Escape close with focus restore, close on outside click and item activation, long-press open, `disabled` no-op.

## Non-goals

- **Submenus (nested menus)** — explicitly out of scope for v1. Design the item markup so a submenu item can be added later without breaking the API, but build none of it.
- Checkbox/radio menu items (`menuitemcheckbox`/`menuitemradio`) — later enhancement if demand shows up.
- Colour/variant dials on the panel itself — the panel is the standard floating surface, full stop.
- Alpine.js support — LiveView.JS + CSS + the hook only.
- Replacing or restyling the browser's native context menu outside the trigger region.

## Process

1. Build on a feature branch: `lib/petal_components/context_menu.ex` + a `pc-context-menu` section in `assets/default.css` + the `PetalContextMenu` hook in `assets/js/petal_components.js` (exported from the hooks index) + tests + playground page (nav entry, disabled dial, both examples). Everything lives in this repo — no external packages, no npm deps.
2. Verify in the playground (`mix run dev.exs`, http://localhost:4000): right-click near all four viewport edges, long-press on a touch device or emulation, keyboard-only (Shift+F10, arrows, Home/End, Escape), light + dark, `prefers-reduced-motion`.

## Definition of done

- `<.context_menu>` (plus item components, whether reused from dropdown or mirrored) merged with the full checklist above
- `PetalContextMenu` hook with passing JS tests
- CSS section in `assets/default.css` sharing the dropdown panel/item visual language
- `mix test` and `npm test` green
- Playground page with nav entry, disabled dial, file-card grid demo, and text-selection demo, keyboard-operable
- PR screenshots in light and dark
- Docs-ready attr docs: every attr has a `doc:` string; the moduledoc includes usage examples covering the pointer, touch, and keyboard paths

mplatts:
Closes #601

## Summary

A net-new `<.context_menu>`: right-click a region of the page and get a menu at the cursor. Touch users get it from a half-second long press, keyboard users from `Shift+F10` or the Menu key. It completes the menu family alongside `<.dropdown>` - same floating surface, same item anatomy, different invocation model.

Five parts, per CONTRIBUTING:

1. **Module** `lib/petal_components/context_menu.ex` - `context_menu/1` plus `context_menu_item/1`, `context_menu_label/1`, `context_menu_separator/1`. Every attr has a `doc:`; the moduledoc carries the pointer, touch and keyboard examples.
2. **Styles** - a `pc-context-menu` section appended to `assets/default.css` in its own `@layer components` block. Composes the dropdown panel and item treatment rather than forking it: `--pc-radius` with the nested-child formula on items, gray ramp neutrals, `dark:` pairs, `focus-visible` only, reduced-motion rest state.
3. **Tests** - `test/petal/context_menu_test.exs` (17 specs) and `test/js/context_menu.test.js` (32 specs, vitest + jsdom, modelled on `popover.test.js`).
4. **Showcase** - `lib/petal_components/showcase/context_menu.ex` plus its registry entry.
5. **Playground** - nav entry under Overlay, `disabled` dial, a file-card grid and a right-click-content paragraph. Slug `context-menu`, verified rendering.

Plus an `### Unreleased` / `#### Added` changelog entry.

- **Items are mirrored, not reused.** The issue left this open. I went with `context_menu_item` / `context_menu_label` / `context_menu_separator` rather than reusing the dropdown ones, because the items need `tabindex="-1"` (roving tabindex), a `data-pc-context-menu-item` marker the hook queries, `aria-disabled`, the `kbd` slot and the `danger` variant - none of which belong on `dropdown_menu_item`. The **visual treatment is not forked**: `pc-context-menu__item` mirrors the `pc-dropdown__menu-item` declarations, and the panel mirrors `pc-dropdown__menu-items-wrapper`.
- **`class` lands on the trigger region, not the outer wrapper.** The outer wrapper is `display: contents` plumbing (it exists to host the hook and pair trigger with panel), so a class on it would do nothing. The attr doc already said "trigger region wrapper", and this makes it true.
- **Added `context_menu_label` gets `role="none"`** and items get `aria-disabled` - both implied by the a11y section, neither in the sketch.
- **Keyboard opens position at the trigger's top-left + 8px**, not centred on it. Matches how a keyboard-invoked OS menu behaves.

## Anchoring: reused, or new?

**Reused, with one deliberate departure.** The dismiss machinery is a direct port of the shape `PetalDataTable` converged on: `onPressStart`/`onPressEnd`/`onPressCancel` tracking presses per `pointerId`, a 10px tap slop, and multi-touch suppression - so a drag that starts on the page is a scroll, not a dismissal, and a pinch never counts. The viewport helper (`visualViewport` when present, window otherwise) and the whole-pixel write discipline come from `PetalPopover`.

The departure is **where the panel lives**. `PetalDataTable`'s menus are in-page and absolutely positioned so the page carries them during scroll. That is right for a panel anchored to a button, but wrong here: a context menu's anchor is a cursor coordinate, which is a *client* coordinate, and the trigger region is arbitrary content that frequently sits inside an `overflow: hidden` card grid or table cell. So the panel is a native `popover="manual"` element positioned `fixed`, which is the precedent `<.popover top_layer>` already sets. Consequences handled:

- `manual`, not `auto`: the UA's light dismiss fires on **pointerdown**, which is exactly the behaviour the data-table fix existed to remove. `auto` would also close this menu whenever any other popover on the page opened.
- Fixed coordinates go stale when the page moves under them, so the hook **closes on scroll** (ignoring scroll inside its own panel) and repositions on resize. Same call Radix makes.
- `showPopover()` and the coordinate writes happen in one task, so there is no 0,0 flash to work around - no `beforetoggle` dance needed.

## Viewport collision

Default is down-and-right of the cursor, the direction every desktop OS opens one. Near an edge it **flips to the other side of the cursor** rather than sliding, so the pointer never lands on top of the first item (which would arm an accidental activation). After flipping, both axes clamp into the padded viewport, and the panel gets a `max-height` + `overflow-y: auto` so a menu taller than the screen scrolls instead of running off it. All four cases have specs.

## Touch long-press

`pointerdown` with `pointerType === "touch"` inside the trigger arms a 500ms timer; any movement past 10px, an early lift, or a `pointercancel` disarms it - so a scroll flick that starts on a card stays a scroll. On fire, the menu opens at the finger. CSS sets `-webkit-touch-callout: none` on the trigger so iOS shows the menu instead of the selection loupe, and `touch-action` is deliberately left alone so the region still scrolls. Android's native long-press `contextmenu` is handled by the same `contextmenu` path as a right-click.

## Accessibility

WAI-ARIA menu pattern with **roving tabindex** (real focus, not `aria-activedescendant`). Panel `role="menu"` + `aria-orientation="vertical"`; items `role="menuitem"` `tabindex="-1"`; separators `role="separator"`; labels `role="none"`. Keyboard-invoked opens focus the first item, pointer opens focus the panel and enter on the first arrow. Arrows skip disabled items and wrap, Home/End, Enter/Space activate, Escape closes and restores focus (with `stopPropagation` so a context menu inside a modal doesn't close both), Tab leaves the widget.

Verified by hand in the playground: Escape dismiss with focus restored to the card, click-away dismiss, Shift+F10 open with the first item focused, arrow navigation skipping the disabled item and wrapping at both ends, Home/End.

## No new dependencies

None added, npm or hex.

## Test counts

| Suite | Before (`origin/main` @ 871b2cf) | After |
|---|---|---|
| `mix test` | 913 tests, 0 failures, 1 skipped | **930 tests, 0 failures, 1 skipped** |
| `npm test` | 157 passing | **189 passing** |

`mix format --check-formatted` clean, `mix compile --force --warnings-as-errors` clean, and `mix credo` is **unchanged** against `origin/main` (14 refactoring opportunities, 31 readability issues on both).

## Screenshots

Light and dark, closed and with the menu open at a pointer position - attached below.


mplatts:
## Screenshots

**Menu open at the pointer**

Context menu open, light
Context menu open, dark

**At rest**

Context menu, light
Context menu, dark

### Verified independently

| Check | Result |
|---|---|
| `mix test` | 930 tests, 0 failures, 1 skipped (+17) |
| `npm test` | 189 passing (+32) |
| new dependencies | none |
| slugs : clauses | 61 : 61 |

Good instincts on two calls here. `popover="manual"` rather than `auto` because the UA's light dismiss fires on pointerdown — which is exactly the behaviour `fix/menu-drag-dismiss` existed to remove, so `auto` would have reintroduced a bug the repo already fixed once. And flipping to the other side of the cursor on collision rather than sliding, so the pointer never lands pre-hovered on the first item.

Hand-verified by the author: Escape with focus restored, click-away, `Shift+F10`, arrow nav skipping the disabled item, wrap at both ends, Home/End.

Not verified: long-press on real touch hardware (unit-tested only), and iOS callout suppression is CSS-only and unverified on-device.

nhobes:
Polish pass done (see the polish commit). One item deliberately left for the device eye-test: on iOS, a long-press over the prose demo can raise the text-selection magnifier alongside the menu (-webkit-touch-callout doesn't suppress selection). If it misfires on a real phone, the fix is suppressing user-select on the trigger only while the 500ms press timer is armed - never statically, or the prose demo loses its text selection.

PR #634, opened 2026-08-12
feat(tree): hierarchical tree view with the full APG keyboard map

### LINKED ISSUE #611: Component: Tree
## Summary

A hierarchical tree view (`<.tree>`), the file-explorer staple: nested nodes with expand/collapse, single selection, connecting indent guides, and the full WAI-ARIA TreeView keyboard map. Server-rendered from nested data (a list of maps with `children`), with a slot escape hatch for full control over node rendering. It fills a real gap in the library — we have flat `menu`, `navigation_menu` and `accordion`, but nothing that renders arbitrary-depth hierarchies (file browsers, org structures, nested settings, category pickers). Milestone: 4.17.0.

## References

- reui.io "Tree" (7 examples) — study the scope: basic tree, indent guides, custom icons, controlled expansion, lazy loading.
- shadcn ecosystem tree implementations (e.g. the popular `tree-view` ports) for visual density and chevron/folder/file iconography.
- WAI-ARIA Authoring Practices, TreeView pattern — this is the a11y contract, follow it exactly: https://www.w3.org/WAI/ARIA/apg/patterns/treeview/
- Radix has no Tree primitive; behaviour reference is the APG pattern plus reui. We build a native HEEx/LiveView equivalent from scratch — reference behaviour and variants, never code.

## API sketch

New module `lib/petal_components/tree.ex`, `use Phoenix.Component`. Check `lib/petal_components.ex` for name collisions before exporting (`tree` is currently free).

```elixir
attr :id, :string, required: true, doc: "unique id; the PetalTree hook mounts here for roving focus"

attr :items, :list, default: [],
  doc: """
  nested node maps: %{id: "src", label: "src", icon: "hero-folder", children: [...]}.
  :id (required, string) and :label are the only required keys; :children (list, default [])
  makes a node a branch; :icon overrides the default leaf/branch icon; :disabled (boolean)
  renders the node non-selectable; :lazy (boolean) marks a branch whose children load async.
  Extra keys pass through untouched to the :item slot for custom rendering.
  """

attr :selected, :string, default: nil, doc: "id of the currently selected node (single selection)"

attr :on_select, JS, default: %JS{},
  doc: "JS commands composed with the built-in selection behaviour when a node is chosen; the node also emits a phx-click style event carrying the node id (see Expansion model below)"

attr :default_expanded, :any, default: [],
  doc: "ids of branches expanded at first render, or :all to expand everything"

attr :show_guides, :boolean, default: false,
  doc: "render connecting indent guide lines down each expanded branch (subtle gray ramp)"

attr :class, :any, default: nil, doc: "extra classes for the tree container"
attr :rest, :global

slot :item,
  doc: "optional custom node rendering; receives the node map via :let. Chevron, indent and ARIA wiring stay owned by the component - the slot replaces only the label/icon row content"

slot :empty, doc: "rendered when :items is empty"

slot :loading,
  doc: "rendered inside a :lazy branch while its children are being fetched (default: small spinner row using the loading component)"
```

Follow `lib/petal_components/accordion.ex` for the data-plus-slot composition style (its `entries` + `:item` slot merge) and `lib/petal_components/command.ex` for moduledoc structure (behaviour explained up front, usage example in `@doc`).

### Expansion model — support both, document the trade

1. **Client-side (default)**: chevron toggles via `Phoenix.LiveView.JS` (toggle a `data-expanded` attribute and `aria-expanded`), collapse animated with the CSS `grid-template-rows: 0fr -> 1fr` technique already used by the accordion (see the `pc-accordion-item__content-container` section of `assets/default.css`, around line 3279). Zero server round-trips, state lives in the DOM, lost on patch unless the server mirrors it.
2. **Server-controlled**: user passes an `expanded` MapSet/list and handles an `on_expand`-style event (`phx-click` on the chevron with the node id in `phx-value-id`); required for `:lazy` branches where children arrive via LiveView after the event.

The moduledoc must state the trade plainly: client-side for static trees (fast, no assigns bookkeeping), server-controlled for lazy loading and for trees that must survive `phx-update` patches. Both paths render identical markup and ARIA.

### Hook

A small `PetalTree` hook in `assets/js/petal_components.js` is justified here, per the roving-tabindex requirement — CSS + LiveView.JS cannot move DOM focus between arbitrary nodes on arrow keys. Precedents: `PetalInputOTP`, `PetalCommand`. Scope the hook to keyboard navigation only (roving `tabindex="0"`/`-1`, arrow/Home/End/`*` handling, typeahead optional non-goal); expansion and selection stay in JS commands/server events so the tree still works pointer-only without the hook running.

Before committing to roving tabindex, study how `lib/petal_components/command.ex` uses `aria-activedescendant` (focus stays on one element, highlight is virtual). The APG allows either for trees; roving tabindex is the recommended default for tree views because nodes are real interactive elements (links/buttons), but write one paragraph in the PR explaining the choice.

## Variants & states to build

- Branch node: collapsed / expanded, default chevron + `hero-folder` / `hero-folder-open` icons
- Leaf node: default `hero-document` icon, no chevron (chevron space reserved so labels align)
- Custom icon per node via `:icon` key; fully custom row via `:item` slot
- Selected state (single selection, `aria-selected="true"`, soft primary background per the design doctrine — no persistent focus fills)
- Disabled node (`aria-disabled`, muted, skipped by selection but still focusable per APG)
- Indent guides on/off (`show_guides`) — 1px lines on the gray ramp, visible in light and dark, `aria-hidden` decorative layer
- Lazy branch: `:lazy` node shows the `:loading` slot row while children are pending
- Empty state via `:empty` slot
- Depth: must render correctly at arbitrary nesting (test to at least 4 levels)
- Reduced motion: expand/collapse is instant, rest states complete and legible
- Dark mode throughout; all colour on the gray ramp + semantic tokens, radius via `--pc-radius` conventions

CSS: new `pc-tree` section in `assets/default.css` (`pc-tree`, `pc-tree__node`, `pc-tree__chevron`, `pc-tree__guide`, `pc-tree__node--selected`, etc.), following the doubled-selector and layer conventions already in that file.

## Playground page

New playground page in `dev.exs` with nav entry. Dials (segmented controls via `handle_event`):

- `show_guides` on/off
- `default_expanded`: none / first level / `:all`
- selection target readout (shows the last `on_select` payload)

Example scenarios (shadcn-grade, no lorem ipsum):

1. **File explorer** (hero): a realistic Phoenix project tree (`lib/`, `assets/`, `test/`, nested modules), guides on, folder/file icons, one lazy folder demonstrating async children via a simulated delay.
2. **Org chart list**: people hierarchy with avatar-ish custom `:item` slot rendering, demonstrating slot composition.
3. **Settings nav**: two-level settings sections as a tree with selection driving a highlighted "current page", demonstrating the server-controlled expansion path.

The page should also demonstrate keyboard-only use (mention the keys in the page copy).

## Accessibility

WAI-ARIA TreeView pattern, exactly:

- Container: `role="tree"`, `aria-label` (add an `:aria_label`-friendly path via `rest` or a dedicated attr)
- Nodes: `role="treeitem"`, correct `aria-level`, `aria-setsize`, `aria-posinset` computed per depth/sibling group
- Branches: `aria-expanded="true|false"`; leaves omit `aria-expanded` entirely
- Groups of children: `role="group"`
- Selection: `aria-selected` on the selected node
- Keyboard map (the hook): Down/Up move focus through visible nodes; Right expands a collapsed branch or moves into children; Left collapses an expanded branch or moves to parent; Home/End jump to first/last visible node; Enter/Space select the focused node; `*` expands all siblings of the focused node
- Roving tabindex: exactly one node with `tabindex="0"` at any time; Tab leaves the tree in one stop
- Guides and icons are `aria-hidden="true"`

## Tests

`test/petal/tree_test.exs` using `ComponentCase` + LazyHTML (follow `test/petal/accordion_test.exs` and `test/petal/command_test.exs` for style):

- [ ] Renders nested items to at least 4 levels; DOM nesting matches the data
- [ ] `aria-level`, `aria-setsize`, `aria-posinset` correct across siblings and depths (assert exact values)
- [ ] Branch nodes carry `aria-expanded`; leaf nodes do not
- [ ] `default_expanded: ids` and `default_expanded: :all` produce the expected expanded classes/attributes at render
- [ ] `selected` id gets `aria-selected="true"` and the selected class; others get `aria-selected="false"`
- [ ] Selection emits the expected event payload (node id via `phx-value-*`); `on_select` JS is composed onto the node
- [ ] `show_guides` toggles the guide elements, and guides are `aria-hidden`
- [ ] `:item` slot renders custom content while ARIA wiring stays intact
- [ ] `:empty` slot renders on empty items; `:loading` slot renders for `:lazy` branches
- [ ] Disabled nodes carry `aria-disabled` and no selection binding
- [ ] Exactly one node has `tabindex="0"` at render (roving baseline)
- [ ] JS hook tests in `npm test` for the keyboard map (arrows, Home/End, Enter/Space, `*`)

## Non-goals

- Multi-selection (`aria-multiselectable`) — v1 is single-select only
- Drag-and-drop reordering
- Inline rename/edit of nodes
- Virtualised rendering for huge trees
- Typeahead (type a character to jump) — nice-to-have, explicitly out of v1
- Checkbox trees

## Process

1. Build on a feature branch: `lib/petal_components/tree.ex` + `pc-tree` section in `assets/default.css` + `PetalTree` hook in `assets/js/petal_components.js` + tests + playground page (nav entry, dials, the three examples). CSS/JS live in this repo — no external packages, no new hex deps.
2. Verify in the playground (`mix run dev.exs`, http://localhost:4000): exercise every dial, light + dark, keyboard-only navigation through all three examples, and `prefers-reduced-motion`.
3. Run the suites: `mix test` and `npm test` (the hook needs JS tests). All green before PR.

## Definition of done

- `<.tree>` component with the API above (deviations flagged in the PR), exported via `use PetalComponents` with no name collisions
- `pc-tree` CSS section, both themes, reduced-motion safe
- `PetalTree` hook implementing the full APG keyboard map, registered in the hooks index
- Both expansion models working and documented (client-side JS + server-controlled/lazy)
- Test checklist above green (`mix test` + `npm test`)
- Playground page with dials and the three example scenarios
- PR screenshots light + dark
- Docs-ready attr docs: every attr has a `doc:` string; the moduledoc explains both expansion models and includes usage examples

mplatts:
Closes #611

## Summary

`<.tree>` renders a nested list of maps to arbitrary depth: expand/collapse, single selection, connecting indent guides, and the WAI-ARIA TreeView pattern in full. Five parts per CONTRIBUTING.md, plus a `PetalTree` hook and a `### Unreleased` changelog entry.

1. **Module** `lib/petal_components/tree.ex`, exported from `lib/petal_components.ex` (`tree` was free). Every attr has a `doc:`; the moduledoc carries the node-map reference, both expansion models, the keyboard table and usage examples.
2. **Styles** - a `pc-tree` section appended to `assets/default.css` in its own `@layer components` block, nothing reflowed. `--pc-radius` with the nested-child formula, gray ramp neutrals, `dark:` pairs, soft primary selection, `focus-visible` only, reduced-motion safe.
3. **Tests** - `test/petal/tree_test.exs` (30 specs) and `test/js/tree.test.js` (24 specs, vitest + jsdom).
4. **Showcase** - `lib/petal_components/showcase/tree.ex` plus its registry entry.
5. **Playground** - nav entry under Navigation, slug `tree`, dials plus three scenarios at `/c/tree`. Verified rendering.

## How recursion is handled

A private `tree_node/1` function component renders its own row and, when it is an expanded branch, a `role="group"` list that calls straight back into `tree_node/1`. Depth is whatever the data has - no level cap, no per-level code.

Shared context (the expanded set, selection, the slots, the tree id, guides on/off) is bundled into one `ctx` map built once in `tree/1` and threaded down, so a node carries only what is genuinely per-node: the node map, its level, and its position in its sibling group. `aria-level`, `aria-setsize` and `aria-posinset` fall out of that on the way down.

Indentation is one custom property per row (`--pc-tree-depth`) times a fixed step. The guides are a separate `aria-hidden` absolutely-positioned layer holding one 1px column per ancestor level, offset half a step so each line falls through its parent's chevron - which is why they line up at any depth without per-level rules.

## Both expansion models

Both render identical markup and identical ARIA; only the chevron's `phx-click` differs.

**Client-side (default).** `default_expanded` seeds it; the chevron toggles `data-expanded` and `aria-expanded` with `JS.toggle_attribute`, and the `0fr -> 1fr` grid technique animates the height. `visibility` rides along on that transition so a collapsed branch leaves the accessibility tree rather than sitting there clipped but exposed. No round-trips.

**Server-controlled.** Passing `:expanded` switches modes: the chevron pushes `:on_expand` with the node id in `phx-value-id`. Required for `:lazy` branches, whose children only exist once the server has been asked, and for trees that must survive a patch.

The moduledoc states the trade plainly, including that client-side state is lost on a patch. I deliberately did **not** reach for `phx-update="ignore"` (the accordion's approach): the issue's own wording describes the state as "lost on patch unless the server mirrors it", and ignoring updates would break any tree whose *data* changes.

## Roving tabindex, and why there is a hook

The APG allows either roving tabindex or `aria-activedescendant`; I went with roving tabindex, which the issue names as the recommended default for tree views. `PetalCommand` uses `aria-activedescendant` because its virtual highlight lives in a listbox the user never leaves - focus stays in the text input while the highlight moves, which is the whole point of a combobox. A tree has no input. Its nodes *are* the interactive elements the user is on, so real DOM focus on the `treeitem` is what screen readers and `focus-visible` both expect, and it keeps "which node am I on" in one place (the document) rather than mirrored in an attribute. Exactly one node carries `tabindex="0"` at render - the selected node when it is reachable without expanding anything, otherwise the first visible node - so the tree is one Tab stop.

**The hook is justified by exactly that.** Moving DOM focus between arbitrary nodes on an arrow key is not something CSS or `Phoenix.LiveView.JS` can do: `JS.focus/2` cannot pick "the next node that is visible given which ancestors happen to be expanded". So `PetalTree` does the keyboard map and the tabindex bookkeeping, and **nothing else** - to expand or select it clicks the chevron or the label the component already wired. Consequences:

- Expansion and selection stay in JS commands and server events, so a pointer-only user gets a fully working tree if the hook never mounts.
- The hook is identical under both expansion models; it never needs to know which one is in play.
- It re-establishes the tab stop in `updated()`, since a server patch re-renders `tabindex` from the server's guess, and hands it back to the first visible node if the branch above it closed underneath.

Keyboard map implemented: Down/Up through visible nodes, Right to expand a collapsed branch or descend into an open one, Left to collapse or ascend, Home/End, Enter/Space to select, `*` to expand the focused node's sibling branches. Typeahead is out, per the issue's non-goals.

All small, all deliberate:

1. **`select_event` added.** The sketch has `on_select` (a `JS` struct) and says the node "also emits a phx-click style event carrying the node id", but names no attr for it. `select_event` is that name; the id rides in `phx-value-id` as the test checklist specifies. `on_select` still composes user JS onto the built-in selection behaviour, exactly as sketched.
2. **`expanded` and `on_expand` added.** The "Expansion model" section requires both but does not list them as attrs. `:expanded` being non-nil is what switches the tree to the server-controlled model.
3. **`label` added.** The issue asked for "an `:aria_label`-friendly path via `rest` or a dedicated attr". `label` sets `aria-label`; `rest` still works if you prefer `aria-labelledby`.
4. **`target` added.** `phx-target`, so a tree inside a LiveComponent can route its events.
5. **Selection also fires client-side.** The component flips `aria-selected` and the selected class on click before the server replies, so the highlight is not a round-trip behind. Not in the sketch; it makes selection feel native and degrades cleanly.
6. **The chevron is a `<span>`, not a `<button>`.** A button inside the tree would be a second Tab stop and break the one-Tab-stop requirement. It is `aria-hidden`, and keyboard users reach expansion through Left/Right per the APG. Chevron and label are siblings inside the row, so a chevron click never bubbles into selection.
7. **`aria-labelledby` on each `treeitem`**, pointing at its own label span. Without it, a branch's accessible name is computed from its entire subtree.
8. **Guides render per row rather than per branch**, for the alignment reason above. Behaviour matches the spec: `show_guides` toggles them and they are `aria-hidden`.

## No new dependencies

None added, npm or hex.

## Test counts

| Suite | Before (`origin/main` @ 871b2cf) | After |
|---|---|---|
| `mix test` | 913 tests, 0 failures, 1 skipped | **943 tests, 0 failures, 1 skipped** |
| `npm test` | 157 passing | **181 passing** |

`mix format --check-formatted` clean, `mix compile --force --warnings-as-errors` clean, and `mix credo` is **unchanged** against `origin/main` (14 refactoring opportunities, 31 readability issues on both).

## Verified by hand

In the playground at `/c/tree`, light and dark, on the real page rather than in jsdom:

- **Every key in the map.** Down/Up walk visible nodes and skip collapsed subtrees; Right descends into an open branch and expands a closed one; Left collapses, then ascends; Home/End hit the ends; `*` expanded the focused node's collapsed siblings and left the already-open one alone; Enter selected the focused node and cleared the previous `aria-selected`. Confirmed by reading `tabindex` / `aria-expanded` / `aria-selected` off the live DOM after each keystroke - including across the server round-trip in the explorer scenario, where the collapse and the selection both went through `handle_event` and the hook put the tab stop back afterwards.
- **Pointer expand/collapse in both models**, and the `:lazy` branch: clicking `deps/` showed the loading row, then its children landed 900ms later with their custom icons.
- Guides, selection fill, the disabled node, the folder icon flip on expand, and label alignment all check out in both themes.

## Screenshots

Captured light and dark at 1280x2200 with branches expanded 3-4 levels deep, so the nesting and the guide lines are actually visible rather than a row of collapsed roots. Not committed to the repo (no image files, no `pr-assets`) - I will attach them here before taking this out of draft.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,37 @@
 # Changelog
+
+### Unreleased
+
+#### Added
+
+- **New `tree` component: a hierarchical tree view for arbitrary-depth
+  data.** The file-explorer staple, and the gap between `menu`,
+  `navigation_menu` and `accordion` - none of which render a hierarchy.
+  Feed it a nested list of maps (`%{id:, label:, children: [...]}`) and
+  it renders itself down as far as the data goes: chevrons and
+  folder/document icons on the branches, a reserved chevron column so
+  leaf labels stay aligned, optional connecting indent guides
+  (`show_guides`), and single selection with a soft primary fill.
+  Two expansion models, both rendering identical markup and ARIA:
+  client-side by default (`default_expanded` seeds it, the chevron
+  toggles `data-expanded` with `Phoenix.LiveView.JS` and CSS animates
+  the height, zero round-trips), or server-controlled by passing
+  `expanded` plus an `on_expand` event - which is what `:lazy` branches
+  need, since their children only exist once the server has been asked.
+  A `:loading` row covers the wait, an `:empty` slot covers no data, and
+  an `:item` slot takes over the row content while the component keeps
+  the chevron, the indent and every ARIA attribute.
+  Accessibility is the WAI-ARIA TreeView pattern in full:
+  `role="tree"` / `treeitem` / `group`, computed `aria-level`,
+  `aria-setsize` and `aria-posinset`, `aria-expanded` on branches only,
+  `aria-selected`, `aria-disabled`, and the complete keyboard map
+  (up/down through visible nodes, right to expand or descend, left to
+  collapse or ascend, Home/End, Enter/Space to select, `*` to expand
+  siblings) over a roving tabindex, so the whole tree is one Tab stop.
+  The new `PetalTree` hook does the keyboard work and nothing else -
+  expansion and selection stay in JS commands and server events, so a
+  pointer user gets a working tree even if the hook never mounts.
+
 ### 4.14.0 - 2026-08-11
 
 #### Added
--- a/assets/default.css
+++ b/assets/default.css
@@ -7586,3 +7586,159 @@
       transition-duration: 1ms;
     }
   }
+
+@layer components {
+  /* Tree (WAI-ARIA TreeView)
+
+     Indentation is a single custom property per row (--pc-tree-depth) times a
+     fixed step, so a node at any depth costs one inline style and nothing else.
+     The guide lines ride an absolutely positioned decorative layer over that
+     same step, which is why they line up under the parent chevron without any
+     per-level rules. */
+  .pc-tree {
+    --pc-tree-step: 1.25rem;
+    @apply m-0 list-none p-0 text-sm text-gray-700 select-none dark:text-gray-300;
+  }
+
+  .pc-tree__item {
+    @apply list-none;
+  }
+
+  .pc-tree__row {
+    @apply relative flex min-w-0 items-center gap-1 py-1 pe-2 transition-colors;
+    padding-inline-start: calc(var(--pc-tree-depth, 0) * var(--pc-tree-step));
+    border-radius: max(calc(min(var(--pc-radius, 0.625rem), 1rem) - 0.25rem), 0px);
+  }
+
+  .pc-tree__row:hover {
+    @apply bg-gray-100 dark:bg-gray-400/10;
+  }
+
+  .pc-tree__item--selected > .pc-tree__row,
+  .pc-tree__item--selected > .pc-tree__row:hover {
+    @apply bg-primary-100 text-primary-800 dark:bg-primary-500/15 dark:text-primary-300;
+  }
+
+  /* Focus is the tree item (the li), but the ring belongs on the row - the li
+     also wraps the whole collapsed subtree and would draw a box around it. */
+  .pc-tree__item:focus {
+    @apply outline-hidden;
+  }
+
+  .pc-tree__item:focus-visible > .pc-tree__row {
+    @apply ring-2 ring-primary-500/50;
+  }
+
+  .pc-tree__item--disabled > .pc-tree__row {
+    @apply opacity-50;
+  }
+
+  .pc-tree__item--disabled > .pc-tree__row:hover {
+    @apply bg-transparent;
+  }
+
+  .pc-tree__chevron {
+    @apply flex shrink-0 cursor-pointer items-center justify-center text-gray-400 dark:text-gray-500;
+    width: var(--pc-tree-step);
+    height: var(--pc-tree-step);
+    border-radius: max(calc(min(var(--pc-radius, 0.625rem), 1rem) - 0.375rem), 0px);
+  }
+
+  .pc-tree__chevron:hover {
+    @apply bg-gray-200 text-gray-600 dark:bg-gray-400/17 dark:text-gray-300;
+  }
+
+  /* The leaf chevron is empty on purpose: it reserves the same column so leaf
+     labels line up with their sibling branches. */
+  .pc-tree__chevron--leaf {
+    @apply cursor-default;
+  }
+
+  .pc-tree__chevron--leaf:hover {
+    @apply bg-transparent;
+  }
+
+  .pc-tree__chevron-icon {
+    @apply h-4 w-4 transition-transform duration-200;
+  }
+
+  .pc-tree__item[data-expanded="true"] > .pc-tree__row .pc-tree__chevron-icon {
+    @apply rotate-90;
+  }
+
+  .pc-tree__content {
+    @apply flex min-w-0 flex-1 cursor-pointer items-center gap-1.5 py-0.5;
+  }
+
+  .pc-tree__item--disabled .pc-tree__content {
+    @apply cursor-not-allowed;
+  }
+
+  .pc-tree__icon {
+    @apply h-4 w-4 shrink-0 text-gray-400 dark:text-gray-500;
+  }
+
+  .pc-tree__item--selected > .pc-tree__row .pc-tree__icon {
+    @apply text-primary-600 dark:text-primary-300;
+  }
+
+  .pc-tree__label {
+    @apply truncate;
+  }
+
+  /* Decorative only - one 1px column per ancestor level, offset half a step so
+     each line falls through the middle of its parent's chevron. */
+  .pc-tree__guides {
+    @apply pointer-events-none absolute inset-y-0 start-0 flex;
+    padding-inline-start: calc(var(--pc-tree-step) / 2);
+  }
+
+  .pc-tree__guide {
+    @apply border-s border-gray-200 dark:border-gray-400/17;
+    width: var(--pc-tree-step);
+  }
+
+  /* The 0fr -> 1fr grid trick animates to the children's natural height with no
+     measured pixel value. visibility rides along so a collapsed branch leaves
+     the accessibility tree instead of sitting there clipped but exposed. */
+  .pc-tree__group-wrap {
+    @apply grid;
+    grid-template-rows: 0fr;
+    visibility: hidden;
+    transition:
+      grid-template-rows 200ms ease,
+      visibility 200ms;
+  }
+
+  .pc-tree__item[data-expanded="true"] > .pc-tree__group-wrap {
+    grid-template-rows: 1fr;
+    visibility: visible;
+  }
+
+  .pc-tree__group {
+    @apply m-0 min-h-0 list-none overflow-hidden p-0;
+  }
+
+  .pc-tree__empty {
+    @apply px-2 py-3 text-sm text-gray-500 dark:text-gray-400;
+  }
+
+  .pc-tree__loading {
+    @apply flex list-none items-center gap-2 py-1 pe-2 text-xs text-gray-500 dark:text-gray-400;
+    padding-inline-start: calc(
+      var(--pc-tree-depth, 0) * var(--pc-tree-step) + var(--pc-tree-step)
+    );
+  }
+
+  .pc-tree__spinner {
+    @apply h-3.5 w-3.5 text-gray-400 dark:text-gray-500;
+  }
+
+  @media (prefers-reduced-motion: reduce) {
+    .pc-tree__group-wrap,
+    .pc-tree__chevron-icon,
+    .pc-tree__row {
+      transition-duration: 1ms;
+    }
+  }
+}
--- a/assets/js/petal_components.js
+++ b/assets/js/petal_components.js
@@ -5424,6 +5424,210 @@ export const PetalDataTable = {
   },
 };
 
+// Keyboard navigation for <.tree> (WAI-ARIA TreeView), and nothing else.
+//
+// Roving tabindex is the whole reason this hook exists: the tree must be a
+// single tab stop with exactly one treeitem carrying tabindex="0", and moving
+// DOM focus between arbitrary nodes on an arrow key is not something CSS or
+// LiveView.JS can do. Expansion and selection are deliberately NOT implemented
+// here - the hook clicks the chevron and the label the component already wired,
+// so both expansion models work identically and a pointer user still gets a
+// working tree if the hook never mounts.
+export const PetalTree = {
+  mounted() {
+    this.onKeydown = this.onKeydown.bind(this);
+    this.onFocusIn = this.onFocusIn.bind(this);
+    this.onClick = this.onClick.bind(this);
+    this.el.addEventListener("keydown", this.onKeydown);
+    this.el.addEventListener("focusin", this.onFocusIn);
+    this.el.addEventListener("click", this.onClick);
+    this.syncTabIndex();
+  },
+
+  // A server patch re-renders tabindex from the server's own guess. Put the
+  // tab stop back on whichever node the user was actually on.
+  updated() {
+    this.syncTabIndex();
+  },
+
+  destroyed() {
+    this.el.removeEventListener("keydown", this.onKeydown);
+    this.el.removeEventListener("focusin", this.onFocusIn);
+    this.el.removeEventListener("click", this.onClick);
+  },
+
+  nodes() {
+    return Array.from(this.el.querySelectorAll("[data-pc-tree-node]"));
+  },
+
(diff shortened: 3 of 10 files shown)


mplatts:
## Screenshots

Branches expanded 3-4 levels so the nesting and guide lines are actually visible.

**Light**

Tree, light

**Dark**

Tree, dark

**Lazy branch loading**

Tree lazy loading

### Verified independently

| Check | Result |
|---|---|
| `mix test` | 943 tests, 0 failures, 1 skipped (+30) |
| `npm test` | 181 passing (+24) |
| new dependencies | none |
| slugs : clauses | 61 : 61 |

The guide-line approach is worth noting: a separate `aria-hidden` absolute layer of 1px columns offset half a step, so alignment holds at any depth with no per-level rules. And roving tabindex over `aria-activedescendant` is the right read — a tree's nodes *are* the interactive elements, whereas `PetalCommand`'s virtual model exists because focus has to stay in its input.

Every key in the map was checked against live DOM state (`tabindex`/`aria-expanded`/`aria-selected` after each keystroke), including the tab stop surviving a server round-trip.

**@nhobes — one design opinion wanted:** the guide lines are deliberately subtle (`gray-200` / `gray-400/17`). Clear at full resolution, easy to lose in a scaled screenshot. Your call whether that's right.

nhobes:
New commit from the omnibus eye-test: **row-click expansion is now the default**, matching reui/shadcn and file-explorer convention (VS Code, Finder).

- New `expand_on_click` attr, default `true`: clicking anywhere on a branch row toggles it, and when selection is wired the same click also selects — one click, both outcomes, selection first so the clicked node stays lit. `expand_on_click={false}` restores chevron-only expansion, for rows whose click has another job (link trees, rows with their own controls — the :icons showcase example now demonstrates this with its settings-nav framing).
- The chevron stays a separate hit target that toggles without selecting, in both modes. No stopPropagation games: LiveView dispatches to the *closest* element carrying `phx-click` and stops, so the nested chevron simply wins.
- Keyboard map unchanged: Enter/Space still select without expanding. Since every visible element on an expand-on-click row is now a row click, the select-only command moved to a hidden trigger (`data-pc-tree-select`) that pointers can't reach and the hook's programmatic click can — exactly one marker per row, pinned by test.
- Degradations verified: disabled branches expand but never select; server-controlled trees without `:on_expand` keep an inert row (nothing to push); unselectable trees toggle without highlight.
- Playground gets an `expand_on_click` dial next to `show_guides` for A/B.

Gates: `mix test` 955/0 (exit 0, +9 specs), vitest 183 (exit 0, +1 spec pinning Enter against the new anatomy), `mix format --check-formatted` clean repo-wide.

nhobes:
Eye-test follow-up: the chevron's hover wash fused with the row hover (at depth 0 the step-wide pill was literally flush with the row's left edge — 0px gap, one L-shaped wash). New commit gives it breathing room:

- The painted pill is now one icon square (16x16) centred in the chevron column, with the leftover step split into inline margins — 2px side gaps, 8px top/bottom, so it reads as a chip sitting *in* the hovered row. The column still measures exactly one step: label/guide/row geometry byte-identical before and after (rect-verified).
- Shrinking the wash did not shrink the target: the invisible `::before -inset-1` extension (same pattern as the house close buttons) takes the hit area to 24x24 — a strict superset of the old 20x20 box, proven by an elementFromPoint scan and a functional click outside the old bounds. Leaves skip it (spacers, not controls).
- Verified hovering in light + dark, and against the selected row's primary tint in both themes — the neutral chip stays legible, no selected-state colour rule needed.

`mix test` 955/0 exit 0, format clean.

nhobes:
Maintainer eye-test round 2 on the chevron: the inset chip, while better, still read cramped — 2px side gaps vs 8px vertical is what a 20px column inside a 32px row forces on any painted box. And no file-tree convention expects a box there anyway (VS Code, Finder, GitHub: row highlights, twisty glyph sharpens). New commit drops the hover box entirely — hover is now the icon colour deepening — while keeping the 24x24 invisible hit target and the one-step column geometry. The chevron's role is unchanged: expand/collapse without selecting.

PR #635, opened 2026-08-12
feat(alert_dialog): the confirm/cancel gate before destructive actions

### LINKED ISSUE #600: Component: Alert Dialog
## Summary

A `<.alert_dialog>` component: the confirm/cancel decision dialog you reach for before a destructive or irreversible action ("Delete this account?", "Discard unsaved changes?"). It is deliberately distinct from the existing `<.modal>` (`lib/petal_components/modal.ex`): it uses `role="alertdialog"`, initial focus lands on the least destructive action, Escape cancels, and click-away does NOT close - the friction is the point. Ships with a default (confirm) variant and a destructive variant with a danger-styled confirm button.

Target release: 4.15.0.

## References

Study these for scope and interaction design only - we build a native HEEx/LiveView equivalent from scratch, never port their code:

- reui.io Alert Dialog (14 examples) - https://reui.io/docs/alert-dialog
- Radix AlertDialog - https://www.radix-ui.com/primitives/docs/components/alert-dialog
- shadcn/ui Alert Dialog - https://ui.shadcn.com/docs/components/alert-dialog
- WAI-ARIA alertdialog pattern - https://www.w3.org/WAI/ARIA/apg/patterns/alertdialog/

In this repo, study before writing any code:

- `lib/petal_components/modal.ex` - the `JS.show`/`JS.hide` show/hide machinery (`show_modal/2`, `hide_modal/1`), the `data-cancel` / `data-cancel-default` `JS.exec` indirection, and `aria-labelledby` wiring.
- `lib/petal_components/command.ex` (`command_dialog/1`) - the native `<dialog>` approach: top layer, focus containment, backdrop and Escape for free, opened via `JS.dispatch` (`open_command/2`) so it works from any element. Also note the `command_trigger` moduledoc explaining why a hook-mounted trigger works on dead views where `phx-click` JS commands do not - relevant if you add a `:trigger` slot.
- `assets/default.css` around `.pc-command-dialog` (line ~2100) for the floating-panel dialog surface and `::backdrop` styling in light and dark.

**Implementation decision to make early**: native `<dialog>` (as `command_dialog` does) versus the modal's overlay-div machinery. Native `<dialog>` is recommended - `showModal()` gives focus containment and top layer for free, and you suppress the default Escape/backdrop-close behaviour selectively (cancel on Escape is wanted; close on backdrop click is not, so a `cancel` event listener or small inline handling is needed to swallow light dismiss). Whichever you pick, flag it in the PR with reasoning.

## API sketch

Starting proposal - refine as needed and flag deviations in the PR.

```elixir
attr :id, :string, required: true, doc: "unique id; open with open_alert_dialog/2"
attr :title, :string, required: true, doc: "the question being asked; the dialog's accessible name"

attr :description, :string,
  default: nil,
  doc: "supporting copy explaining the consequence; wired to aria-describedby. Omit when the inner_block carries the body"

attr :variant, :string,
  default: "default",
  values: ["default", "destructive"],
  doc: "destructive styles the confirm button on the danger ramp and adds a danger icon accent"

attr :confirm_label, :string, default: "Continue", doc: "confirm button text"
attr :cancel_label, :string, default: "Cancel", doc: "cancel button text"

attr :on_confirm, Phoenix.LiveView.JS,
  default: %JS{},
  doc: "JS commands run when the user confirms (e.g. JS.push(\"delete_account\")). The dialog closes after"

attr :on_cancel, Phoenix.LiveView.JS,
  default: %JS{},
  doc: "JS commands run on cancel (cancel button or Escape). The dialog closes after"

attr :rest, :global

slot :inner_block, required: false, doc: "custom body content rendered below the description"
slot :icon, required: false, doc: "optional icon rendered in the accent slot; destructive variant supplies a default danger icon when this is empty"
slot :trigger, required: false, doc: "optional trigger button rendered adjacent to the dialog, pre-wired to open it"
```

Plus a public helper composable onto any element, following the `open_command/2` precedent:

```elixir
def open_alert_dialog(js \\ %JS{}, id)
```

Semantics to honour (from the WAI-ARIA pattern and the design intent):

- `role="alertdialog"`, `aria-modal="true"`, `aria-labelledby` pointing at the title, `aria-describedby` pointing at the description when present.
- On open, focus lands on the **cancel button** (the least destructive action), not confirm.
- Escape cancels (runs `on_cancel`, closes). Backdrop / click-away does **not** close.
- Focus is contained while open and restored to the opener on close.
- Confirm and cancel both close the dialog after running their JS.

## Variants & states to build

- [ ] `default` - neutral confirm (primary button treatment)
- [ ] `destructive` - confirm button on the danger ramp, danger icon accent (overridable via the `:icon` slot)
- [ ] With and without `description`
- [ ] With custom `inner_block` body (e.g. a count summary, a checklist)
- [ ] With and without the `:icon` slot
- [ ] Long-content state: body scrolls inside the panel, header and action row stay visible, page behind does not scroll
- [ ] Light and dark mode on the floating-panel surface (`bg-white dark:bg-gray-900`, real borders both modes)
- [ ] Radius via the `--pc-radius` token (small-panel rule: `min(token, 16px)`; nested buttons use the nested-child radius rule)
- [ ] Open/close transitions as CSS (respecting `prefers-reduced-motion`; rest state complete and legible)

New `pc-alert-dialog` section in `assets/default.css` (`pc-alert-dialog`, `pc-alert-dialog__title`, `pc-alert-dialog__description`, `pc-alert-dialog__actions`, `pc-alert-dialog--destructive`, etc.).

## Playground page

Nav entry + hero example, then dials wired via `handle_event` segmented controls:

- **Variant**: default / destructive
- **Description**: with / without
- **Content length**: short / long (proves the overflow behaviour)

1. **Delete account** - destructive variant, danger icon, description spelling out the consequence, confirm pushes an event the playground acknowledges (e.g. a toast).
2. **Unsaved changes** - default variant triggered from a mock editor's "Back" button: "Discard unsaved changes?" with Keep editing / Discard.
3. **Bulk delete with count** - destructive, custom `inner_block` showing "You are about to delete 14 invoices" with the count rendered from an assign, proving the body slot composes with live data.

The page should demonstrate keyboard-only use: open via trigger, see focus land on Cancel, Tab between actions, Escape to cancel.

## Accessibility

Follow the WAI-ARIA alertdialog pattern exactly:

- `role="alertdialog"` + `aria-modal="true"` on the dialog element
- `aria-labelledby` -> title element id; `aria-describedby` -> description element id (attribute omitted when no description)
- Initial focus on the cancel button; focus contained while open; focus restored to the invoking element on close
- Keyboard map: `Tab`/`Shift+Tab` cycle the dialog's focusables; `Escape` cancels; `Enter`/`Space` activate the focused button
- Click on the backdrop does nothing (deliberate friction - the user must choose)
- Backdrop/decorative layers carry `aria-hidden="true"` where applicable
- Reduced motion: transitions collapse to instant show/hide

## Tests

`test/petal/alert_dialog_test.exs` using `ComponentCase` + LazyHTML:

- [ ] Renders `role="alertdialog"` and `aria-modal="true"`
- [ ] `aria-labelledby` resolves to the rendered title id; title text renders
- [ ] `aria-describedby` resolves to the description id when `description` is set, and is absent when not
- [ ] Default variant renders primary-treatment confirm; `variant="destructive"` renders the danger classes and default danger icon
- [ ] `:icon` slot overrides the default icon; no icon renders for default variant without the slot
- [ ] `confirm_label` / `cancel_label` defaults and overrides render
- [ ] `on_confirm` / `on_cancel` JS is attached to the correct buttons (assert the `phx-click` payloads)
- [ ] `inner_block` content renders inside the body
- [ ] `:trigger` slot renders and is wired to open the dialog (assert the open command/attribute)
- [ ] No click-away close binding exists (assert the absence of `phx-click-away` or equivalent light-dismiss wiring)
- [ ] Escape wiring cancels (assert the keydown/cancel binding)
- [ ] `class` and `rest` pass-through

If a JS hook ends up necessary (it should not - CSS + `Phoenix.LiveView.JS` + native `<dialog>` should cover this), it needs `npm test` coverage and justification in the PR.

## Non-goals

- Not a replacement or refactor of `<.modal>` - it stays untouched. Alert dialog is a separate, narrower component.
- No async/loading confirm state (spinner-in-button while the event round-trips) in v1 - users can compose that with `<.button loading>` in a custom body if needed.
- No stacking/nesting of alert dialogs.
- No form embedding - the dialog is a two-button decision. Anything needing inputs is a `<.modal>`.
- No Alpine.js support - LiveView.JS + CSS only, per current policy.
- No new hex or npm dependencies.

## Process

1. Build on a feature branch: `lib/petal_components/alert_dialog.ex` + `pc-alert-dialog` section in `assets/default.css` + tests + playground page (nav entry, dials, the three scenarios). Register the import in `lib/petal_components.ex` (check for name collisions with `Phoenix.Component` and core_components first - `alert_dialog/1` is clear today).
2. Verify in the playground (`mix run dev.exs`, http://localhost:4000) - exercise every dial, light + dark, keyboard-only (focus lands on Cancel, Escape cancels, click-away does nothing), and `prefers-reduced-motion`.
3. Run the suites: `mix test` (and `npm test` only if a hook was added). All green before PR.

## Definition of done

- `alert_dialog/1` (plus `open_alert_dialog/2` helper) implemented per the semantics above, exported via `use PetalComponents`
- `pc-alert-dialog` CSS section in `assets/default.css`, light + dark, radius token respected, reduced-motion honoured
- Full test coverage per the checklist, suite green
- Playground page with the three dials and three scenarios, keyboard demo included
- PR with light + dark screenshots, automated review passing
- Docs-ready: every attr has a `doc:` string; the moduledoc explains when to use alert_dialog vs modal and includes usage examples for both variants

mplatts:
Closes #600

## Summary

`<.alert_dialog>` is the two-answer question you ask before something irreversible. `<.modal>` is the general-purpose overlay and it lets you out cheaply, which is the wrong behaviour when the next click deletes an account, so this inverts it: `role="alertdialog"`, focus opens on cancel, Escape cancels, backdrop clicks do nothing.

What shipped:

- `lib/petal_components/alert_dialog.ex` - `alert_dialog/1` plus the `open_alert_dialog/2` helper, exported via `use PetalComponents`. Every attr carries a `doc:`; the moduledoc covers alert_dialog vs modal and both variants.
- `pc-alert-dialog` section appended to `assets/default.css`, inside `@layer components`. Nothing above it moved.
- `test/petal/alert_dialog_test.exs` - 32 tests.
- `test/js/alert_dialog.test.js` - 15 tests for the two hooks.
- `PetalComponents.Showcase.AlertDialog` with four examples, registered in `showcase/registry.ex`.
- Playground page at `/c/alert-dialog` with the three dials, a keyboard map, and the showcase examples.
- `### Unreleased` / `#### Added` entry at the top of `CHANGELOG.md`.

## Native `<dialog>` vs the modal's overlay machinery

**Native `<dialog>`**, the same call `command_dialog/1` made. Reasoning:

- `showModal()` gives the top layer, **focus containment**, and **focus restoration to the opener** on close. The WAI-ARIA alertdialog pattern requires all three, and `<.modal>` does not actually contain focus today. Hand-rolling a focus trap in an overlay div would have been the largest and least reliable part of this component.
- The suppression the issue flagged turned out not to be needed. A native modal dialog **already ignores backdrop clicks** - light dismiss is opt-in via `closedby="any"`, which this component simply never sets. So "click-away does not close" is the default rather than something fought for. The only native behaviour intercepted is Escape, and only so it runs your `on_cancel` rather than closing silently.
- `::backdrop` replaces the overlay div and the `body { overflow: hidden }` juggling entirely.

## Focus, Escape and backdrop

- **Focus**: cancel carries `autofocus` (which `showModal()` honours) *and* the hook focuses it explicitly on open, so the least destructive action is the default however a consumer's body content is composed. Cancel is also rendered before confirm, so Tab order reaches the safe action first. Focus restores to the opener via the native element.
- **Escape**: the hook listens for the native `cancel` event, calls `preventDefault()`, and clicks the cancel button. Escape and a Cancel click therefore run the identical `on_cancel` path - one code path, not two that can drift.
- **Backdrop**: nothing is wired. No `phx-click-away`, no click handler on the dialog element, no `closedby`. A test asserts the absence rather than trusting it stays absent.

## The hook, and why

House rule is CSS-first, so naming what forced it: `showModal()` and `close()` are DOM methods with no `Phoenix.LiveView.JS` equivalent - there is no JS command that invokes a DOM method - and the native `cancel` event has to be intercepted in JS to reroute Escape through `on_cancel`. Everything else (surface, radius token, transitions, `prefers-reduced-motion`, dark mode) is CSS.

`PetalAlertDialogTrigger` is a second ~10-line hook backing the `:trigger` slot, mirroring `PetalCommandTrigger` for the same documented reason: hooks mount on dead views, `phx-click` JS commands only run inside a LiveView.

Both are unit-tested in `test/js/alert_dialog.test.js`. **No new dependencies**, hex or npm.

**None** on the public surface - every attr, slot, default and the `open_alert_dialog/2` signature match the sketch as written.

Two additions the sketch did not spell out, both internal:

- The CSS section grew `pc-alert-dialog__panel`, `__header`, `__body`, `__content`, `__icon`, `__icon-svg`, `__cancel`, `__confirm` and `__trigger` alongside the names the issue listed. The header/body/actions split is what keeps the title and the action row visible while a long body scrolls.
- `data-pc-alert-dialog-close` / `-cancel` / `-confirm` markers on the action buttons are how the hook finds them.

One thing worth a maintainer's eye: the confirm and cancel buttons are `<.button>` (`color="danger"`/`"primary"` and `color="gray" variant="outline"`) rather than bespoke `pc-alert-dialog` button styles, so they inherit the theme rail and the radius rules for free. Say the word if you would rather they were standalone.

## Verification

| Check | Result |
|---|---|
| `mix format --check-formatted` | pass |
| `mix compile --force --warnings-as-errors` | pass |
| `mix credo` | **zero** new entries vs `origin/main` (diffed `--all --format=oneline`) |
| `mix test` | 913 -> **945**, 0 failures, 1 skipped |
| `npm test` | 157 -> **172** passing |

Verified by hand in the playground, light and dark, at `/c/alert-dialog`:

- **Initial focus lands on Cancel** - opened the dialog and pressed Enter without touching anything else; the page acknowledged `cancel`, not `confirm`.
- **Escape cancels** - closes the dialog and acknowledges `cancel`, i.e. it runs `on_cancel` rather than closing silently.
- **Backdrop click does not close** - with the dialog open the rest of the page is inert (the a11y tree contains only the dialog's own heading and two buttons) and the dialog stays up.
- **Tab** moves from Cancel to the confirm button with a visible `focus-visible` ring, and does not escape the dialog.
- **Long content** scrolls inside the body while the title and the action row stay fixed.

One bug found and fixed during that pass: a percentage `max-height` on the panel resolved against the dialog's `auto` height and was ignored, so a long body overflowed and clipped its own header. Both now carry the same explicit `--pc-alert-dialog-max-h`.

Light and dark screenshots of the playground page were captured during that pass and are available on request - I have not pushed image assets to the repo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,37 @@
 # Changelog
+### Unreleased
+
+#### Added
+
+- **New `alert_dialog` component: the confirm/cancel question you ask
+  before something irreversible.** `<.modal>` is the general-purpose
+  overlay, and it lets you out cheaply - close button, click away, done.
+  That is the wrong behaviour when the next click deletes an account, so
+  `alert_dialog` inverts it. It carries `role="alertdialog"` so assistive
+  tech announces it as an interruption that wants a decision, opens with
+  focus on the **cancel** button rather than the first focusable, cancels
+  on Escape, and ignores backdrop clicks entirely. The friction is the
+  feature: there are two answers and you have to pick one. A `default`
+  variant renders the primary confirm treatment; `destructive` moves the
+  confirm button to the danger ramp and adds a danger icon accent, which
+  the `:icon` slot overrides. `description` is the one-line consequence
+  and wires itself to `aria-describedby`; the body slot composes live
+  data underneath it, so "you are about to delete 14 invoices" reads off
+  an assign instead of a hardcoded string. Long bodies scroll inside the
+  panel while the title and the action row stay put. Open it from
+  anywhere with `open_alert_dialog/2`, or drop an opener in the
+  `:trigger` slot and skip repeating the id.
+
+  Built on the native `<dialog>`, the same call `command_dialog` makes:
+  `showModal()` supplies the top layer, focus containment and focus
+  restoration to the opener for free, and a native modal dialog already
+  ignores backdrop clicks, which is exactly what this pattern wants. The
+  hook is small and only covers what `Phoenix.LiveView.JS` cannot reach -
+  `showModal()`/`close()` are DOM methods with no JS-command equivalent,
+  and Escape's native `cancel` event has to be intercepted so it runs
+  your `on_cancel` instead of quietly closing. The surface, the
+  transitions and the reduced-motion behaviour are all CSS.
+
 ### 4.14.0 - 2026-08-11
 
 #### Added
--- a/assets/default.css
+++ b/assets/default.css
@@ -7586,3 +7586,114 @@
       transition-duration: 1ms;
     }
   }
+
+/* Alert dialog - the two-answer decision before something destructive.
+   Built on the native <dialog>, so the top layer, focus containment and
+   ::backdrop come from the element. Deliberately narrower than the modal:
+   one question, one accent icon, one action row that stays put while a long
+   body scrolls under it. */
+
+@layer components {
+  .pc-alert-dialog {
+    /* The panel needs a DEFINITE height to bound its scrolling body: a
+       percentage max-height on the panel would resolve against the dialog's
+       auto height and be ignored, so the panel would overflow and clip its
+       own header. Both carry the same explicit cap instead. */
+    --pc-alert-dialog-max-h: min(32rem, calc(100dvh - 4rem));
+    @apply fixed p-0 overflow-hidden shadow-lg;
+    inset: 0;
+    margin: auto;
+    width: min(30rem, calc(100vw - 2rem));
+    max-height: var(--pc-alert-dialog-max-h);
+    color: inherit;
+    border: 1px solid var(--color-gray-200);
+    background: var(--color-white, #fff);
+    border-radius: min(var(--pc-radius, 0.625rem), 1rem);
+  }
+
+  .dark .pc-alert-dialog {
+    border-color: color-mix(in srgb, white 15%, transparent);
+    background: var(--color-gray-900);
+  }
+
+  .pc-alert-dialog::backdrop {
+    background: rgb(0 0 0 / 0.5);
+  }
+
+  .pc-alert-dialog__panel {
+    @apply flex flex-col;
+    max-height: var(--pc-alert-dialog-max-h);
+  }
+
+  .pc-alert-dialog__header {
+    @apply flex items-start gap-3 shrink-0 px-6 pt-6;
+  }
+
+  .pc-alert-dialog__icon {
+    @apply flex items-center justify-center shrink-0 w-9 h-9 text-gray-500 dark:text-gray-400;
+    background: var(--color-gray-100);
+    border-radius: min(calc(var(--pc-radius, 0.625rem) * 0.8), 0.625rem);
+  }
+
+  .dark .pc-alert-dialog__icon {
+    background: color-mix(in srgb, white 8%, transparent);
+  }
+
+  .pc-alert-dialog--destructive .pc-alert-dialog__icon {
+    @apply text-danger-600 dark:text-danger-400;
+    background: color-mix(in srgb, var(--color-danger-500) 12%, transparent);
+  }
+
+  .pc-alert-dialog__icon-svg {
+    @apply w-5 h-5;
+  }
+
+  .pc-alert-dialog__title {
+    @apply self-center text-base font-semibold text-gray-900 dark:text-gray-100;
+  }
+
+  .pc-alert-dialog__body {
+    @apply flex-1 min-h-0 overflow-y-auto px-6 pt-2;
+  }
+
+  .pc-alert-dialog__description {
+    @apply text-sm text-gray-500 dark:text-gray-400;
+  }
+
+  .pc-alert-dialog__content {
+    @apply mt-3 text-sm text-gray-600 dark:text-gray-300;
+  }
+
+  .pc-alert-dialog__actions {
+    @apply flex flex-col-reverse gap-2 shrink-0 px-6 pt-5 pb-6;
+    @apply sm:flex-row sm:justify-end;
+  }
+
+  .pc-alert-dialog__actions .pc-alert-dialog__cancel,
+  .pc-alert-dialog__actions .pc-alert-dialog__confirm {
+    @apply justify-center sm:w-auto w-full;
+  }
+
+  @media (prefers-reduced-motion: no-preference) {
+    .pc-alert-dialog[open] {
+      animation: pc-alert-dialog-in 0.15s ease-out;
+    }
+
+    .pc-alert-dialog[open]::backdrop {
+      animation: pc-alert-dialog-backdrop-in 0.15s ease-out;
+    }
+  }
+
+  @keyframes pc-alert-dialog-in {
+    from {
+      opacity: 0;
+      transform: scale(0.96);
+    }
+  }
+
+  @keyframes pc-alert-dialog-backdrop-in {
+    from {
+      opacity: 0;
+    }
+  }
+}
--- a/assets/js/petal_components.js
+++ b/assets/js/petal_components.js
@@ -1909,6 +1909,81 @@ export const PetalCommandDialog = {
   },
 };
 
+// The alertdialog: a native <dialog> asking one question with two answers.
+//
+// The hook exists for the three things CSS and Phoenix.LiveView.JS cannot do:
+// showModal()/close() are DOM methods with no JS-command equivalent; initial
+// focus has to land on cancel (the least destructive action) rather than the
+// first focusable; and Escape's native `cancel` event must be intercepted so
+// it runs the component's on_cancel JS instead of quietly closing.
+//
+// Note what is deliberately absent: there is NO backdrop click handler. A
+// native modal dialog ignores backdrop clicks, so click-away does nothing -
+// which is the behaviour the WAI-ARIA alertdialog pattern asks for. Focus
+// containment and focus restoration to the opener also come from the native
+// element, not from here.
+export const PetalAlertDialog = {
+  mounted() {
+    this.onOpen = () => this.open();
+    // Escape. Swallow the native close and route through the cancel button so
+    // Escape and a Cancel click run the exact same on_cancel JS - one path.
+    this.onCancel = (e) => {
+      e.preventDefault();
+      const cancel = this.cancelButton();
+      cancel ? cancel.click() : this.close();
+    };
+    // Both actions close the dialog after their phx-click JS has run.
+    this.onAction = (e) => {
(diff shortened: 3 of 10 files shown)


mplatts:
## Screenshots

**Dialog open**

Alert dialog open, light
Alert dialog open, dark

**Page**

Alert dialog page, light
Alert dialog page, dark

**Behaviour proofs** — Escape, backdrop click, Enter landing on Cancel, long-content scroll

Escape
Backdrop click
Enter on cancel
Long content scroll

### Verified independently

| Check | Result |
|---|---|
| `mix test` | 945 tests, 0 failures, 1 skipped (+32) |
| `npm test` | 172 passing (+15) |
| new dependencies | none |
| slugs : clauses | 61 : 61 |

## A correction to the issue's premise

The brief expected backdrop-close would need suppressing. It doesn't — a native modal `&lt;dialog&gt;` already ignores backdrop clicks, and light dismiss is opt-in via `closedby="any"`, which this component never sets. So the only native behaviour intercepted is Escape's `cancel` event, rerouted through the cancel button so Escape and a Cancel click share one `on_cancel` path. Less machinery than the issue anticipated.

The a11y requirements were verified functionally rather than visually: opening and pressing Enter without touching anything acknowledged `cancel` (proving initial focus is on the least destructive action), and with the dialog open the a11y tree contains only its heading and two buttons.

**A real bug was caught and fixed mid-pass**: with long content the panel's percentage `max-height` resolved against the dialog's `auto` height, was ignored, and the panel overflowed and clipped its own header. Found by opening a zoomed screenshot and looking at it. Both dialog and panel now carry the same explicit `--pc-alert-dialog-max-h`.

**@nhobes — two calls for you:** confirm/cancel are composed from `&lt;.button&gt;` rather than bespoke `pc-alert-dialog` button styles (buys the theme rail and nested-radius rules free, but you may want it the other way); and this page carries both dials *and* registry examples, where recent pages like modal and command are registry-only.

nhobes:
Polish pass done. One follow-up flagged outside this PR's scope: **PetalCommandDialog shares the missing background scroll lock** - the same body overflow-hidden treatment should mirror there (the alert dialog's landed hook-side, so the command dialog wants the same two lines + spec). Happy to do it as a tiny separate PR after this merges.

PR #636, opened 2026-08-12
feat(slider): single and dual range slider on native inputs

### LINKED ISSUE #608: Component: Slider
Target milestone: **4.16.0**

## Summary

A net-new `<.slider>` component: a fully styled range slider built on native `<input type="range">`, in both single-thumb and dual-thumb (range) forms. Today the library only has a native range input routed through `field.ex` (with a primary fill trick and a separate `range-dual` type). This component promotes the slider to a first-class control with marks/ticks, step snapping, a value tooltip anchored to the thumb, inline value display, vertical orientation, and sizes, and it absorbs the dual-thumb display currently living in `field.ex`/`input.ex` (which gets deprecated in favour of `<.slider>`).

Building on native range inputs keeps keyboard and screen-reader behaviour free: arrows/Home/End/PageUp/PageDown, `aria-valuemin/max/now`, and form posting all come from the browser.

## References

- reui.io "Slider" (12 examples): study for marks, tooltip-on-drag, vertical, and dual-range interaction design
- Radix Slider: https://www.radix-ui.com/primitives/docs/components/slider (behaviour and keyboard contract only; we build native HEEx/LiveView from scratch, never port their code)
- WAI-ARIA Slider pattern: https://www.w3.org/WAI/ARIA/apg/patterns/slider/ (mostly satisfied for free by native inputs)

**Prior art in this repo — study before writing anything:**

- `lib/petal_components/input.ex` — the existing `type="range"` clause (`fill` attr, `pc-range-input` styling) and the `type="range-dual"` clause (two overlaid inputs, `data-pc-range-min/max/track/display` markers, `calculate_slider_position/3`, server-rendered initial track position)
- `lib/petal_components/field.ex` — the `range` and `range-dual` field clauses (`min_field`/`max_field`, `range_min`/`range_max`, `value_prefix`/`value_suffix`, error merging from both fields)
- `assets/js/petal_components.js` — the `PetalRangeFill` hook (keeps `--pc-range-fill` in sync for the webkit fill paint) and `PetalDualRangeSlider` hook (thumb ordering, z-index lift when thumbs meet, track sync, display label)
- `assets/default.css` — the `pc-range-input` and `pc-dual-range` sections (track/thumb pseudo-element styling for both webkit and Firefox, dark-mode track via gray-400 alpha mix)

Much of the mechanics already exist; this issue is about consolidating them into one coherent, richer component.

## API sketch

New module `lib/petal_components/slider.ex`, `use Phoenix.Component`, exported from `lib/petal_components.ex` (check for name collisions there first — `slider` is currently free).

```elixir
attr :field, :any, default: nil,
  doc: "a Phoenix.HTML.FormField for single-thumb use; sets name, id and value"

attr :min_field, :any, default: nil,
  doc: "FormField for the lower value; dual-thumb mode when set with max_field"

attr :max_field, :any, default: nil,
  doc: "FormField for the upper value; dual-thumb mode when set with min_field"

attr :name, :string, default: nil, doc: "input name when not using field"
attr :id, :string, default: nil
attr :value, :any, default: nil, doc: "current value (single-thumb)"

attr :values, :list, default: nil,
  doc: "[min, max] current values (dual-thumb, when not using min_field/max_field)"

attr :min, :any, default: 0, doc: "lower bound"
attr :max, :any, default: 100, doc: "upper bound"
attr :step, :any, default: 1, doc: "step increment; values snap to it natively"

attr :marks, :list, default: [],
  doc: ~s|labelled stops rendered under the track, e.g. [%{value: 0, label: "Min"}, %{value: 50, label: ""}]; an empty label renders a tick only|

attr :show_value, :string, default: "none", values: ["tooltip", "inline", "none"],
  doc: "tooltip shows the value in a bubble above the thumb while dragging/focused; inline renders it beside the track"

attr :value_prefix, :string, default: "", doc: ~s|prepended to displayed values, e.g. "$"|
attr :value_suffix, :string, default: "", doc: ~s|appended to displayed values, e.g. "%"|

attr :orientation, :string, default: "horizontal", values: ["horizontal", "vertical"]
attr :size, :string, default: "md", values: ["sm", "md", "lg"],
  doc: "track thickness and thumb diameter"

attr :class, :any, default: nil, doc: "CSS class for the outer wrapper"
attr :rest, :global
```

No slots needed. Single vs dual mode is inferred: `min_field`/`max_field` or `values` means dual, otherwise single. Raise a clear `ArgumentError` on ambiguous combinations (e.g. `field` plus `values`).

Attr style follows `lib/petal_components/progress.ex` (every attr has a `doc:` string, `values:` lists for enums).

### Structure

Single-thumb: one `<input type="range">` inside a `pc-slider` wrapper. Dual-thumb: two overlaid `<input type="range">` elements (the standard technique, as the existing `pc-dual-range` markup does), each posting its own form name, with a shared styled track and a primary fill segment.

The fill/thumb position is driven by a `--pc-slider-pct` custom property (per thumb: `--pc-slider-pct-min` / `--pc-slider-pct-max` in dual mode) set server-side for the initial render (see `calculate_slider_position/3` in `input.ex`) and kept in sync by the hook on `input` events. Tooltip and inline value displays anchor off the same property, so tooltip positioning is pure CSS (`left: var(--pc-slider-pct)` on an absolutely positioned bubble) with no measurement JS.

### Hook

One small hook, `PetalSlider`, in `assets/js/petal_components.js`, replacing/superseding `PetalRangeFill` + `PetalDualRangeSlider` for this component (leave those hooks in place for the deprecated field types until removal):

- Syncs `--pc-slider-pct*` on input
- Dual mode: enforces thumb ordering (clamp min <= max) and the z-index lift when thumbs meet, exactly as `PetalDualRangeSlider` does today
- Updates the tooltip/inline value text (with prefix/suffix, trailing-zero-stripped formatting as in `syncDisplay`)
- Emits a `petal:slider-change` CustomEvent on the wrapper with `detail: {value}` (single) or `detail: {values: [min, max]}` (dual), bubbling, so LiveView can listen via `phx-hook`-free `window.addEventListener` or `JS.dispatch` patterns

### Deprecation of the field.ex dual display

- `type="range-dual"` in `field.ex`/`input.ex` stays working but its docs point to `<.slider>`; add a `@doc` deprecation note (no compile warnings — plenty of apps use it)
- The single `type="range"` field also remains (it is the plain unstyled-ish native input); `<.slider>` is the styled component

## Variants & states to build

- [ ] Single-thumb, horizontal (default)
- [ ] Dual-thumb, horizontal
- [ ] Sizes: `sm`, `md`, `lg`
- [ ] Vertical orientation (single and dual; uses `writing-mode`/`appearance: slider-vertical` fallbacks — pick the approach that works in current Chrome/Firefox/Safari and document it in the CSS comment)
- [ ] Marks: tick marks positioned along the track, optional labels beneath (beside, when vertical); marks inside the filled region take the on-primary treatment
- [ ] `show_value`: `tooltip` (visible while dragging or focus-visible), `inline`, `none`
- [ ] Step snapping (native `step` attr; marks do not have to align with steps, but the playground should show a snapped example)
- [ ] Disabled (reduced opacity, no pointer events, native `disabled` on inputs)
- [ ] Dark mode throughout — track rides the gray ramp (see the existing gray-400 alpha mix in `pc-range-input`), fill and thumb use primary tokens, thumb sits on the field surface
- [ ] Focus: `focus-visible` ring in primary on the thumb only, never a persistent fill

## Playground page

New page in `dev.exs` with a nav entry, following the existing playground page pattern.

- Mode: single / dual
- Marks: off / on
- `show_value`: none / tooltip / inline
- Orientation: horizontal / vertical
- Step: 1 / 5 / 25
- Size: sm / md / lg

Scenarios (shadcn-grade, real product moments):

1. **Price-range filter** — dual-thumb with `$` prefix, marks at 0 / 500 / 1000, inside a filter card, wired to a LiveView assign via `petal:slider-change` so a result count updates live
2. **Volume control** — single-thumb with inline value and a speaker icon, `%` suffix
3. **Year picker** — single-thumb, `min: 1990, max: 2030, step: 5`, marks with labels at each step, tooltip display

Demonstrate keyboard operation on the page (a hint line: arrows, Home/End, PageUp/PageDown).

## Accessibility

The WAI-ARIA slider pattern, satisfied primarily by the native inputs:

- Native `<input type="range">` carries `role`, `aria-valuemin/max/now` and full keyboard support (Left/Right/Up/Down arrows, Home, End, PageUp/PageDown) for free — do not re-implement any of it
- Each input gets an accessible name: `aria-label` derived from the field/label, and in dual mode distinct labels ("Minimum price", "Maximum price" style — derive from the humanised field name plus "minimum"/"maximum", overridable via `rest`)
- Tooltip and mark labels are presentational duplicates of information the input already exposes; mark tick layers get `aria-hidden="true"`
- Vertical orientation must still announce correctly (native vertical range inputs do; verify with VoiceOver once)
- `prefers-reduced-motion`: no animated tooltip entrance beyond a simple fade, and the rest state (no tooltip) is complete and legible

## Tests

`test/petal/slider_test.exs` using `ComponentCase` + LazyHTML:

- [ ] Single-thumb renders one `input[type=range]` with correct `name`, `id`, `min`, `max`, `step`, `value`
- [ ] Dual-thumb renders two inputs posting BOTH form names (assert each input's `name` from `min_field`/`max_field`, and from `values` + explicit names)
- [ ] Clamp/order invariant: initial values outside `[min, max]` are clamped server-side; dual values render in order (min input <= max input) even when given reversed
- [ ] Marks render one tick per entry, labels only where non-empty, positioned via the percentage style
- [ ] `show_value` modes: tooltip element present/absent, inline element present/absent, prefix/suffix applied
- [ ] Orientation and size classes applied (`pc-slider--vertical`, `pc-slider--sm` etc.)
- [ ] Disabled: `disabled` attr on every input, modifier class present
- [ ] A11y: native inputs carry `min`/`max` and `value`; assert `aria-label` present on each input in dual mode
- [ ] Field integration: errors from the form field(s) render via the standard field error markup if `<.slider>` is used inside `<.field_wrapper>` patterns (match how `field.ex` merges min/max errors)
- [ ] Hook tests in `npm test` for `PetalSlider`: ordering clamp, `--pc-slider-pct` sync, `petal:slider-change` detail payload for single and dual

## Non-goals

- No removal of `type="range"` / `type="range-dual"` from `field.ex`/`input.ex` in this PR — deprecation notes only
- No multi-thumb beyond two
- No logarithmic/non-linear scales
- No RTL-specific work beyond what native inputs give us
- No colour variants — the slider is primary-only, like the existing range fill

## Process

1. Build on a feature branch: `lib/petal_components/slider.ex` + a `pc-slider` section in `assets/default.css` + the `PetalSlider` hook in `assets/js/petal_components.js` (registered in the hooks export) + tests + playground page (nav entry, dials, scenarios). CSS/JS live in this repo — no external packages, no new hex deps.
2. Verify in the playground (`mix run dev.exs`, http://localhost:4000) — exercise every dial, light + dark, keyboard-only, and `prefers-reduced-motion`. Check webkit AND Firefox (range pseudo-elements differ; the existing CSS section shows the split).

## Definition of done

- `<.slider>` component with single and dual modes, marks, tooltip/inline value, vertical orientation, sizes, disabled
- `pc-slider` CSS section in `assets/default.css` (webkit + Firefox pseudo-elements, dark mode, focus-visible ring, reduced motion respected)
- `PetalSlider` hook with ordering invariant and `petal:slider-change` event
- Deprecation notes on the `range-dual` field docs pointing to `<.slider>`
- Tests green (`mix test` + `npm test`), every attr/variant asserted, a11y assertions included
- Playground page with all dials and the three scenarios
- PR screenshots light + dark
- Docs-ready attr docs: every attr has a `doc:` string; the moduledoc includes usage examples for single, dual, and marks

mplatts:
Closes #608

## Summary

A first-class `<.slider>`: single and dual thumb, marks, a tooltip or inline value readout, sizes, vertical orientation and disabled. It absorbs the display work that lived in `field.ex`/`input.ex` and adds what those types could not carry.

Five parts per CONTRIBUTING.md:

- `lib/petal_components/slider.ex` (exported from `lib/petal_components.ex`)
- a `pc-slider` section in `assets/default.css`, appended as its own `@layer components` block, nothing reflowed
- `test/petal/slider_test.exs` (32 tests) and `test/js/slider.test.js` (18 tests)
- `lib/petal_components/showcase/slider.ex` + its `registry.ex` entry
- the `slider` playground page in `dev.exs`, rebuilt around `<.slider>`

Plus `### Unreleased` in `CHANGELOG.md` and deprecation notes on the `range` / `range-dual` docs.

## Native `<input type="range">`, not `role="slider"`

I went native, as the issue proposed, and the keyboard argument decided it on its own. The WAI-ARIA slider pattern's whole keyboard map — arrows, Home, End, PageUp, PageDown — plus `role="slider"`, `aria-valuemin`/`valuemax`/`valuenow`, touch and pointer handling, and form posting, all arrive for free and stay correct across browsers and assistive tech. A custom `role="slider"` implementation would mean re-implementing every one of those and getting them all right, in exchange for styling freedom I did not actually need: the track, fill, thumb, marks and tooltip are drawn as **real sibling elements** under the input, with the input's own track and thumb painted out. That sidesteps the usual native-styling ceiling, and it fixes a real bug class along the way — `dark:` and `--pc-radius` resolve on real elements but cannot on `::-webkit-slider-runnable-track` (a `dark:` variant compiles to a trailing `:where(.dark, …)` ancestor test a pseudo-element can never satisfy, which is why `.pc-range-input`'s track used to stay light in dark mode).

**On dual and native:** correct, one native input cannot carry two thumbs. Dual mode is two overlaid native inputs, each posting its own name — the technique the existing `pc-dual-range` already uses. That keeps both thumbs individually focusable and keyboard-operable, which a single custom two-thumb widget makes materially harder. It did not push me toward a custom implementation; it is the one place a hook is genuinely required (see below).

I verified the keyboard map by hand in the playground rather than assuming it: on the volume slider, 60 → 62 (two ArrowRights), → 72 (PageUp), → 0 (Home), → 100 (End).

## CSS-first, and the one thing the hook is for

Geometry rides `--pc-slider-pct` (and `-min`/`-max`) custom properties that the **server renders inline**, so the control paints correctly before any JS connects and with JS off entirely. The fill, both tooltips and every tick anchor off those same properties, so positioning is pure CSS with nothing measured.

`PetalSlider` does only what the client alone can know:

1. keeping the percentages live while you drag
2. the ordering invariant two overlaid inputs cannot enforce themselves, including the z-index lift when the thumbs meet (without it, one thumb becomes unreachable once they overlap)
3. emitting a bubbling `petal:slider-change` CustomEvent for apps that want the value without a form

`JS.dispatch` and CSS cannot do (1) or (2) — there is no declarative hook for "the user is dragging a native range thumb". Everything else is CSS.

Three additions, no removals. Everything in the sketch is present with the sketched names and defaults.

1. **`label` added.** The sketch has no attr that can name the inputs, but an input must never be nameless. `label` is the accessible-name base *and* a visible label above the track; it defaults to the humanised field name. Dual thumbs become `"<label> minimum"` / `"<label> maximum"`, which is the naming the issue's a11y section asks for.
2. **`min_name` / `max_name` added.** The test checklist asks for dual names "from `values` + explicit names", but a single `name` cannot express two. `values` + `name` derives `"<name>_min"` / `"<name>_max"`; these two override it.
3. **`rest` applies to the wrapper, not the inputs.** The issue suggests overriding the dual thumbs' `aria-label` via `rest`, but `rest` is one bag of attributes and dual mode needs two *different* labels, so it cannot express that. `label` is the override instead. Wrapper is also where `phx-hook` and the geometry properties live, so it is the natural target.

Also: errors from the form field(s) render through `PetalComponents.Field.field_error` (merged and de-duplicated across both thumbs, matching how `field.ex` merges min/max errors), so `<.slider>` is usable directly in a form without wrapping it in `<.field>`.

**Not done from the issue:** the vertical orientation is verified in Chrome only (screenshot below). The issue asks for a VoiceOver pass, which I could not run; `writing-mode: vertical-lr` + `direction: rtl` is the standards-track approach with `appearance: slider-vertical` as the legacy WebKit fallback, and the input is unchanged otherwise, so the announcement should be identical — but that is reasoning, not a test. Firefox was likewise not exercised; the CSS carries both the `::-webkit-*` and `::-moz-*` branches, following the existing split.

## The pre-existing slider situation

Worth reporting, since it was ambiguous going in:

- **There was no `Slider` module and no slider commit history.** `git log --all -- '*slider*'` is empty, and `lib/petal_components.ex` had `slider` free, as the issue said.
- **The playground nav entry was not stale.** `main` already had a full, working `slider` page — but it demoed `<.field type="range">` and `type="range-dual"`, pulling its examples from the *Field* showcase (`~w(sliders slider_dual)a`). So the slug was real and occupied, not a placeholder.
- **Given the strict one-slug-per-component rule, I replaced that page rather than adding a second one.** The new page keeps the `slider` slug and is built entirely on `<.slider>`, with the six dials the issue specifies (mode, marks, show_value, orientation, step, size — plus disabled) and the three scenarios (price-range filter, volume, year picker), a keyboard hint line, and a closing note pointing at the deprecated field types. Verified it renders and does not fall back to the Button page.
- **`origin/dual-range-slider` is stale and superseded.** Its 10 commits are not reachable from `main`, but the feature they built *is* on `main` (squash-merged), so the branch is history, not pending work. I did not build on it; the issue's brief supersedes it.
- The Field showcase keeps its `sliders` / `slider_dual` examples so the deprecated types stay documented on petal.build; both descriptions now say they are superseded by `<.slider>`.

## Verification

| Gate | Result |
|---|---|
| `mix credo` | **zero new entries** (44 issues before and after, diffed line-for-line vs `origin/main`) |
| `mix test` | 913 → **945**, 0 failures, 1 skipped (+32) |
| `npm test` | 157 → **175** (+18) |

No test was weakened or deleted. No new dependencies — no npm package, no hex dep.

Exercised in the playground on :4023: every dial, light and dark, dual, marks, vertical, tooltip-on-hover, and keyboard-only.

## Screenshots

Light and dark of the playground page, thumbs at non-default positions so the fill/track relationship is visible.

Also captured while verifying (happy to attach if useful): dual + vertical, marks with the tick inversion inside the fill, and the tooltip bubble anchored to the thumb.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,39 @@
 # Changelog
+
+### Unreleased
+
+#### Added
+
+- **New `slider` component: a range slider with one thumb or two, marks, a
+  value readout, sizes and vertical orientation.** It is a native
+  `<input type="range">` underneath, so the arrow / Home / End / PageUp /
+  PageDown keyboard map, the `role="slider"` and `aria-valuemin`/`valuemax`/
+  `valuenow` announcement, touch handling and form posting all come from the
+  browser rather than being re-implemented. What the library adds is paint: a
+  track and fill that ride the gray and primary ramps in both schemes, tick
+  marks that invert once the fill swallows them, and a value bubble or inline
+  readout with a prefix and suffix.
+
+  Dual mode is two overlaid inputs posting two form fields, so a price filter
+  is a plain `phx-change` with no JavaScript in your app:
+
+  ```heex
+  <.slider min_field={@form[:min]} max_field={@form[:max]} min={0} max={1000} step={50} label="Price" value_prefix="$" show_value="inline" />
+  ```
+
+  Geometry rides CSS custom properties that the server renders inline, so the
+  control paints correctly before the `PetalSlider` hook connects and with JS
+  off entirely. The hook only does what the client alone can know: keeping
+  those percentages live while you drag, enforcing the thumb ordering two
+  native inputs cannot enforce themselves, and emitting a bubbling
+  `petal:slider-change` event for apps that want the value without a form.
+
+#### Deprecated
+
+- `<.field type="range">` and `<.field type="range-dual">` are superseded by
+  `<.slider>`. Both still work and nothing is being removed in this release -
+  the docs now point at the slider for new code.
+
 ### 4.14.0 - 2026-08-11
 
 #### Added
--- a/assets/default.css
+++ b/assets/default.css
@@ -7586,3 +7586,294 @@
       transition-duration: 1ms;
     }
   }
+
+@layer components {
+  /* ---- Slider ---------------------------------------------------------- */
+  /* A native <input type="range"> doing all the work (keyboard, ARIA, form
+     posting) with its own track and thumb painted out, and our track, fill,
+     marks and tooltip drawn as real elements underneath it. Real elements are
+     the point: `dark:` and `--pc-radius` resolve on them, which they cannot on
+     ::-webkit-slider-runnable-track (a `dark:` variant compiles to a trailing
+     :where(.dark, ...) ancestor test that a pseudo-element can never satisfy -
+     the bug that used to leave .pc-range-input's track light in dark mode).
+
+     Geometry rides three custom properties so nothing needs measuring in JS:
+       --pc-slider-track  track thickness
+       --pc-slider-thumb  thumb diameter
+       --pc-slider-pct    thumb position (single), or -min / -max (dual)
+     The server renders the initial percentages inline, so the control paints
+     correctly with JS off; the PetalSlider hook keeps them live while dragging. */
+  .pc-slider {
+    @apply relative select-none;
+    --pc-slider-track: 0.375rem;
+    --pc-slider-thumb: 1.25rem;
+    --pc-slider-length: 11rem;
+    --pc-slider-surface: var(--color-gray-200);
+  }
+
+  .pc-slider:where(.dark, .dark *) {
+    --pc-slider-surface: color-mix(in oklab, var(--color-gray-400) 17%, transparent);
+  }
+
+  .pc-slider--sm {
+    --pc-slider-track: 0.25rem;
+    --pc-slider-thumb: 1rem;
+  }
+
+  .pc-slider--lg {
+    --pc-slider-track: 0.5rem;
+    --pc-slider-thumb: 1.5rem;
+  }
+
+  .pc-slider--disabled {
+    @apply opacity-50 pointer-events-none;
+  }
+
+  /* Label row: name on the left, inline readout on the right. */
+  .pc-slider__header {
+    @apply flex items-baseline justify-between gap-3 mb-1.5;
+  }
+
+  .pc-slider__label {
+    @apply mb-0;
+  }
+
+  .pc-slider__value {
+    @apply text-sm font-medium text-gray-700 tabular-nums dark:text-gray-200;
+  }
+
+  /* Stacking context for track, fill, marks and the input(s). */
+  .pc-slider__track-wrapper {
+    @apply relative flex items-center;
+    min-height: var(--pc-slider-thumb);
+  }
+
+  .pc-slider__track {
+    @apply absolute inset-x-0 rounded-full;
+    height: var(--pc-slider-track);
+    background-color: var(--pc-slider-surface);
+  }
+
+  /* Single fills from the start to the thumb; dual spans between the thumbs. */
+  .pc-slider__fill {
+    @apply absolute rounded-full bg-primary-500;
+    height: var(--pc-slider-track);
+    left: 0;
+    right: calc(100% - var(--pc-slider-pct, 0%));
+  }
+
+  .pc-slider--dual .pc-slider__fill {
+    left: var(--pc-slider-pct-min, 0%);
+    right: calc(100% - var(--pc-slider-pct-max, 100%));
+  }
+
+  /* Ticks sit above the fill so a mark inside the filled span can invert. */
+  .pc-slider__marks {
+    @apply absolute inset-x-0 pointer-events-none;
+    height: var(--pc-slider-track);
+  }
+
+  .pc-slider__mark {
+    @apply absolute top-1/2 rounded-full;
+    left: var(--pc-slider-mark, 0%);
+    width: 0.25rem;
+    height: 0.25rem;
+    transform: translate(-50%, -50%);
+    background-color: color-mix(in oklab, var(--color-gray-500) 55%, transparent);
+  }
+
+  /* On-primary treatment once the fill has swallowed the tick. */
+  .pc-slider__mark--filled {
+    background-color: color-mix(in oklab, white 70%, transparent);
+  }
+
+  .pc-slider__mark-labels {
+    @apply relative h-5 mt-1.5;
+  }
+
+  .pc-slider__mark-label {
+    @apply absolute text-xs text-gray-500 whitespace-nowrap dark:text-gray-400 tabular-nums;
+    left: var(--pc-slider-mark, 0%);
+    transform: translateX(-50%);
+  }
+
+  /* The native input: its own track and thumb are painted out, everything
+     visible comes from the elements above. */
+  .pc-slider__input {
+    @apply w-full bg-transparent appearance-none cursor-pointer focus:outline-hidden disabled:cursor-not-allowed;
+    height: var(--pc-slider-thumb);
+    margin: 0;
+  }
+
+  .pc-slider__input::-webkit-slider-runnable-track {
+    @apply bg-transparent;
+    height: var(--pc-slider-track);
+  }
+
+  .pc-slider__input::-moz-range-track {
+    @apply bg-transparent;
+    height: var(--pc-slider-track);
+  }
+
+  /* box-border is load-bearing: preflight's border-box reset cannot reach a
+     pseudo-element (the universal selector does not match one) and the UA
+     defaults disagree - WebKit sizes range thumbs border-box, Firefox
+     content-box - so without it Firefox draws the ring outside the diameter. */
+  .pc-slider__input::-webkit-slider-thumb {
+    @apply relative appearance-none rounded-full cursor-pointer bg-primary-500 border-[3px] border-white box-border shadow shadow-black/25 transition-transform duration-100 ease-out dark:border-gray-900;
+    width: var(--pc-slider-thumb);
+    height: var(--pc-slider-thumb);
+    margin-top: calc((var(--pc-slider-track) - var(--pc-slider-thumb)) / 2);
+  }
+
+  .pc-slider__input::-moz-range-thumb {
+    @apply rounded-full cursor-pointer bg-primary-500 border-[3px] border-white box-border shadow shadow-black/25 transition-transform duration-100 ease-out dark:border-gray-900;
+    width: var(--pc-slider-thumb);
+    height: var(--pc-slider-thumb);
+  }
+
+  .pc-slider__input:hover::-webkit-slider-thumb {
+    @apply scale-110;
+  }
+
+  /* focus-visible only - a persistent :focus ring would sit on the thumb for
+     the whole of a mouse drag. */
+  .pc-slider__input:focus-visible::-webkit-slider-thumb {
+    @apply scale-110 ring-2 ring-primary-500/50;
+  }
+
+  .pc-slider__input:hover::-moz-range-thumb {
+    transform: scale(1.1);
+  }
+
+  .pc-slider__input:focus-visible::-moz-range-thumb {
+    transform: scale(1.1);
+    outline: none;
+    box-shadow: 0 0 0 2px color-mix(in oklab, var(--color-primary-500) 50%, transparent);
+  }
+
+  /* Dual: two zero-height inputs stacked over the shared track. Only the
+     thumbs take pointer events, so each one stays independently draggable
+     where they overlap. Zero height plus `top: auto` means the flex parent's
+     align-items: center resolves their static position for us. */
+  .pc-slider--dual .pc-slider__input {
+    @apply absolute w-full outline-none pointer-events-none;
+    height: 0;
+  }
+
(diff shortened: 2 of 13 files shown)


mplatts:
## Screenshots

**Light / dark**, thumbs at non-default positions

Slider, light
Slider, dark

**Dials** — dual + marks + tooltip, vertical, marks + tooltip

Dual, marks, tooltip
Vertical
Marks and tooltip

**Keyboard** — End key driving the value to 100

Keyboard End

### Verified independently

| Check | Result |
|---|---|
| `mix test` | 945 tests, 0 failures, 1 skipped (+32) |
| `npm test` | 175 passing (+18) |
| `mix credo` | zero new entries vs `main` (the `field.ex` complexity finding just shifts line 249 → 257) |
| new dependencies | none |
| slugs : clauses | 60 : 60 |

## The slug situation is worth reading

The `slider` nav entry on `main` was **not** a placeholder — it was a full working page demoing `&lt;.field type="range"&gt;` / `"range-dual"`, pulling from the *Field* showcase. Given the strict one-slug-per-component rule, this PR **replaces** that page rather than adding a second, which is why the count stays at 60 rather than going to 61. Separately, `origin/dual-range-slider` turned out to be history not pending work: its commits aren't reachable from `main`, but the feature they built is (squash-merged).

**The `field.ex` / `input.ex` changes are documentation only** — I checked specifically. No runtime warning, no raise, no behaviour change; `type="range"` keeps working. Just an info callout pointing new code at `&lt;.slider&gt;`. Whether to actually deprecate is your call, not something this PR forces.

Going native `&lt;input type="range"&gt;` with the track/fill/thumb drawn as real sibling elements also dodges a live bug class: `dark:` can't resolve on `::-webkit-slider-runnable-track`, which is why `.pc-range-input`'s track used to stay light in dark mode.

Keyboard verified by hand: 60 → 62 (ArrowRight ×2) → 72 (PageUp) → 0 (Home) → 100 (End).

Not verified: Firefox (CSS carries both `::-webkit-*` and `::-moz-*` branches), a VoiceOver pass on vertical orientation, and `prefers-reduced-motion` with the media feature actually forced on.

PR #637, opened 2026-08-12
feat(slide_over): bottom-sheet drawer mode for origin="bottom"

### LINKED ISSUE #614: Enhancement: SlideOver — bottom-sheet drawer mode (origin="bottom" upgrade)
## Summary

`<.slide_over>` already accepts `origin="bottom"`, but today that just slides a full-width panel up from the bottom edge with the same chrome as the side sheets. This issue upgrades `origin="bottom"` into a first-class mobile drawer: a grab-handle pill, drag-to-dismiss with velocity-aware physics, optional snap points, rounded top corners, and safe-area padding. The behaviour reference is vaul (the drawer primitive behind shadcn's Drawer); reui's Drawer shows the scope of examples worth matching. The keyboard and assistive-tech path is untouched — the drag layer is a pointer-only enhancement on top of the existing dialog semantics.

## References

- reui Drawer — https://reui.io/components/drawer (5 examples; study for scope and demo ideas)
- shadcn/ui Drawer — https://ui.shadcn.com/docs/components/drawer
- vaul — https://vaul.emilkowal.ski / https://github.com/emilkowalski/vaul (the behaviour reference: drag-to-dismiss threshold + velocity, snap points, spring-back, optional background scaling)

We build a native HEEx/LiveView equivalent from scratch. Reference their behaviour and variants, never their code.

## Where to start in this repo

- `lib/petal_components/slide_over.ex` — the existing component. Note `get_transition_classes/1`, `get_classes/3`, and how `show_slide_over/3` / `hide_slide_over/3` drive `Phoenix.LiveView.JS` transitions on `#{id}-overlay` and `#{id}-content`. The drawer mode extends this module; it is not a new component.
- `assets/default.css` — the `pc-slideover__*` section (around line 3986). New drawer classes (`pc-slideover__handle`, bottom-drawer box modifiers) join this section.
- `assets/js/petal_components.js` — all hooks live in this single file. Study `PetalInputOTP` and `PetalPopover` as the precedent for "hook only when CSS + LiveView.JS genuinely cannot do it" — pointer-driven drag physics qualifies. The new hook is `PetalDrawer`.
- `test/petal/slide_over_test.exs` — existing test conventions (ComponentCase + LazyHTML).
- `test/js/` — vitest suites (`popover.test.js`, `toast_pause.test.js` show the style). `npm test` runs them.

## API sketch

All additions go on the existing `slide_over/1`. Proposed attrs (builder may refine, flagging deviations in the PR):

```elixir
attr(:handle, :boolean,
  default: nil,
  doc: "show the grab-handle pill. Defaults to true when origin=\"bottom\", false otherwise. Set explicitly to override."
)

attr(:drag_to_dismiss, :boolean,
  default: true,
  doc: "for origin=\"bottom\": drag the sheet down past a threshold (or flick with enough velocity) to dismiss. Pointer-only; Escape and the close button always work regardless. Ignored for other origins."
)

attr(:snap_points, :list,
  default: nil,
  doc: "optional list of viewport-height fractions the drawer can rest at, e.g. [0.4, 0.9]. Drag between them; a fast flick skips to the next point in the flick direction. nil means content-height with a single rest position."
)

attr(:initial_snap, :float,
  default: nil,
  doc: "which snap point the drawer opens at. Must be a member of snap_points. Defaults to the first entry."
)

attr(:scale_background, :boolean,
  default: false,
  doc: "vaul-style effect: scales and rounds the page behind the drawer while it is open. Off by default because it transforms the whole page (see trade-offs in the moduledoc). Requires the consumer to mark the page wrapper with data-pc-drawer-wrapper."
)
```

Behavioural notes for the builder:

- `handle: nil` resolving per-origin keeps the API zero-config for the common case: `<.slide_over origin="bottom">` gets the pill for free, side sheets stay as they are.
- The Elixir side renders configuration only. Snap/drag config is emitted as data attributes on `#{id}-content` (e.g. `data-snap-points="0.4,0.9"`, `data-initial-snap="0.4"`, `data-drag-dismiss`), and the `PetalDrawer` hook (`phx-hook="PetalDrawer"`, attached only when `origin == "bottom"` and `drag_to_dismiss` or `snap_points` is set) reads them. No hook attaches for plain non-draggable bottom sheets — those stay CSS + LiveView.JS only.
- Dismiss must route through the existing `hide_slide_over/3` path (including `close_slide_over_target`), so the server-side `"close_slide_over"` event contract is identical whether the user drags, taps the overlay, presses Escape, or clicks the close button. The hook can trigger this via `JS.exec` on the existing `phx-window-keydown` binding's command or a dedicated `data-hide` attribute holding the composed JS — follow the pattern LiveView recommends for hook-triggered JS commands.

### The `PetalDrawer` hook

Single-finger pointer events on the sheet (pointerdown on the handle or sheet body, pointermove, pointerup):

- Translate follows the finger 1:1 while dragging downward; upward drag past the top snap gets heavy resistance (rubber-band, e.g. delta * 0.15) so the sheet feels anchored.
- On release: dismiss when the sheet has travelled past a distance threshold (a fraction of sheet height, vaul uses ~25%) OR the release velocity exceeds a flick threshold (px/ms, computed from the last few pointer samples). Otherwise animate back to the nearest snap point.
- With `snap_points`, release settles on the nearest point by position, biased by velocity direction; only a downward release below the lowest point dismisses.
- Springs are CSS transitions applied on release (transform transition toggled on for the settle, off during active drag). Under `prefers-reduced-motion`, no spring: snap instantly to the settle position or dismiss with the existing reduced-motion behaviour.
- Text selection and vertical page scroll must not fight the drag: `touch-action: none` on the handle; on the sheet body, only capture the drag when the sheet's scrollable content is at scroll-top (the vaul rule), so inner scrolling keeps working.
- Keep the physics maths (threshold check, velocity estimate, nearest-snap resolution) in small pure functions exported for testing, as the other hooks in `assets/js/petal_components.js` do.

## Variants & states to build

- [ ] `origin="bottom"` default: full-width sheet, rounded top corners riding the large-panel radius rule (`min(var(--pc-radius) * 1.2, 20px)` on the top corners only), border-top, handle pill shown, `env(safe-area-inset-bottom)` padding on the content/footer.
- [ ] Handle pill: centred, muted gray, comfortable hit target (min 44px tall touch area even though the pill is visually ~4px x 40px). `aria-hidden="true"` — it is decorative; dragging is the interaction, not the element.
- [ ] Drag-to-dismiss on (default) and off (`drag_to_dismiss={false}`).
- [ ] No snap points (content height) vs `snap_points={[0.4, 0.9]}` with `initial_snap`.
- [ ] Overlay behaviour with snaps: overlay stays fully shown at every snap (do not fade it proportionally — keep it simple).
- [ ] `scale_background={true}` demo path: page wrapper scales to ~0.96, gains radius, over a dark backdrop. Document in the moduledoc that this transforms the whole page (breaks `position: fixed` descendants, costs a repaint) and that off is the default for that reason.
- [ ] Dark mode for all of the above (existing `pc-slideover` surface rules).
- [ ] `prefers-reduced-motion`: open/close use the existing reduced-motion handling; drag settles instantly, no spring.
- [ ] Side/top sheets (`origin` left/right/top) completely unchanged — no handle, no hook, no drag.

## Playground page

Extend the existing slide_over playground page (or add a "Drawer" section to it) in `dev.exs`:

- Dials: origin (to show bottom vs side contrast), handle on/off, drag_to_dismiss on/off, snap points off / [0.4, 0.9], scale_background on/off.
- Example scenarios:
  1. Mobile filter sheet — a product-list filter drawer with checkboxes and an "Apply" footer, content-height, drag to dismiss.
  2. Music-player queue sheet — `snap_points={[0.4, 0.9]}` with `initial_snap={0.4}`: peek shows now-playing + next track, drag up to 0.9 for the full queue.
  3. Share/action sheet — simple list of actions, the classic iOS-style use.
- Review the page with the browser resized to the mobile preset — this is a mobile-first pattern and the demos should be judged in a phone-width viewport. Also exercise touch emulation for the drag.

## Accessibility

- The dialog semantics are already correct and must not change: `role="dialog"`, `aria-modal="true"`, labelled by title, `focus_first` on open, Escape closes (`close_on_escape`), click-away closes, body scroll locked.
- The drag layer is a pointer-only progressive enhancement. Keyboard and screen-reader users get exactly today's behaviour: Tab cycles the panel, Escape or the close button dismisses. No new keyboard bindings, no ARIA on the drag mechanics.
- Handle pill is `aria-hidden="true"`.
- Snap-point changes are visual only; do not announce them.
- `prefers-reduced-motion` honoured as described above.

## Tests

ExUnit (`test/petal/slide_over_test.exs`):
- [ ] `origin="bottom"` renders the handle by default; `handle={false}` removes it; `handle={true}` on `origin="right"` renders it (explicit override wins).
- [ ] Handle element has `aria-hidden="true"` and the `pc-slideover__handle` class.
- [ ] Bottom origin emits the rounded-top/safe-area drawer classes; side origins do not.
- [ ] `snap_points={[0.4, 0.9]}` + `initial_snap={0.4}` emit the expected `data-*` config attributes and attach `phx-hook="PetalDrawer"`.
- [ ] `drag_to_dismiss={false}` with no snap points attaches no hook.
- [ ] Side origins never attach the hook regardless of attrs.
- [ ] Existing assertions (dialog role, aria wiring, close targets, footer slot) still pass untouched.

Vitest (`test/js/drawer.test.js`):
- [ ] Threshold maths: release below distance threshold with low velocity → spring back; past threshold → dismiss.
- [ ] Velocity: a short flick under the distance threshold but over the velocity threshold → dismiss.
- [ ] Nearest-snap resolution with and without velocity bias; downward release below lowest snap → dismiss.
- [ ] Rubber-band resistance above the top snap.
- [ ] Test the exported pure functions in isolation — no DOM gymnastics needed for the maths.

## Non-goals

- Horizontal drag-to-dismiss on left/right side sheets. Bottom only.
- Nested drawers.
- Non-modal / persistent drawer mode.
- Proportional overlay fade tied to drag position.
- Any npm dependency — the hook is hand-rolled in `assets/js/petal_components.js` like every other hook in the library.

## Process

1. Build on a feature branch: changes to `lib/petal_components/slide_over.ex`, the `pc-slideover` section of `assets/default.css`, the `PetalDrawer` hook in `assets/js/petal_components.js`, tests, and the playground page updates (dials + examples). CSS/JS live in this repo — no external packages.
2. Verify in the playground (`mix run dev.exs`, http://localhost:4000) — exercise every dial, light + dark, keyboard-only, `prefers-reduced-motion`, and a mobile-width viewport with touch emulation for the drag.
4. Open a PR against `main` with screenshots (light + dark, mobile viewport) of the playground page. The automated review must pass.

## Definition of done

- `slide_over/1` gains `handle`, `drag_to_dismiss`, `snap_points`, `initial_snap`, `scale_background` — every attr has a `doc:` string, and the moduledoc gains a drawer usage example plus the scale_background trade-off note.
- `PetalDrawer` hook with exported pure physics functions; drag, flick, snap, and spring-back working on a real touch device or emulation.
- CSS additions in the `pc-slideover` section: handle pill, rounded-top drawer box, safe-area padding, optional background-scale rules.
- ExUnit and vitest suites green; new assertions per the Tests section.
- Playground drawer section with the three example scenarios and all dials.
- PR screenshots (light + dark, mobile viewport); automated review passing; side-sheet behaviour visibly unchanged.

Milestone: 4.17.0

mplatts:
Closes #614

## Summary

`origin="bottom"` already existed, but it slid a plain full-width panel up with the same chrome as a side sheet. This turns it into a proper mobile drawer.

- Rounded top corners on the large-panel radius rule (`min(calc(var(--pc-radius) * 1.2), 20px)`), top corners only since the bottom two are off-screen
- Grab-handle pill, `aria-hidden` because it is decorative
- `env(safe-area-inset-bottom)` padding on the body and footer
- `max-height: 92dvh` with the body scrolling inside it
- Drag-to-dismiss: distance threshold (25% of sheet height) **or** release velocity, with rubber-band resistance above the top snap
- Optional `snap_points` as viewport-height fractions, opening at `initial_snap`. A flick skips to the next point in its direction; only a downward release below the lowest point dismisses
- Opt-in `scale_background` for the vaul-style page shrink

Five new attrs, all on the existing `slide_over/1`: `handle`, `drag_to_dismiss`, `snap_points`, `initial_snap`, `scale_background`. `handle` defaults to `nil`, which resolves to on for bottom sheets and off elsewhere, so a bare `<.slide_over origin="bottom">` gets the pill with no configuration.

Dismissal always routes through `hide_slide_over/3` via a `data-pc-drawer-hide` attribute the hook `execJS`'s, so the server sees one `"close_slide_over"` event with the same `close_slide_over_target` whether the user drags, taps away, presses escape or clicks the close button.

Accessibility is unchanged. The drag is a pointer-only enhancement layered over the existing dialog: `role="dialog"`, `aria-modal`, `focus_first` on open, escape and click-away all behave exactly as before. No new keyboard bindings, no ARIA on the drag mechanics, snap changes not announced, `prefers-reduced-motion` settles instantly instead of springing.

## Backward compatibility

**No existing attr default changed and no existing test was modified.** The diff on `test/petal/slide_over_test.exs` is additions only — `git diff` shows zero deleted lines in that file, which is the compatibility proof. All 913 pre-existing tests pass untouched.

Left, right and top sheets emit byte-identical markup: no handle, no hook, no data attributes, no new classes. A test asserts this explicitly for all three origins even when every drawer attr is passed. Verified visually too — the side sheet renders exactly as it did on `main`.

The only changed lines outside additions are three, all additive in effect:
- `class={get_classes(...)}` became `class={[get_classes(...), @drawer? && "pc-slideover__box--drawer"]}` — the modifier is `false` (dropped) for every non-bottom origin
- the playground page root gained `data-pc-drawer-wrapper` (for the `scale_background` demo)
- the playground's showcase list gained the three new example ids

`assets/default.css` is appended to in a fresh `@layer components { }` block; nothing existing was reflowed.

## Did you ship a hook, and why

Yes — `PetalDrawer`. Pointer-driven drag physics is the one thing CSS and `Phoenix.LiveView.JS` genuinely cannot do: following a finger 1:1 and then deciding, from distance **and** release velocity, whether the sheet springs back or closes. The specific failure without it is that there is no way to express "translate follows pointerY, and on pointerup dismiss if travelled > 25% of height OR |velocity| > 0.5px/ms" declaratively.

Kept minimal:
- The physics is three exported pure functions — `drawerResistance`, `drawerVelocity`, `drawerSettle` — unit-tested with no DOM.
- The Elixir side renders configuration only, as `data-*` attributes.
- **The hook attaches only when a bottom sheet is actually draggable, snapped or scaling its background.** A plain `drag_to_dismiss={false}` bottom sheet stays CSS + `LiveView.JS` only, and side origins never get it. There is a test for each of those.
- No new dependencies, npm or hex. The hook is hand-rolled in `assets/js/petal_components.js` like every other one.

## Deviations from the issue's sketch

1. **Handle hit target.** The issue asks for a min 44px tall touch area on the pill. A 44px block above the header pushed the title down noticeably, and faking it with a `::before` overlay would have covered the close button. Instead the handle is ~28px visually and **the hook captures pointerdown across the whole sheet head**, not just the pill — so the real drag target is the full sheet width by everything above the scrolled content, far larger than 44px. Happy to change this if you'd rather have the literal 44px pill.

2. **`initial_snap` validation.** The issue says it "must be a member of `snap_points`". Rather than raise, a value outside the list falls back to the lowest point — an `initial_snap` you can never drag back to would be a worse failure mode than a silent correction. There is a test for it. Say the word if you'd prefer it to raise.

3. **`snap_points` normalisation.** They are sorted ascending on render regardless of the order passed, so `[0.9, 0.4]` and `[0.4, 0.9]` behave identically.

4. **`scale_background` needs the hook.** It is not pointer-driven, but it has to follow the panel opening and closing, and the open/close is driven by JS commands that write inline `display` on the content element. The hook watches that with a `MutationObserver` and toggles `.pc-drawer-scaled` on `[data-pc-drawer-wrapper]`. I could not find a CSS-only way to reach a sibling page wrapper from the sheet's open state; if there's a house pattern for this I've missed, I'll switch to it.

5. **Overlay does not fade with drag position** — matching the issue's "keep it simple" instruction.

## Tests

| Suite | Before | After |
|---|---|---|
| `mix test` | 913, 0 failures, 1 skipped | **930, 0 failures, 1 skipped** |
| `npm test` | 157 | **191** |

- 17 new ExUnit tests covering handle resolution per origin, the `aria-hidden` handle, drawer-only box classes, hook attachment and non-attachment, the emitted `data-*` config, snap sorting and `initial_snap` fallback, `close_slide_over_target` in the drag command, and that dialog semantics are untouched in drawer mode.
- 34 new vitest specs in `test/js/drawer.test.js`: the threshold and velocity maths, nearest-snap resolution with and without velocity bias, rubber-band resistance, plus hook-level specs for initial snap offset, spring-back, dismiss routing through the shared hide command, ignoring taps on controls, the scroll-top capture rule, `scale_background` toggling, and reduced-motion settling.

`mix format --check-formatted`, `mix compile --force --warnings-as-errors` clean. **Credo: one entry fewer than `main`, zero new** — `SlideOver` picked up a moduledoc, which closed the existing `Modules should have a @moduledoc tag` finding.

## Verified by hand in the playground

- **Escape closes** the drawer.
- **Focus management**: pressing Enter immediately after opening activates the close button, i.e. `focus_first` still lands inside the panel and the handle does not steal focus.
- **The hook runs**: the snapped queue drawer opens at `translate3d(0px, 422px, 0px)` on a `90dvh` sheet in an 844px viewport, which is exactly the `initial_snap={0.4}` peek.
- **Side sheets visibly unchanged.**

**What I could not verify:** an end-to-end pointer drag in the browser. The automation tool's coordinate resolution mishandles `position: fixed` elements and its `eval` was unavailable, so every scripted drag landed on the page header instead of the sheet. The drag path is covered at the unit level (including dismiss routing through `data-pc-drawer-hide` with a stubbed `liveSocket`), but **it has not been exercised on a real touch device or with real pointer events** — worth a manual pass on a phone before merge.

## Screenshots

Light and dark playground pages, the drawer open at desktop and phone width, and the snapped queue drawer peeking at 0.4 — attached separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,38 @@
 # Changelog
+### Unreleased
+
+#### Added
+
+- **`slide_over` gains a bottom-sheet drawer mode.** `origin="bottom"`
+  already existed, but it slid a plain full-width panel up with the same
+  chrome as a side sheet. It is now a proper mobile drawer: rounded top
+  corners on the large-panel radius rule, a grab-handle pill, a border,
+  `env(safe-area-inset-bottom)` padding under the body and footer, and
+  drag-to-dismiss. Pull the sheet down past about a quarter of its height,
+  or flick it, and it closes - through the same `hide_slide_over` path and
+  the same single `"close_slide_over"` event as escape, click-away and the
+  close button, `close_slide_over_target` included.
+
+  Five new attrs, all on the existing `slide_over/1`: `handle` (defaults to
+  `nil`, which resolves to on for bottom sheets and off everywhere else, so
+  `<.slide_over origin="bottom">` gets the pill for free), `drag_to_dismiss`,
+  `snap_points` (viewport-height fractions the drawer can rest at, e.g.
+  `[0.4, 0.9]`), `initial_snap`, and `scale_background` (off by default -
+  it transforms the page wrapper, which re-parents `position: fixed`
+  descendants; see the moduledoc).
+
+  The drag is a pointer-only enhancement layered on the dialog. Keyboard and
+  screen-reader users get exactly what they had: `role="dialog"`,
+  `aria-modal`, focus into the panel, escape to close. The handle is
+  `aria-hidden` because it is decorative, snap changes are not announced, and
+  under `prefers-reduced-motion` the sheet settles instantly instead of
+  springing. The new `PetalDrawer` hook only attaches to bottom sheets that
+  are actually draggable, snapped or scaling their background - a plain
+  bottom sheet stays CSS and `LiveView.JS` only.
+
+  Left, right and top sheets are unchanged: no handle, no hook, no new
+  markup, and every existing attr default is the same.
+
 ### 4.14.0 - 2026-08-11
 
 #### Added
--- a/assets/default.css
+++ b/assets/default.css
@@ -7586,3 +7586,84 @@
       transition-duration: 1ms;
     }
   }
+
+@layer components {
+  /* Slideover - bottom-sheet drawer mode (origin="bottom").
+     Everything here is scoped to .pc-slideover__box--drawer, a class only the
+     bottom origin gets, so left/right/top sheets are untouched. */
+
+  /* The large-panel radius rule: token x 1.2, capped, and top corners only -
+     the sheet is anchored to the bottom edge, so its bottom corners sit
+     off-screen and rounding them would just expose a sliver of overlay. */
+  .pc-slideover__box--drawer {
+    border-top-left-radius: min(calc(var(--pc-radius, 0.625rem) * 1.2), 20px);
+    border-top-right-radius: min(calc(var(--pc-radius, 0.625rem) * 1.2), 20px);
+    max-height: 92dvh;
+    /* the sheet is the drag surface and its body is the scroller; letting the
+       sheet itself scroll would put a second scroller under the finger */
+    overflow: hidden;
+    will-change: transform;
+  }
+
+  /* The pill is decorative (aria-hidden) - the drag is the interaction, not
+     this element. The hook captures pointerdown across the whole sheet head,
+     so the real drag target is far larger than the 40x4 pill. touch-action
+     stops the browser claiming the gesture as a page scroll. */
+  .pc-slideover__handle {
+    @apply flex flex-none items-center justify-center pt-3 pb-2;
+    min-height: 1.75rem;
+    touch-action: none;
+    cursor: grab;
+  }
+
+  .pc-slideover__handle__pill {
+    @apply block w-10 rounded-full bg-gray-300 dark:bg-gray-600;
+    height: 0.25rem;
+  }
+
+  .pc-slideover__box--dragging {
+    @apply select-none;
+  }
+
+  .pc-slideover__box--dragging .pc-slideover__handle {
+    cursor: grabbing;
+  }
+
+  /* Home-indicator clearance on phones. env() resolves to 0 everywhere else,
+     so this costs nothing on desktop. */
+  .pc-slideover__box--drawer .pc-slideover__content {
+    padding-bottom: calc(1.25rem + env(safe-area-inset-bottom, 0px));
+  }
+
+  .pc-slideover__box--drawer .pc-slideover__footer {
+    padding-bottom: calc(1rem + env(safe-area-inset-bottom, 0px));
+  }
+
+  /* A drawer with a footer already clears the home indicator there; doubling
+     it on the body would leave a dead gap above the action row. */
+  .pc-slideover__box--drawer:has(.pc-slideover__footer) .pc-slideover__content {
+    padding-bottom: 1.25rem;
+  }
+
+  /* scale_background: the hook toggles .pc-drawer-scaled on the consumer's
+     [data-pc-drawer-wrapper] while the drawer is open. Opt-in only - see the
+     moduledoc for why a transformed ancestor re-parents fixed descendants. */
+  [data-pc-drawer-wrapper] {
+    transition:
+      transform 300ms cubic-bezier(0.32, 0.72, 0, 1),
+      border-radius 300ms cubic-bezier(0.32, 0.72, 0, 1);
+  }
+
+  .pc-drawer-scaled {
+    transform: scale(0.96);
+    transform-origin: top center;
+    border-radius: min(calc(var(--pc-radius, 0.625rem) * 1.2), 20px);
+    overflow: hidden;
+  }
+
+  @media (prefers-reduced-motion: reduce) {
+    [data-pc-drawer-wrapper] {
+      transition-duration: 1ms;
+    }
+  }
+}
--- a/assets/js/petal_components.js
+++ b/assets/js/petal_components.js
@@ -5424,6 +5424,283 @@ export const PetalDataTable = {
   },
 };
 
+// Bottom-sheet drag layer for <.slide_over origin="bottom">.
+//
+// Pointer physics is the one thing CSS and LiveView.JS genuinely cannot do:
+// following a finger 1:1 and then deciding, from distance AND release
+// velocity, whether the sheet springs back or closes. Everything else about
+// the drawer (radius, safe area, handle, open/close transitions) stays in CSS
+// and JS commands, and this hook is only attached when a bottom sheet is
+// actually draggable, snapped or scaling its background.
+//
+// The maths lives in pure functions below so it can be tested without a DOM.
+// Offsets are pixels DOWN from the drawer's tallest resting position: 0 is
+// fully open, larger is further off-screen.
+
+// Above the top snap the sheet should feel anchored rather than detached, so
+// overshoot is admitted at a fraction of its real size.
+export const drawerResistance = (offset, minOffset = 0, factor = 0.15) =>
+  offset < minOffset ? minOffset + (offset - minOffset) * factor : offset;
+
+// px/ms across the last few samples. A single frame pair is too noisy to tell
+// a flick from a slow drag that happened to end with one fast frame.
+export const drawerVelocity = (samples, sampleCount = 5) => {
+  if (!Array.isArray(samples) || samples.length < 2) return 0;
+  const recent = samples.slice(-sampleCount);
+  const first = recent[0];
+  const last = recent[recent.length - 1];
+  const elapsed = last.time - first.time;
+  if (elapsed <= 0) return 0;
+  return (last.y - first.y) / elapsed;
+};
+
+// Where a release lands: dismiss, or the snap offset to settle on.
+export const drawerSettle = ({
+  offset,
+  velocity = 0,
+  height = Infinity,
+  snapOffsets = [0],
+  dismissThreshold = 0.25,
+  velocityThreshold = 0.5,
+  dismissible = true,
+}) => {
+  const points = [...snapOffsets].sort((a, b) => a - b);
+  // largest offset is the LOWEST resting position - the one you fall off
+  const lowest = points[points.length - 1];
+  const travelled = offset - lowest;
+  const flickDown = velocity >= velocityThreshold;
+  const flickUp = velocity <= -velocityThreshold;
+
+  // Only a release below the lowest snap can dismiss. Between snaps a flick
+  // just moves to the next point, which is what keeps a snapped drawer from
+  // closing every time someone shoves it downward.
+  if (
+    dismissible &&
+    travelled > 0 &&
+    (flickDown || travelled >= height * dismissThreshold)
+  ) {
+    return { type: "dismiss" };
+  }
+
+  if (points.length > 1 && (flickDown || flickUp)) {
+    const ahead = points.filter((p) => (flickDown ? p > offset : p < offset));
+    if (ahead.length) {
+      return {
+        type: "settle",
+        offset: flickDown ? Math.min(...ahead) : Math.max(...ahead),
+      };
+    }
+  }
+
+  const nearest = points.reduce(
+    (best, p) => (Math.abs(p - offset) < Math.abs(best - offset) ? p : best),
+    points[0],
+  );
+  return { type: "settle", offset: nearest };
+};
+
+export const PetalDrawer = {
+  mounted() {
+    this.dragging = false;
+    this.pointerId = null;
+    this.samples = [];
+    this.offset = 0;
+    this.height = 0;
+
+    this.readConfig();
+
+    this.onPointerDown = this.onPointerDown.bind(this);
+    this.onPointerMove = this.onPointerMove.bind(this);
+    this.onPointerUp = this.onPointerUp.bind(this);
+    this.onResize = this.onResize.bind(this);
+
+    this.el.addEventListener("pointerdown", this.onPointerDown);
+    // move/up on the window, not the sheet: a fast flick outruns the element
+    // and we still need the release that happens past its edge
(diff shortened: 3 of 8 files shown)


mplatts:
## Screenshots

**Bottom sheet open** — desktop and mobile (390×844)

Sheet open
Sheet on mobile
Snapped sheet on mobile

**Side sheet unchanged** — the backward-compatibility proof

Side sheet unchanged

**Page**

Slide over, light
Slide over, dark

### Verified independently

| Check | Result |
|---|---|
| `mix test` | 930 tests, 0 failures, 1 skipped (+17) |
| `npm test` | 191 passing (+34) |
| `mix credo` | one entry *fewer*, zero new (SlideOver gained a moduledoc) |
| new dependencies | none |
| new nav slug | none — correct, this extends the existing slide-over page |

### Backward compatibility — checked directly

`git diff` on `test/petal/slide_over_test.exs` shows **zero deletions**. Every pre-existing test passes unmodified, which is the proof that matters for a component already in real apps. No existing attr default changed, and the drawer class modifier evaluates to `false` and drops out entirely for left/right/top origins.

Note the new attrs use `attr(...)` with parens, against the house bare style — that's deliberate and right here: `slide_over.ex` already had 10 parenthesised attrs, so matching the file beats matching the global convention and leaving one module internally split.

**The gap worth knowing before merge:** an end-to-end pointer *drag* was never exercised. `agent-browser`'s coordinate resolution mishandles `position: fixed` elements and `eval` is blocked in this sandbox, so every scripted drag landed on the page header instead. The physics is unit-tested as three pure exported functions (34 specs, including dismiss routing with a stubbed `liveSocket`), and the hook demonstrably runs — the snapped drawer opens at exactly `translate3d(0px, 422px, 0px)` on a 90dvh sheet in an 844px viewport, which is the `initial_snap={0.4}` peek. But drag-to-dismiss deserves a real phone pass before this leaves draft.

nhobes:
Eye-test bug, root-caused and fixed in the new commit — maintainer repro: *handle off → open → handle on → drag handle → drawer snaps down*.

**Root cause is a LiveView mechanic, not drawer logic**: the drag offset lives in the sheet's inline `transform`, but the sheet also renders a server-side `style="height: Ndvh"`. Every patch syncs the `style` attribute back to the server's version, and the only inline style LiveView carries across is its own sticky `display` — not `transform`. So ANY re-render of an open drawer silently dropped the offset from the DOM: the sheet painted at 0 while the hook still believed offset=450, and the first pointermove re-applied 450px in one frame. `updated()` only re-seated when the snap config changed, and a handle toggle doesn't change it. (Same mechanism made a dial flip while closed open the drawer at full height instead of its initial snap.)

**Fix, hook-only**: `updated()` now restores the offset (re-asserting `transition: none` mid-drag) on any patch it doesn't re-seat on; `onPointerDown` baselines the drag on the offset the sheet is *rendered* at (parsed from the inline transform) rather than what the hook last wrote — when the two disagree, the DOM is the truth. Any consumer re-render mid-open is now safe; no Elixir/CSS changes.

**CDP-verified** (420×900): the exact repro tracks 1:1 with the finger from the correct position; both snaps up/down, drag-to-dismiss, a patch landing mid-gesture, unsnapped mode, and scale_background all still behave.

Gates: `mix test` 931/0, vitest 208 (4 new specs, all 4 fail with the hook change stashed), format clean.

PR #638, opened 2026-08-12
feat(file_upload): the dropzone family over LiveView uploads

### LINKED ISSUE #605: Component: File Upload
# Component: File Upload

## Summary

A net-new `<.file_upload>` dropzone family that composes with `Phoenix.LiveView` uploads. The component receives an `@uploads.<name>` config (a `%Phoenix.LiveView.UploadConfig{}`) from `allow_upload/3` and renders the whole upload surface: a dashed dropzone with drag-over highlight (`phx-drop-target`) and click-to-browse, a file list with per-entry name/size/type icon, image thumbnail previews via `live_img_preview`, per-entry progress bars, cancel buttons, and human-readable rendering of `upload_errors/2`.

This is a Phoenix-native differentiator. Client-only React libraries (reui's "File Upload" has 10 examples) stop at the browser edge; ours is server-integrated end to end - validation, progress, and previews all come from the LiveView upload machinery, which no client-only library can match.

Note: this component **requires a LiveView**. A plain static file input already exists in `lib/petal_components/field.ex` (`<.field type="file">`, styled by `pc-file-input`); this component does not replace it and the docs must say so plainly.

## References

Study these for scope and interaction design only - we build a native HEEx/LiveView equivalent from scratch, never port code:

- reui.io File Upload (https://reui.io/components/file-upload) - variant coverage: dropzone, compact button, avatar, gallery grid
- shadcn ecosystem file upload patterns (e.g. https://github.com/origin-space/originui file upload examples) - entry list layout, error presentation
- Phoenix.LiveView uploads guide (https://hexdocs.pm/phoenix_live_view/uploads.html) - the canonical wiring this component wraps: `allow_upload/3`, `live_file_input/1`, `live_img_preview/1`, `upload_errors/2`, `cancel_upload/3`
- WAI-ARIA button pattern for the zone activation (see Accessibility below)

## API sketch

New module `lib/petal_components/file_upload.ex`, `use Phoenix.Component`. Follow the attr style in `lib/petal_components/progress.ex` (every attr has `doc:`, `values:` where finite). The builder may refine this sketch, flagging deviations in the PR.

```elixir
attr :upload, Phoenix.LiveView.UploadConfig,
  required: true,
  doc: "the upload config from allow_upload/3, e.g. @uploads.avatar"

attr :label, :string, default: nil, doc: "heading text inside the dropzone"

attr :description, :string,
  default: nil,
  doc:
    "hint line under the label. When nil, auto-derive from the config where possible: " <>
    "accepted extensions from upload.accept, max size from upload.max_file_size, " <>
    "entry count from upload.max_entries (e.g. \"PNG or JPG, up to 8 MB, max 4 files\")"

attr :variant, :string,
  default: "dropzone",
  values: ["dropzone", "compact", "avatar", "gallery"],
  doc: "dropzone = full dashed zone; compact = browse button + list, no zone; " <>
       "avatar = single circular image with replace overlay; gallery = grid of image previews"

attr :cancel_event, :string,
  default: "cancel-upload",
  doc: "phx-click event name emitted by each entry's cancel button; the parent LiveView " <>
       "handles it and calls cancel_upload/3 with the entry ref (sent as phx-value-ref)"

attr :cancel_target, :any,
  default: nil,
  doc: "phx-target for the cancel event when used inside a LiveComponent"

attr :class, :any, default: nil, doc: "CSS class for the outer wrapper"
attr :rest, :global

slot :entry,
  doc: "optional custom rendering for each in-progress entry; receives the " <>
       "%Phoenix.LiveView.UploadEntry{} as the slot argument, replacing the default entry row"
```

Design notes:

- The zone wraps `<.live_file_input upload={@upload} />` (visually hidden but focusable) and sets `phx-drop-target={@upload.ref}` on the surface. Drag-over highlight is pure CSS driven by the `phx-drop-target` active state plus a `pc-file-upload--dragover` treatment.
- Default entry row: type icon (image/document/generic via MIME sniff on `entry.client_type`), truncated `entry.client_name`, humanised `entry.client_size` (add a private `format_bytes/1` helper), progress bar (reuse the internal bar treatment from `lib/petal_components/progress.ex` or emit a slim `pc-file-upload__progress`), and a cancel button carrying `phx-click={@cancel_event} phx-value-ref={entry.ref}`.
- Image entries render `<.live_img_preview entry={entry} />` as the thumbnail; non-image entries fall back to the type icon.
- Errors: render `upload_errors(@upload)` (config-level, e.g. `:too_many_files`) and `upload_errors(@upload, entry)` (per entry, `:too_large`, `:not_accepted`) with human messages via a private `error_to_string/1`. Messages live in the component; no i18n machinery in scope.
- Entry-level `:external_client_failure` should degrade to a generic "upload failed" message rather than crash.
- Because the component only renders state it is handed, it needs no hook and no JS beyond `Phoenix.LiveView.JS` - CSS-first per house style. No new hex or npm dependencies.
- Name check: `file_upload/1` must not collide with anything exported by `use PetalComponents` (check `lib/petal_components.ex`) or by `Phoenix.Component`.

## Variants & states to build

- [ ] `dropzone` - full dashed field-surface zone (see the field surface in `lib/petal_components/input.ex` / `assets/default.css`), label + description + browse affordance, entry list below
- [ ] `compact` - browse button plus entry list, no zone
- [ ] `avatar` - single circular image preview with a hover/focus "replace" overlay; when empty, a circular dashed placeholder; forces single-entry presentation
- [ ] `gallery` - responsive grid of image preview tiles, each with progress overlay and cancel; an "add more" tile while under `max_entries`
- [ ] States per variant: empty, drag-over, uploading (partial progress), complete, config-level error, per-entry error
- [ ] `auto_upload: true` and `false` both behave (with `false`, entries sit at 0% until the form submits - the entry row must still read sensibly)
- [ ] Dark mode across all of the above (field surface + gray ramp, no invented grays)
- [ ] `--pc-radius` consumed on the zone and tiles per the radius contract

New `pc-file-upload` section in `assets/default.css` (`pc-file-upload`, `pc-file-upload--dragover`, `pc-file-upload__entry`, `pc-file-upload__progress`, variant modifiers, etc.).

## Playground page

This page is different from most: it needs a real LiveView wired with `allow_upload/3` in the playground app (`dev.exs`), not just static renders. Add a nav entry and:

- Hero: the `dropzone` variant wired to a live `allow_upload` (accept images + PDFs) so drag-and-drop, progress, and cancel actually work in the browser
- Dials (segmented controls via `handle_event`): `variant` (all four) and `max_entries` (1 / 4 / 8)
- Show both `auto_upload: true` and `auto_upload: false` configs side by side, labelled
- Scenarios (shadcn-grade, real product moments):
  1. Avatar upload - `avatar` variant, single image, replace overlay
  2. Multi-image gallery - `gallery` variant, e.g. a listing photos uploader
  3. Document list - `compact` or `dropzone` with PDFs/docs showing per-entry progress bars and a deliberately triggerable `:too_large` error (set a small `max_file_size` so visitors can see error rendering)
- Demonstrate keyboard-only operation on the page (tab to zone, Enter/Space opens the picker, tab to cancel buttons)

## Accessibility

- The zone's activation target is the (visually hidden, focusable) `live_file_input`; the zone acts as its large label/click surface. Keyboard: Tab reaches the input, Enter/Space opens the native file picker. Do not trap focus; there is no overlay here.
- The zone is a labelled region: `aria-labelledby` pointing at the label (or `aria-label` when no label attr), so screen readers announce what the zone accepts.
- Drag-over state change is announced via an `aria-live="polite"` region (e.g. "Drop files to upload"), following the pattern already used by `pc-progress__status` in `lib/petal_components/progress.ex`.
- Per-entry progress bars carry `role="progressbar"` with `aria-valuenow`/`aria-valuemin`/`aria-valuemax` and an `aria-label` naming the file.
- Cancel buttons have accessible names ("Cancel upload of filename.png"), not icon-only silence.
- Error messages are associated with the zone/entries and rendered in the danger treatment; decorative icons get `aria-hidden="true"`.
- Focus-visible ring in primary only, per house rules.

## Tests

`test/petal/file_upload_test.exs` using `ComponentCase` + LazyHTML, following the shape of existing suites (e.g. `test/petal/input_otp_test.exs`). Rendering tests run against a constructed `%Phoenix.LiveView.UploadConfig{}` (and `%Phoenix.LiveView.UploadEntry{}` structs) - no full LiveView required for the render layer. Assert:

- [ ] Each variant renders its `pc-file-upload` classes and structure
- [ ] `phx-drop-target` is set to the config ref; the `live_file_input` is present and focusable
- [ ] Entry rows: name, humanised size, cancel button with `phx-click={cancel_event}` and `phx-value-ref`
- [ ] `format_bytes/1` output (B / KB / MB boundaries)
- [ ] Image entries render a preview element; non-image entries render the type icon
- [ ] Progress bar `aria-valuenow` matches `entry.progress`; `role="progressbar"` present
- [ ] Config-level and entry-level errors render human messages for `:too_large`, `:not_accepted`, `:too_many_files`
- [ ] `description` auto-derivation from accept / max_file_size / max_entries; explicit `description` overrides it
- [ ] Custom `:entry` slot replaces the default row
- [ ] A11y: `aria-labelledby`/`aria-label` on the zone, `aria-live="polite"` region present, cancel buttons have accessible names, decorative icons `aria-hidden`

If a keyboard/drag interaction proves untestable at the render layer, cover it in the playground verification instead and note it in the PR.

## Non-goals

- No direct-to-cloud (S3/external uploader) presentation - the component renders whatever the UploadConfig reports, but external uploader docs/recipes are out of scope
- No chunk retry / pause-resume UI
- No client-side image cropping or editing (the avatar variant replaces, it does not crop)
- No replacement of the static `<.field type="file">` in `field.ex`
- No i18n system for error messages (plain English strings; users can fork via the `:entry` slot or CSS)
- No JS hook unless a genuinely un-CSS-able gap appears - flag it in the PR before adding one

## Process

1. Build on a feature branch: `lib/petal_components/file_upload.ex` + `assets/default.css` section + tests + playground page (nav entry, live `allow_upload` wiring, dials, scenarios). CSS lives in this repo - no external packages.
2. Verify in the playground (`mix run dev.exs`, http://localhost:4000) - exercise every dial and both auto_upload modes, real drag-and-drop, light + dark, keyboard-only, and `prefers-reduced-motion`.
3. Run the suites: `mix test` (and `npm test` only if a hook was added, which is not expected). All green before PR.
4. Open a PR against `main` with screenshots (light + dark) of the playground page, including an in-progress upload and an error state. The automated review must pass.

## Definition of done

- `<.file_upload>` with all four variants, entry list, previews, progress, cancel, and error rendering, composing cleanly with `allow_upload/3`
- `pc-file-upload` CSS section in `assets/default.css`, both colour schemes, radius token consumed
- Test suite green with every checklist item above asserted
- Playground page with live upload wiring, dials, and the three scenarios
- PR screenshots (light + dark)
- Docs-ready attr docs: every attr has a `doc:` string; the moduledoc shows the full wiring (mount with `allow_upload/3`, the `handle_event` for cancel, form `phx-change`/`phx-submit`) and states plainly that a LiveView is required

Target milestone: 4.16.0

mplatts:
Closes #605

## What this is

`<.file_upload>` renders the whole upload surface for a `%Phoenix.LiveView.UploadConfig{}`: the drop zone, the entry list with thumbnails and type icons, per-entry progress, cancel buttons, and human-readable errors.

Four variants (`dropzone`, `compact`, `avatar`, `gallery`), an `:entry` slot to replace the default row, and a hint line derived from the config so it cannot drift from what the server will actually accept.

## How it composes with `allow_upload/3`

It sits on top of LiveView's upload machinery rather than reimplementing any of it. Nothing here uploads bytes:

| Concern | Who does it |
|---|---|
| Drag and drop | `phx-drop-target={@upload.ref}` on the zone. LiveView adds `phx-drop-target-active`; the highlight is a plain CSS rule on that class. |
| Browse | `<.live_file_input>`, wrapped in the zone's `<label for={@upload.ref}>`, so the whole surface opens the picker. |
| Thumbnails | `<.live_img_preview>` for image entries; a MIME-sniffed type icon otherwise. |
| Progress | `entry.progress`, straight onto the bar's width and `aria-valuenow`. |
| Validation | `upload_errors/1` and `upload_errors/2`, mapped to English by a private `error_to_string/1`. |
| Cancel | A button pushing `cancel_event` with `phx-value-ref={entry.ref}`; the parent calls `cancel_upload/3`. |

The moduledoc carries a complete LiveView: `allow_upload/3` in `mount`, the component inside a `phx-change` form, the cancel `handle_event`, and `consume_uploaded_entries/3` on submit. It also says plainly that this needs a LiveView and does not replace `<.field type="file">`.

## Did you ship a hook?

**No.** No hook, no JS, no new hex or npm dependencies. LiveView's own upload machinery covered everything the issue asked for, and the two things that looked like they might need JS did not:

- **Drag-over highlight** - LiveView already adds `phx-drop-target-active` to the drop target, so it is one CSS rule.
- **Focus ring on a visually-hidden input** - the zone draws it with `:has(.pc-file-upload__input:focus-visible)`. The input is CSS-clipped (`clip-path: inset(50%)`), never `display: none`, so it stays in the tab order.

1. **Added `attr :id`** (defaults to `pc-file-upload-<ref>`). The sketch had none, but the ARIA wiring needs stable ids - `aria-labelledby` on the group and `aria-describedby` from each entry row to its error. Callers can override it.
2. **Added `attr :cancel_label`** (default `"Cancel upload of"`). The issue requires named cancel buttons; since error copy is deliberately un-i18n'd, this is the one string worth letting callers swap without taking over the whole `:entry` slot.
3. **`description=""` drops the line**, rather than only `nil` auto-deriving. Needed a way to opt out of the hint entirely.
4. **Progress bar is a local `pc-file-upload__progress`, not `<.progress>`.** The issue offered either. A slim two-element bar avoids pulling Progress's size/colour matrix into an entry row, and it still mirrors `pc-progress`'s track/fill/transition idiom so they stay visually consistent.
5. **`format_bytes/1` is SI (1 KB = 1000 B)**, so LiveView's `8_000_000` default reads as "8 MB" rather than "7.6 MB". `format_bytes/1` and `derive_description/1` are `@doc false` public so the suite can assert them directly.
6. **The avatar variant renders the entry's progress and errors in a row under the circle**, since the circle itself is the preview and has nowhere to put a bar.

## One bug worth calling out

`allow_upload/3` stores `:accept` **comma-joined as a string** (`".png,.jpg"`), not as a list - it is built for the input's `accept` attribute. The struct's typespec says `list() | :any` and its default is `[]`, so a list-only implementation passes every unit test and then silently drops the whole accepted-types phrase against a real config. Caught it in the playground, not the suite. Both forms are handled now, with a regression test naming the reason.

## Accessibility

- The native input stays focusable and operable: Tab reaches it, Enter/Space opens the picker. The zone shows the focus ring on its behalf.
- Wrapper is `role="group"`, `aria-labelledby` the label (or `aria-label="File upload"` when there is none).
- Each bar is `role="progressbar"` with `aria-valuenow`/`min`/`max`/`valuetext` and an `aria-label` naming the file.
- Cancel buttons read "Cancel upload of q3.pdf", never icon-only silence.
- Entry errors are tied to their row via `aria-describedby`; decorative icons are `aria-hidden="true"`.
- Focus-visible ring in primary only, no persistent `:focus` fills. Reduced motion kills the decorative fades and leaves the progress width transition alone, matching `pc-progress`.

**One honest caveat:** the drag-over hint is an `aria-live="polite"` region that CSS fades in on `phx-drop-target-active`. A CSS visibility change is not a text mutation, so it will not reliably *announce* on dragenter - it would need a hook to swap the text. I did not add one: screen reader users reach this through the file input, not by dragging, and the house rule is to name the gap rather than reach for JS. Happy to add a hook if you want the announcement.

## Playground

`/c/file-upload`, wired to **six real `allow_upload/3` configs** (not static mocks):

- Hero dropzone with the `variant` and `max_entries` dials. `max_entries` is fixed at `allow_upload` time, so changing it does `disallow_upload` then `allow_upload` again.
- `auto_upload: true` and `false` side by side, labelled - the manual one shows the entry sitting at 0% until submit.
- Avatar, gallery, and a documents zone with `max_file_size: 20_000` so `:too_large` is trivially triggerable.
- Save consumes via `consume_uploaded_entries/3` (nothing is written to disk) and echoes the consumed names.

Verified in a real browser with real files: thumbnails render, progress runs to 100%, cancel works, the error row turns red with the message inside it, and the variant dial swaps all four. Light and dark both checked.

## Tests

| Suite | Before | After |
|---|---|---|
| Elixir | 913 (1 skipped) | **968** (1 skipped) |
| JS | 157 | **157** (no hook, so unchanged) |

56 new tests in `test/petal/file_upload_test.exs`, built against hand-constructed `UploadConfig`/`UploadEntry` structs - every variant, the drop target and input wiring, entry rows, `format_bytes/1` boundaries, previews vs type icons, progress ARIA, all four error classes, description derivation (both accept representations), the `:entry` slot, the full a11y surface, and `class`/`rest`/`id` pass-through.

`mix credo` shows **zero new entries** vs `origin/main` (diffed entry-by-entry, 45 both sides). `mix format --check-formatted` and `mix compile --force --warnings-as-errors` are clean. No test was weakened or removed.

## Screenshots

Light, dark, and a live upload in flight are attached below.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,24 @@
 # Changelog
+
+### Unreleased
+
+#### Added
+
+- **New `file_upload` component: the whole upload surface for a Phoenix
+  LiveView upload.** Hand it an `@uploads.<name>` from `allow_upload/3`
+  and it renders the drop zone, the file list with thumbnails and type
+  icons, per-entry progress, cancel buttons and the errors from
+  `upload_errors/1` and `upload_errors/2` in plain English. It sits on
+  top of LiveView's upload machinery rather than beside it: drag and
+  drop is `phx-drop-target`, previews are `live_img_preview`, progress
+  is `entry.progress`. No hook, no JS, no new dependencies.
+  Four variants - `dropzone`, `compact`, `avatar` and `gallery` - plus
+  an `:entry` slot when you want to render the rows yourself. The hint
+  line ("PNG or JPG, up to 8 MB, max 4 files") is derived from the
+  config, so it cannot drift from what the server will actually accept.
+  Requires a LiveView; `<.field type="file">` is still the static
+  option and is unchanged.
+
 ### 4.14.0 - 2026-08-11
 
 #### Added
--- a/assets/default.css
+++ b/assets/default.css
@@ -7586,3 +7586,267 @@
       transition-duration: 1ms;
     }
   }
+
+/* File upload - the dropzone family over Phoenix.LiveView uploads. Drag-over
+ * is LiveView's own `phx-drop-target-active` class, so the highlight is plain
+ * CSS and the component ships no hook. The native input stays in the tab
+ * order (clipped, never display:none), and because it is invisible the zone
+ * that labels it draws the focus ring on its behalf via :has(). */
+@layer components {
+  .pc-file-upload {
+    @apply flex flex-col w-full gap-3;
+  }
+
+  /* Clipped to nothing but still focusable: keyboard users tab to the input
+     and Enter/Space opens the native picker. `display: none` or `hidden`
+     would drop it out of the tab order and strand them. */
+  .pc-file-upload__input {
+    @apply absolute w-px h-px p-0 -m-px overflow-hidden whitespace-nowrap border-0;
+    clip-path: inset(50%);
+  }
+
+  /* The drop surface: a label, so clicking anywhere on it opens the picker
+     and LiveView's phx-drop-target lands on the same element. */
+  .pc-file-upload__zone {
+    @apply relative flex flex-col items-center justify-center w-full gap-1 px-6 py-10 text-center cursor-pointer;
+    @apply bg-transparent border border-dashed border-gray-300 shadow-xs;
+    @apply transition-[color,background-color,border-color,box-shadow] duration-200 ease-out;
+    @apply hover:border-gray-400 hover:bg-gray-50;
+    @apply dark:bg-gray-400/8 dark:border-gray-400/25 dark:hover:border-gray-400/40 dark:hover:bg-gray-400/12;
+    border-radius: var(--pc-radius, 0.625rem);
+  }
+
+  .pc-file-upload__zone:has(.pc-file-upload__input:focus-visible) {
+    @apply outline-hidden border-primary-500 ring-2 ring-primary-500/50;
+  }
+
+  .pc-file-upload__zone.phx-drop-target-active {
+    @apply border-primary-500 bg-primary-500/8 dark:bg-primary-500/12;
+  }
+
+  .pc-file-upload__heading {
+    @apply flex flex-col gap-0.5;
+  }
+
+  .pc-file-upload__glyph {
+    @apply w-8 h-8 text-gray-400 dark:text-gray-500;
+  }
+
+  .pc-file-upload__label {
+    @apply text-sm font-medium text-gray-900 dark:text-gray-100;
+  }
+
+  .pc-file-upload__description {
+    @apply text-sm text-gray-500 dark:text-gray-400;
+  }
+
+  /* The drag hint doubles as the polite live region, so it is faded rather
+     than removed - a live region has to stay in the accessibility tree. */
+  .pc-file-upload__live {
+    @apply absolute inset-x-0 bottom-2 text-xs font-medium opacity-0 pointer-events-none;
+    @apply text-primary-600 transition-opacity duration-150 dark:text-primary-400;
+  }
+
+  .pc-file-upload__zone.phx-drop-target-active .pc-file-upload__live {
+    @apply opacity-100;
+  }
+
+  /* File upload - entries */
+
+  .pc-file-upload__entries {
+    @apply flex flex-col gap-2 p-0 m-0 list-none;
+  }
+
+  .pc-file-upload__entry {
+    @apply list-none;
+  }
+
+  .pc-file-upload__entry-inner {
+    @apply flex items-center w-full gap-3 px-3 py-2 bg-transparent border border-gray-200;
+    @apply dark:border-gray-400/20 dark:bg-gray-400/8;
+    border-radius: max(calc(var(--pc-radius, 0.625rem) - 0.25rem), 0.25rem);
+  }
+
+  .pc-file-upload__entry-inner--invalid {
+    @apply border-danger-500 dark:border-danger-500/70;
+  }
+
+  .pc-file-upload__media {
+    @apply flex items-center justify-center w-10 h-10 overflow-hidden shrink-0;
+    @apply bg-gray-100 dark:bg-gray-400/12;
+    border-radius: max(calc(var(--pc-radius, 0.625rem) - 0.25rem), 0.25rem);
+  }
+
+  .pc-file-upload__thumb {
+    @apply object-cover w-full h-full;
+  }
+
+  .pc-file-upload__type-icon {
+    @apply w-5 h-5 text-gray-500 dark:text-gray-400;
+  }
+
+  .pc-file-upload__entry-body {
+    @apply flex flex-col min-w-0 gap-1 grow;
+  }
+
+  .pc-file-upload__entry-name {
+    @apply block text-sm font-medium truncate text-gray-900 dark:text-gray-100;
+  }
+
+  .pc-file-upload__entry-meta {
+    @apply block text-xs tabular-nums text-gray-500 dark:text-gray-400;
+  }
+
+  /* File upload - progress. Mirrors pc-progress: washed track, solid fill,
+     width driven inline from entry.progress. The width transition is
+     informative, so reduced motion leaves it alone. */
+
+  .pc-file-upload__progress {
+    @apply relative flex w-full h-1 overflow-hidden rounded-full bg-primary-500/15;
+  }
+
+  .pc-file-upload__progress-bar {
+    @apply relative z-[1] block h-full rounded-full bg-primary-600;
+    @apply transition-[width] duration-500 ease-out;
+  }
+
+  .pc-file-upload__entry-inner--invalid .pc-file-upload__progress {
+    @apply bg-danger-500/15;
+  }
+
+  .pc-file-upload__entry-inner--invalid .pc-file-upload__progress-bar {
+    @apply bg-danger-600;
+  }
+
+  /* File upload - errors, matching the pc-form-field-error treatment */
+
+  .pc-file-upload__error {
+    @apply flex items-center gap-1.5 text-sm text-danger-600 dark:text-danger-400;
+  }
+
+  .pc-file-upload__error--entry {
+    @apply mt-1.5 text-xs;
+  }
+
+  .pc-file-upload__error-icon {
+    @apply w-4 h-4 shrink-0;
+  }
+
+  /* File upload - cancel */
+
+  .pc-file-upload__cancel {
+    @apply flex items-center justify-center w-7 h-7 shrink-0 cursor-pointer;
+    @apply text-gray-400 transition-colors duration-150 hover:text-gray-700;
+    @apply dark:text-gray-500 dark:hover:text-gray-200;
+    @apply focus:outline-hidden focus-visible:ring-2 focus-visible:ring-primary-500/50;
+    border-radius: max(calc(var(--pc-radius, 0.625rem) - 0.25rem), 0.25rem);
+  }
+
+  .pc-file-upload__cancel-icon {
+    @apply w-4 h-4;
+  }
+
+  /* File upload - compact: a browse button, no zone */
+
+  .pc-file-upload__bar {
+    @apply flex flex-wrap items-center gap-3;
+  }
+
+  .pc-file-upload--compact .pc-file-upload__zone {
+    @apply flex-row w-auto gap-2 px-4 py-2 border-solid;
+    @apply text-sm font-medium text-gray-700 dark:text-gray-200;
+  }
+
+  .pc-file-upload--compact .pc-file-upload__glyph {
+    @apply w-4 h-4;
+  }
+
+  /* File upload - avatar: one circular preview with a replace overlay */
+
+  .pc-file-upload--avatar .pc-file-upload__zone {
+    @apply relative w-24 h-24 p-0 overflow-hidden rounded-full shrink-0;
+  }
+
+  .pc-file-upload__avatar-img {
+    @apply object-cover w-full h-full;
+  }
+
+  .pc-file-upload--avatar .pc-file-upload__glyph {
+    @apply w-10 h-10;
+  }
+
+  .pc-file-upload__overlay {
+    @apply absolute inset-0 flex items-center justify-center text-white opacity-0;
+    @apply bg-gray-900/60 transition-opacity duration-150;
+  }
+
+  .pc-file-upload--avatar .pc-file-upload__zone:hover .pc-file-upload__overlay,
+  .pc-file-upload--avatar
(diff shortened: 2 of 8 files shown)


mplatts:
## Screenshots

**With real entries** — actual PNG uploads, thumbnails, progress, cancel buttons

File upload with entries

**Error state** — a deliberate `:too_large`

Upload errors

**Light / dark**

File upload, light
File upload, dark

### Verified independently

| Check | Result |
|---|---|
| `mix test` | 968 tests, 0 failures, 1 skipped (**+55**) |
| `npm test` | 157 passing, unchanged (no hook) |
| new dependencies | none |
| slugs : clauses | 61 : 61 |

## The bug worth reading

`allow_upload/3` stores `:accept` **comma-joined as a string**, not a list — even though the struct's typespec says `list() | :any` and it defaults to `[]`. A list-only implementation passes every unit test and then silently drops the entire accepted-types phrase against a real config.

That is a genuinely nasty class of bug: the types say one thing, the runtime does another, and no amount of unit testing catches it. It surfaced only because the playground renders a real `allow_upload` and the live page read "up to 8 MB, max 4 files" with no file types. Fixed, with a regression test that names the reason.

This is a good argument for the CONTRIBUTING rule that the playground page is part of the deliverable rather than an afterthought.

Shipping zero JS is the right call here — `phx-drop-target-active` is a class LiveView adds itself, so drag-over is one CSS rule, and the clipped-but-focusable input's focus ring is drawn by the zone via `:has()`.

**@nhobes — one honest a11y gap flagged by the author:** the drag-over `aria-live` region is faded in by CSS, and a CSS visibility change is not a text mutation, so it will not reliably *announce* on dragenter. Fixing it needs a hook to swap the text. Not added unilaterally. Screen reader users reach this via the file input rather than by dragging, so it is a polish item rather than a blocker.

nhobes:
Eye-test find, root-caused and fixed in the new commit — the gallery demo's broken-image glyphs.

**Where `src="null"` actually came from**: `live_img_preview` ships an `` with no src and lets its hook fill one in from the File the browser holds. When the hook can't find the File (a static showcase, a restored preview), LiveView's own hook assigns the `null` that `getEntryDataURL` returns — and `img.src = null` serializes to the literal string `"null"`, a relative URL that 404s into the broken glyph. Real apps also flash the same glyph on every first paint before the hook runs.

**Fix 1 — CSS guard**: thumbs (and the avatar circle) are `opacity-0` whenever src is missing/empty/"null", with a 150ms fade when hydration lands — the media placeholder shows through as a quiet pending state instead of a broken one.

**Fix 2 — the `:existing` slot** (the real API this surfaced): edit forms show *already-uploaded* photos by URL alongside new client uploads. Slot items (`src` + `name` required) render as plain `` tiles before the in-flight entries — no hook, no progress bar, and their remove button is the consumer's own event (`remove_event`/`remove_value`/`remove_target`), because a saved photo is a database row, not something `cancel_upload/3` knows about. The moduledoc spells out the edit-listing pattern and the `max_entries` semantics (it caps what the browser may still add, not the saved set). The avatar variant takes the same slot — "here is your current photo, click to replace it" — with a pick superseding the stored image.

**Showcase**: the gallery demo now renders real static images instead of hookless phantoms — the broken glyphs are gone and the demo shows genuine thumbnails next to the faked progress bars.

Gate note: the building agent stalled before finishing its own verification, so the suite run here is the gate: `mix test` 987/0 (exit 0, includes 196 lines of new specs pinning URL-items render plain img with no hook, hook entries unchanged, avatar precedence, remove-button wiring), format clean. Browser-side both-theme verification was reported complete before the stall but not written up — worth an extra glance at /c/file-upload during the eye-test.

PR #639, opened 2026-08-12
feat(filters): standalone filter bar driven by DataTable.State

### LINKED ISSUE #619: Component: Filters

## Summary

A standalone, schema-driven filter bar: `<.filters>` renders the active filters as removable chips (field, operator, value) plus an "Add filter" popover that walks field -> operator -> value, all driven by the `PetalComponents.DataTable.State` struct that already ships. Because it speaks `State`, it drops in next to a `<.data_table>` sharing the same struct, or stands alone above any list, card grid, or custom query UI. This is the LiveView answer to the URL-state filter panels React needs nuqs/zod for: `handle_params` plus `State.from_params/2` already make the URL the state store, so the component only has to render state and emit changes.

Before designing anything, study two files in this repo:

- `lib/petal_components/data_table/state.ex` - the 17-operator filter vocabulary (`State.ops/0`, `valueless_op?/1`), the pinned semantics (case folding, `:between` inclusivity, `:is_empty` rules, date-shaped ops), `put_filter/4`, `handle_op/3`, and the URL round-trip via `from_params/2` / `to_params/1`.
- `lib/petal_components/data_table.ex` - the two wiring modes (link mode via `path` + patch URLs, event mode via `on_change` + op-shaped payloads), and the existing per-column filter popover editors (around the `pc-data-table__filter` markup).

**One source of truth rule (non-negotiable):** do not fork the operator vocabulary. The component's type-to-operators mapping selects subsets of `State.ops/0`; it never defines its own operator atoms, never re-pins semantics, and never bypasses `put_filter/4` / `handle_op/3` when mutating state. If the data table and the filter bar ever disagree about what `:contains` means, this component has failed.

## References

- reui.io "Filters" (12 examples) - study the chip anatomy (field + op + value + remove), the add-filter flow, and which value editors they ship per type. Theirs is nuqs/zod-based React; we reference behaviour and variants only, never code.
- shadcn/Radix Popover and Command patterns for the add-filter popover interaction.
- Linear and Notion filter bars for the field -> operator -> value progressive flow.

## API sketch

New module `lib/petal_components/filters.ex`, `use Phoenix.Component`, exporting `filters/1`. Wiring mirrors `data_table` exactly: mode is inferred from which of `path` / `on_change` is set, and passing neither raises, same as `data_table/1` does today.

```elixir
attr :state, PetalComponents.DataTable.State, required: true

attr :path, :string, default: nil,
  doc: "link mode: the base path filter changes patch to, state encoded via State.to_params/1. Required unless on_change is set."

attr :on_change, :string, default: nil,
  doc: "event mode: the event filter edits push, with the same op-shaped payloads State.handle_op/3 already accepts (op filter / clear_filters)."

attr :target, :any, default: nil, doc: "event mode: the phx-target (e.g. @myself)"

attr :add_filter_label, :string, default: "Add filter"
attr :clear_filters_label, :string, default: "Clear filters"
attr :apply_label, :string, default: "Apply", doc: "the value editor's submit label, localizable"
attr :filter_op_labels, :map, default: %{},
  doc: "overrides for operator display names, e.g. %{contains: \"enthält\"} - same attr name and shape as data_table"

slot :field, required: true do
  attr :field, :atom, required: true
  attr :label, :string
  attr :type, :string,
    values: ["text", "select", "multi", "date_range", "boolean", "number_range"]
  attr :options, :list,
    doc: "select/multi: the pickable values, as strings or {label, value} tuples - same shape as data_table's :col options"
end
```

The `:field` slot is the registry, following the `:col` slot precedent in `data_table`. Type selects the operator subset and the value editor:

| type | operators (all from `State.ops/0`) | value editor |
|---|---|---|
| `text` | `contains`, `not_contains`, `eq`, `neq`, `starts_with`, `is_empty`, `is_not_empty` | text input |
| `select` | `eq`, `neq`, `is_empty`, `is_not_empty` | single select from `options` |
| `multi` | `in`, `not_in` | checkbox list from `options` |
| `number_range` | `eq`, `neq`, `gt`, `gte`, `lt`, `lte`, `between`, `is_empty`, `is_not_empty` | number input, two inputs for `between` |
| `date_range` | `before`, `on`, `after`, `between`, `is_empty`, `is_not_empty` | date input, two for `between` |
| `boolean` | `eq` | yes/no toggle (no operator picker - there is only one) |

Valueless operators (`is_empty`, `is_not_empty`) render no value input, per `State.valueless_op?/1`.

Emissions:

- **Event mode**: chip removal, filter apply, and clear-all push `on_change` with the exact payloads `State.handle_op/3` already parses (`%{"op" => "filter", "field" => f, "filter_op" => op, "value" => v, "value2" => v2}`, `values` list for multi, `%{"op" => "clear_filters"}`). A LiveView already wired for an event-mode `data_table` needs zero new handler clauses to gain a filter bar.
- **Link mode**: every change is a `patch` URL built with `State.to_params/1` against `path`, exactly like `data_table` link mode. `handle_params` + `State.from_params/2` is the whole backend.

The builder may refine this sketch (flagging deviations in the PR). One refinement to consider seriously: the filter editors inside `data_table.ex` (typed value inputs, operator select, apply/clear buttons) overlap heavily with what this component needs. Prefer extracting them into a shared private module (e.g. `PetalComponents.DataTable.FilterEditor`) consumed by both, over copy-pasting, but do not change the data table's rendered markup or event grammar in the process - its test suite is the contract.

## Variants & states to build

- Empty state: just the "Add filter" trigger, no chips, no clear-all.
- One or more active chips: each shows field label, humanised operator (via `filter_op_labels` merge, same defaults as `data_table`), formatted value, and a remove button. `between` renders both bounds; `in` renders a count or truncated list ("Status: 3 selected"); valueless ops render label + operator only.
- "Clear filters" affordance, shown only when at least one filter is active.
- Add-filter popover: field list -> operator picker (skipped for boolean) -> value editor -> apply. Reuse the `<.popover>` / floating-panel surface conventions (`bg-white dark:bg-gray-900`, real borders, `--pc-radius` token).
- Editing an existing chip: clicking the chip body reopens the editor pre-filled with the current operator and value.
- Chip overflow: chips wrap; no horizontal scroll on the page body.
- Both wiring modes, both colour schemes, `prefers-reduced-motion` (any popover transition must have a complete rest state).

New `pc-filters` section in `assets/default.css` (`pc-filters`, `pc-filters__chip`, `pc-filters__chip-remove`, `pc-filters__add`, `pc-filters__editor`, etc.). If a hook is genuinely needed for the popover (the data table uses its `PetalDataTable` hook for in-page filter panels; `PetalPopover` exists for top-layer flip), justify it against the CSS-first rule in the PR and prefer reusing an existing hook over adding a new one.

## Playground page

A products list (name, category, price, in stock, added date) demonstrating:

- Dials: mode (link | event), and toggles for which field types are registered.
- Example 1: filter bar alone above a plain card/list of products, event mode.
- Example 2: filter bar next to a `<.data_table>` sharing one `State` - filter from the bar, watch the table update; filter from a column header, watch a chip appear. This is the composition proof and the hero shot.
- Example 3: link mode "shareable URL" demo - apply filters, show the resulting query string, and demonstrate that pasting the URL reproduces the state (the back button works).

Keyboard-only operation of the add-filter flow should be demonstrable on the page.

## Accessibility

- The add-filter trigger is a `button` with `aria-haspopup` and `aria-expanded`; the popover follows the WAI-ARIA menu/dialog hybrid used by the existing filter editors - study the roles the data table's filter popovers already emit and match them.
- Keyboard map: Enter/Space opens the popover; arrow keys move through the field and operator lists; Tab reaches the value inputs; Enter applies; Escape closes without applying and returns focus to the trigger. Focus returns to a sensible neighbour after chip removal (next chip, else the add trigger), never to `body`.
- Each chip's remove button has an accessible name naming the filter ("Remove filter: Status is active"), not a bare "x".
- The chip list is announced as a group ("Active filters") via `role="group"` + `aria-label`.
- Decorative icons get `aria-hidden`.
- Focus-visible ring in primary only, no persistent focus fills.

## Tests

`test/petal/filters_test.exs` using `ComponentCase` and LazyHTML. Reuse the link-mode assertion patterns from `test/petal/data_table_test.exs` (patch URL href assertions built from `State.to_params/1`).

- Chip rendering from state: for a `State` with active filters, each chip renders field label, operator label, value; valueless ops render without a value; `between` renders both bounds; `in` renders the multi summary.
- Operator vocabulary per type: for each registry type, the operator picker offers exactly the subset in the table above, and every offered atom is a member of `State.ops/0` (assert this programmatically so a future vocabulary change fails loudly here).
- Add-filter flow, event mode: the apply form's hidden inputs and names produce exactly the payload shapes `State.handle_op/3` accepts; a round-trip test feeds the rendered payload through `handle_op/3` and asserts the expected `State`.
- Chip removal and clear-all payloads, event mode.
- Link mode: chip remove links and the clear-all link carry patch hrefs whose query params decode via `State.from_params/2` back to the expected state; defaults are omitted from URLs (clean-URL contract of `to_params/1`).
- Raises when neither `path` nor `on_change` is given.
- A11y assertions: `aria-expanded`, chip group labelling, remove-button accessible names.
- `filter_op_labels` overrides reach both chips and the operator picker.
- If editor internals are extracted from `data_table.ex`, the full data table suite must still pass untouched.

## Non-goals

- No new filter operators, and no changes to operator semantics - `State` owns the vocabulary.
- No saved/named filter sets, no AND/OR grouping, no nested predicates (the state model is a flat one-filter-per-field list; honour it).
- No date-picker calendar widget - native date inputs only.
- No quick-search input (that is `data_table`'s `searchable`; this component covers structured filters only).
- No sorting or pagination UI.
- No Alpine.js wiring.

## Process

1. Build on a feature branch: `lib/petal_components/filters.ex` + `pc-filters` section in `assets/default.css` + tests + playground page (nav entry, dials, the three examples). CSS/JS live in this repo, no external packages.
2. Verify in the playground (`mix run dev.exs`, http://localhost:4000): every dial, light + dark, keyboard-only, `prefers-reduced-motion`, and the shared-State demo against a real `data_table`.
3. Run the suites: `mix test` (and `npm test` if a JS hook was touched). All green before PR.
4. Open a PR against `main` with light + dark screenshots of the playground page. The automated review must pass.
6. Release (Hex publish, docs, changelog, MCP schema sync) is handled by maintainers after merge, not part of the contribution.

## Definition of done

- `<.filters>` shipped with the API above (or a flagged refinement), exported via `use PetalComponents` without name collisions.
- Operator vocabulary sourced from `State.ops/0` with the per-type subsets tested.
- Both wiring modes proven: event mode drives `State.handle_op/3` with zero new grammar; link mode round-trips through `to_params/1` / `from_params/2`.
- `pc-filters` CSS section, floating-panel surface and radius tokens respected, both colour schemes.
- Tests green, including the untouched `data_table` suite.
- Playground page with dials, the standalone example, the shared-State composition example, and the shareable-URL demo.
- PR screenshots (light + dark).
- Docs-ready attr docs: every attr and slot attr has a `doc:` string; the moduledoc includes usage examples for both modes and the shared-State composition.

mplatts:
Closes #619

## Summary

`<.filters>` renders the active filters as removable chips (field, humanised operator, formatted value) next to an "Add filter" popover that walks field -> operator -> value inside one panel, no server round trip. Both wiring modes ship, inferred the same way `data_table/1` infers them (passing neither raises).

Five parts per CONTRIBUTING.md:

1. `lib/petal_components/filters.ex` - the component, exported via `use PetalComponents`. Every attr and slot attr carries a `doc:`; the moduledoc carries usage examples for both modes and for the shared-State composition.
2. `assets/default.css` - a `pc-filters` section appended inside `@layer components`, nothing above it reflowed. Consumes `--pc-radius`, rides the gray ramp, dark via `dark:`, `focus-visible` only (no persistent `:focus` fills), `prefers-reduced-motion` respected.
3. `test/petal/filters_test.exs` - 41 tests.
4. `lib/petal_components/showcase/filters.ex` + its `registry.ex` entry.
5. A playground page: nav entry `filters` (one slug, 1:1), a field-types dial, and three scenarios - the standalone bar over a product list, the shared-State composition against a live `<.data_table>`, and a link-mode demo printing the query string it decoded.

Plus `### Unreleased` / `#### Added` in `CHANGELOG.md`.

## Relationship to `DataTable.State` - it drives it directly

Not independent. The component takes a `%State{}` and mutates it only through `State.put_filter/4` (link mode, server-rendered hrefs) and `State.handle_op/3` (event mode). It never defines an operator atom, never re-pins semantics, and never invents a payload shape. The per-type operator lists are subsets of `State.ops/0`, asserted programmatically in the suite so a vocabulary change fails loudly here.

Concretely: chip removal posts `%{"op" => "filter", "field" => f}` with no `filter_op`, which is exactly how the data table's inline clear already reads as a removal; clear-all posts `%{"op" => "clear_filters"}`; the editor form posts `filter_op` / `value` / `value2` / `values[]`, byte-for-byte the shapes `handle_op/3` parses. A LiveView already wired for an event-mode `data_table` gains a filter bar with **zero** new handler clauses. Link mode patches URLs built from `State.to_params/1`, so `handle_params` + `from_params/2` is the whole backend and the state is curl-able.

Because both components read one struct, pointing them at the same assign composes with no glue - visible in the playground's second example and in `active-light.png`, where the bar's chips and the table's column-filter triggers show the same predicates and the table has narrowed to the matching rows.

### The shared extraction

The issue suggested extracting the overlapping editors rather than copy-pasting, so `lib/petal_components/data_table/filter_editor.ex` (`@moduledoc false`) now owns the typed value editors, the operator display names, the value formatting, and the `State` -> URL encoding. `data_table.ex` consumes it.

Two deliberate choices there:

- Editors are named by **shape** (`text`, `number`, `date`, `options`, `choice`, `boolean`), not by either caller's type vocabulary - the data table's `select` is a checkbox list while the filter bar's `select` is a single choice, so a shared name keyed on "select" would have been a trap.
- The `pc-data-table__filter-*` class names are kept on the shared form. The `PetalDataTable` hook queries them and `default.css` already carries the JS-free `:has` rules that hide the value input for a valueless op and reveal the second bound for `:between`. Renaming would have meant touching the hook and its specs for no behavioural gain.

**The data table's rendered markup and event grammar are unchanged, and its full suite passes unmodified** - that is the backward-compatibility proof. `data_table.ex` net-shrinks by ~180 lines.

## Did I ship a hook? No

No new hook and **no JS changes at all** (`npm test` is untouched at 157). The existing `PetalDataTable` hook already does everything this needs: it drives in-page `[data-pc-menu]` panels (open/close, outside-tap dismiss, Escape with focus restore, flip and clamp on resize) and, in link mode, reads a `.pc-data-table__filter-form` back into a patch URL from a `:filters` nav template. The filter bar mounts the same hook and speaks the same data attributes. The issue's own guidance was to prefer reusing an existing hook, and this reuse cost zero new lines of JS.

Everything else is CSS-first: the add panel's step swap is `Phoenix.LiveView.JS` class toggling over CSS-owned `display`, so there is a complete rest state at both ends and nothing for reduced motion to sit through; the `between` second bound is still the existing `:has` rule.

1. **`multi` supports `:in` only, not `:not_in`.** Both paths that turn an editor into a filter hardcode `:in` when a `values[]` list is present - `State.handle_op/3` (`is_list(params["values"])` wins before `filter_op` is read) and the hook's `readFilter`. Offering `:not_in` in the picker would have rendered an operator that silently became `:in`. Getting it working would need either a `State` grammar change or a hook change, both of which the issue rules out ("no changes to operator semantics", "never bypasses `handle_op/3`"). I chose the honest subset. **Flagging this as the one place the sketch and the shipped API differ on vocabulary** - happy to follow up with an `:not_in` path in `State` if you want it.

2. **The field list is a `role="group"` of buttons navigated by Tab, not a `role="menu"` navigated by arrow keys.** Arrow-key roving needs JS to move focus, and the only way to get it was new hook code. By not claiming menu semantics the pattern stays WAI-ARIA conformant as-is (a labelled group of buttons; Tab is the correct navigation for that role) and the whole flow is keyboard-operable: Tab to the trigger, Enter/Space opens, Tab through the field list, Enter picks, Tab reaches the inputs, Enter applies, Escape closes and returns focus to the trigger. The operator list is a native `<select>`, so arrow keys work there natively as the issue asks. Say the word and I'll add the roving handler.

3. **The "Add filter" trigger always renders**, even when every field is filtered (the panel then says so). It doubles as the guaranteed focus target after the last chip is removed - a trigger that appears only *after* the removal would be missing at the moment `JS.focus` runs.

4. **The add list offers only unfiltered fields.** The state model is one filter per field, and an already-filtered field is edited from its chip, so listing it twice would have meant two forms for one field.

5. **Extra localizable attrs** beyond the sketch: `remove_filter_label`, `active_filters_label`, `no_filters_label`, `all_fields_used_label`, `filter_options_placeholder`. Every user-facing string is overridable, asserted in the suite.

6. **`date_range` gains `:between`** (the data table's date editor has none), so the shared date editor renders a second bound only when `:between` is in the offered ops - which keeps the data table's markup identical.

## Accessibility

- Add trigger and chip bodies: `aria-haspopup="dialog"`, `aria-expanded`, `aria-controls`.
- Chip list: `role="group"` + `aria-label="Active filters"`.
- Remove buttons name the filter, never a bare x: `aria-label="Remove filter: Category is Power tools"` (option **labels**, not posted values).
- Focus after removal goes to the next chip, else the add trigger - never `body`. Each chip's remove chains `JS.focus` onto the push (event mode) or the patch link (link mode).
- An `sr-only` `role="status"` region reports the resulting filter set, so a removal is announced rather than silent.
- Every editor control has an accessible name; decorative icons carry `aria-hidden="true"`.
- `focus-visible` rings only.

## Verification

| check | result |
|---|---|
| `mix credo` | **zero new entries** vs `origin/main` (diffed `--all --format=oneline`; the only delta is a line-number shift on a pre-existing `data_table.ex` complexity entry) |
| `mix test` | **913 -> 954**, 0 failures, 1 skipped (+41 new; every pre-existing data-table test passes unmodified) |
| `npm test` | **157 -> 157**, 0 failures (no JS touched) |

No test was weakened or deleted.

Verified in the playground on light and dark, and driven end to end in a real browser: opened the add panel, picked a field, watched the panel swap to the editor, selected `between` and watched the second bound appear from the `:has` rule alone, applied, and saw the chip, the narrowed list, the narrowed shared table and the URL round-trip in link mode.

## Screenshots

Captured from the playground page at 1280x2200, both schemes, empty and with filters applied. Not committed to the repo - I'll drop them into this thread rather than into `pr-assets`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,24 @@
 # Changelog
+
+### Unreleased
+
+#### Added
+
+- **New `filters` component: a standalone filter bar driven by
+  `DataTable.State`.** Active filters render as removable chips (field,
+  humanised operator, formatted value) next to an "Add filter" popover
+  that walks field -> operator -> value in one panel, with no server
+  round trip. It speaks the same struct `<.data_table>` does, so pointing
+  both at one `State` composes with no glue: filter from the bar and the
+  table updates, filter from a column header and a chip appears. Six
+  registry types (`text`, `select`, `multi`, `number_range`,
+  `date_range`, `boolean`) each select a subset of `State.ops/0` - the
+  operator vocabulary is never forked. Both wiring modes ship: link mode
+  patches `State.to_params/1` URLs so filters are shareable and the back
+  button works, and event mode pushes the exact payloads
+  `State.handle_op/3` already parses, so a LiveView wired for an
+  event-mode data table gains a filter bar with zero new handler clauses.
+
 ### 4.14.0 - 2026-08-11
 
 #### Added
--- a/assets/default.css
+++ b/assets/default.css
@@ -7586,3 +7586,97 @@
       transition-duration: 1ms;
     }
   }
+
+  /* Filters: the standalone filter bar (chips + add-filter popover).
+     Shares the data table's editor form styling (pc-data-table__filter-*),
+     because the editor markup itself is shared - one home for the operator
+     select, the typed value inputs, and the JS-free :has rules that hide the
+     value for a valueless op and reveal the second bound for :between. */
+
+  .pc-filters {
+    @apply flex flex-wrap items-center gap-2;
+  }
+  .pc-filters__chips {
+    @apply flex flex-wrap items-center gap-2;
+  }
+  /* wins over pc-popover's inline-block: a chip is a row of parts */
+  .pc-filters__chip {
+    @apply relative inline-flex max-w-full items-center border border-gray-300 bg-white text-sm dark:border-gray-400/25 dark:bg-gray-400/8;
+    border-radius: min(var(--pc-radius, 0.625rem), 1rem);
+  }
+  .pc-filters__chip-body {
+    @apply flex min-w-0 items-center gap-1.5 py-1 pl-2.5 pr-1 text-left transition-colors duration-150 hover:bg-gray-50 focus:outline-hidden focus-visible:ring-2 focus-visible:ring-primary-500/50 dark:hover:bg-gray-400/10;
+    border-start-start-radius: inherit;
+    border-end-start-radius: inherit;
+  }
+  .pc-filters__chip-field {
+    @apply shrink-0 font-medium text-gray-900 dark:text-gray-100;
+  }
+  .pc-filters__chip-op {
+    @apply shrink-0 text-gray-500 dark:text-gray-400;
+  }
+  .pc-filters__chip-value {
+    @apply min-w-0 truncate font-medium text-primary-700 dark:text-primary-300;
+    max-width: 12rem;
+  }
+  .pc-filters__chip-remove {
+    @apply mr-1 inline-flex h-6 w-6 shrink-0 items-center justify-center rounded-full text-gray-400 transition-colors duration-150 hover:bg-gray-100 hover:text-gray-600 focus:outline-hidden focus-visible:ring-2 focus-visible:ring-primary-500/50 dark:hover:bg-gray-400/10 dark:hover:text-gray-300;
+  }
+  /* Doubled selector + !important: icon sizing contract (see pagination chevrons). */
+  .pc-filters__chip-remove-icon.pc-filters__chip-remove-icon {
+    @apply w-4! h-4!;
+  }
+  .pc-filters__add-icon.pc-filters__add-icon {
+    @apply w-4! h-4! -ml-0.5;
+  }
+  .pc-filters__panel {
+    @apply w-60;
+  }
+  .pc-filters__editor {
+    @apply w-60;
+  }
+  .pc-filters__form.pc-filters__form {
+    @apply w-full;
+  }
+  .pc-filters__fields {
+    @apply flex max-h-72 flex-col gap-0.5 overflow-y-auto;
+  }
+  .pc-filters__fields--closed {
+    @apply hidden;
+  }
+  .pc-filters__field {
+    @apply flex w-full items-center rounded-md px-2 py-1.5 text-left text-sm text-gray-700 transition-colors duration-150 hover:bg-gray-100 focus:outline-hidden focus-visible:ring-2 focus-visible:ring-primary-500/50 dark:text-gray-200 dark:hover:bg-gray-400/10;
+  }
+  .pc-filters__none {
+    @apply px-2 py-1.5 text-sm text-gray-500 dark:text-gray-400;
+  }
+  /* The add panel's second step. Display-only toggling, so there is a complete
+     rest state at both ends and nothing for reduced motion to sit through. */
+  .pc-filters__step {
+    @apply hidden;
+  }
+  .pc-filters__step--open {
+    @apply flex flex-col gap-2;
+  }
+  .pc-filters__back {
+    @apply -mt-1 inline-flex items-center gap-1 self-start text-xs text-gray-500 transition-colors duration-150 hover:text-gray-700 focus:outline-hidden focus-visible:ring-2 focus-visible:ring-primary-500/50 dark:text-gray-400 dark:hover:text-gray-200;
+  }
+  /* Doubled selector + !important: icon sizing contract (see pagination chevrons). */
+  .pc-filters__back-icon.pc-filters__back-icon {
+    @apply w-3.5! h-3.5!;
+  }
+  .pc-filters__step-title {
+    @apply text-sm font-semibold text-gray-900 dark:text-gray-100;
+  }
+  .pc-filters__clear {
+    @apply rounded-md px-1 text-sm font-medium text-primary-600 hover:underline focus:outline-hidden focus-visible:ring-2 focus-visible:ring-primary-500/50 dark:text-primary-400;
+  }
+  @media (prefers-reduced-motion: reduce) {
+    .pc-filters__chip-body,
+    .pc-filters__chip-remove,
+    .pc-filters__field,
+    .pc-filters__back,
+    .pc-filters .pc-popover__panel[data-pc-menu] {
+      transition-duration: 1ms;
+    }
+  }
--- a/dev.exs
+++ b/dev.exs
@@ -284,6 +284,7 @@ defmodule Dev.PlaygroundLive do
         %{slug: "table", name: "Table", ready: true},
         %{slug: "data-table", name: "Data table", ready: true},
         %{slug: "data-table-link", name: "Data table · links", ready: true},
+        %{slug: "filters", name: "Filters", ready: true},
         %{slug: "chart", name: "Chart", ready: true},
         %{slug: "local-time", name: "Local time", ready: true}
       ]
@@ -333,6 +334,9 @@ defmodule Dev.PlaygroundLive do
   ]
 
   @slugs Enum.flat_map(@nav, fn g -> Enum.map(g.items, & &1.slug) end)
+  # the whitelist both filter-bar demos decode against - link mode's
+  # from_params and event mode's handle_op take the same list
+  @filters_fields [:name, :category, :price, :in_stock, :added_on]
   @locale_codes ~w(en fr de es pt)
   @playground_languages [
     %{locale: "en", flag: "🇬🇧", label: "English"},
@@ -772,6 +776,10 @@ defmodule Dev.PlaygroundLive do
        dt_hidden: [],
        dt_order: [],
        dt_refunded: [],
+       # the filter bar demo: one State shared with a data table (event mode),
+       # plus the registered field types the dials switch on and off
+       filters_state: PetalComponents.DataTable.State |> struct(page_size: 5) |> run_filters(),
+       filters_types: ~w(text select number_range boolean date_range),
        radio: %{
          style: "cards",
          variant: "outline",
@@ -881,6 +889,7 @@ defmodule Dev.PlaygroundLive do
       # mobile menu so picking a component reveals it immediately.
       |> assign(:nav_open, false)
       |> maybe_run_dt_link(params, uri)
+      |> maybe_run_filters_link(params, uri)
 
     {:noreply, maybe_start_progress_sim(socket)}
   end
@@ -906,6 +915,24 @@ defmodule Dev.PlaygroundLive do
   end
 
   defp maybe_run_dt_link(socket, _params, _uri), do: socket
+
+  # The filter page's third example: link mode, where the URL IS the filter
+  # state. Same decode as the data table's link page - the query string is
+  # the only place the filters live, so a shared or curled URL renders the
+  # right rows on first paint and the back button walks the history.
+  defp maybe_run_filters_link(%{assigns: %{active: "filters"}} = socket, params, uri) do
+    alias PetalComponents.DataTable.State
+
+    query = uri |> URI.parse() |> Map.get(:query) || ""
+    params = Map.merge(Plug.Conn.Query.decode(query), params)
+    state = State.from_params(params, fields: @filters_fields)
+
+    socket
+    |> assign(:filters_link, run_filters(state))
+    |> assign(:filters_query, query)
+  end
+
+  defp maybe_run_filters_link(socket, _params, _uri), do: socket
 
   # The progress flagship simulates a live upload while you watch. One
   # timer at a time; it dies quietly when you leave the page or take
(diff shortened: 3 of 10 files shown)


mplatts:
## Screenshots

**Filters applied** — chips present, predicates matching the data table's own column-filter triggers on shared state

Filters active, light
Filters active, dark

**At rest**

Filters, light
Filters, dark

### Verified independently

| Check | Result |
|---|---|
| `mix test` | 954 tests, 0 failures, 1 skipped (+41) |
| `npm test` | 157 passing, unchanged (no JS touched) |
| `mix credo` | zero new entries vs `main` (the `data_table.ex` complexity finding shifts 242 → 243) |
| new dependencies | none |
| slugs : clauses | 61 : 61 |
| **existing data-table tests** | **zero deletions, all pass unmodified** |

That last row is the one that matters. This PR refactors `data_table.ex` by −256 lines to extract `DataTable.FilterEditor` (+360), which is the largest change to the repo's most complex component in this batch. I checked specifically: `test/petal/data_table_test.exs` and `test/js/data_table.test.js` are untouched, no deletions, and the only test file added is `filters_test.exs`. Retaining the `pc-data-table__filter-*` classes so the hook, the `:has` rules and the existing suite all keep working is the right instinct.

Driving `DataTable.State` directly rather than inventing a parallel state model is also the right call — event mode posting byte-identical payloads means an existing event-mode handler needs zero new clauses.

## One correction: the new CSS is **not** inside an `@layer`

The PR describes the `pc-filters` section as appended "inside `@layer components`". It isn't. I checked the file directly:

- `assets/default.css` has exactly three `@layer components` blocks, opening at lines 102, 5028 and 6860.
- The one at 6860 **closes at line 7016**.
- `.pc-filters` first appears at **line ~7590**, in the unlayered tail alongside carousel and toast.

Unlayered rules beat consumer utilities passed through `class`, so `<.filters class="...">` overrides will silently lose to the component's own styles — the exact problem raised on #626, where the fix was to wrap new rules in their own `@layer components { }` block. Worth doing here too; it is a two-line change (open a layer before the section, close after) and nothing else needs to move.

Not a blocker, and it matches what the surrounding tail already does — but it is the difference between "overrides work" and "overrides silently don't".

## Deviations worth your eye

The author flagged one real API gap honestly: **`multi` ships `:in` only, not `:not_in`**, because both `State.handle_op/3` and the hook's `readFilter` hardcode `:in` whenever `values[]` is present — so a `not_in` picker would be a lie. Fixing it needs a `State` grammar change, which the issue lists as an explicit non-goal. Correctly left alone rather than half-shipped.

The field list also uses Tab rather than arrow keys, with `role="menu"` dropped so the pattern stays conformant as a labelled button group. A one-line follow-up if you want roving focus.

nhobes:
Polish pass done. Two scope decisions are yours to rule on when closing #619 (both flagged by the author, both verified real by the audit):

1. **Arrow-key navigation through the add-filter field list** - the brief's keyboard map asked for it; the PR ships Tab-only and flags the substitution. shadcn/reui arrow-navigate their equivalents, so the 'exceeds' bar argues for adding a roving-focus handler to the PetalDataTable hook's keydown (the author offered this). I left it out of the polish pass deliberately: it touches the shared table hook, which felt like your call to green-light.

2. **Multi fields are :in-only** - offering :not_in would silently become :in today (State.handle_op's is_list branch wins before filter_op is read). Either bless the :in-only subset when closing the issue, or open a follow-up for a :not_in path in State.handle_op/3 + the hook's readFilter. Don't ask for it inside this PR.

Also: the hero example shows link and event modes as separate sections rather than the brief's mode dial - fine in practice, but note it when closing the issue if you accept it.

PR #641, opened 2026-08-12
feat(sortable): drag-to-reorder lists and grids, keyboard-first

### LINKED ISSUE #620: Component: Sortable

## Summary

A net-new `<.sortable>` component: drag-to-reorder lists and grids for LiveView. The user drags (or keyboard-lifts) an item, the DOM reorders optimistically with animated gaps, and on drop the component pushes a single reorder event (`%{"id" => ..., "from" => ..., "to" => ...}`) that the app persists — LiveView remains the source of truth for order. Implemented as one `PetalSortable` hook plus a HEEx list component; no dnd-kit, no npm dependencies. v1 is a single container (vertical list or grid); cross-container drag (kanban) is an explicit follow-up built on top of this.

## References

- reui.io "Sortable" (8 examples, dnd-kit based) — study behaviour and variants only, never code: drag overlay treatment, handle vs whole-item drag, grid reorder, disabled items.
- dnd-kit's accessibility model is the a11y bar to meet: keyboard reorder (Space lifts, arrows move, Space drops, Escape cancels) with `aria-live` announcements.
- shadcn has no first-party sortable; reui is the primary reference.

## API sketch

New module `lib/petal_components/sortable.ex` (`use Phoenix.Component`), exported via `use PetalComponents` (check `lib/petal_components.ex` for name collisions first — `sortable` is free).

```elixir
attr :id, :string, required: true, doc: "DOM id; the hook mounts here"
attr :on_reorder, :string, required: true,
  doc: "event name pushed on drop with %{id, from, to} (string-key params)"
attr :handle, :boolean, default: false,
  doc: "when true only the grip handle initiates drag; when false the whole item does"
attr :disabled, :boolean, default: false, doc: "disables all drag and keyboard reorder"
attr :orientation, :string, default: "vertical", values: ["vertical", "grid"],
  doc: "vertical = single-column list; grid = wrapping grid, arrow keys move in 2D"
attr :rest, :global

slot :item, required: true do
  attr :id, :string, doc: "stable item identity, rendered as data-sortable-id (required)"
  attr :disabled, :boolean, doc: "this item cannot be dragged or lifted"
end
```

Follow the string-event convention used by `attr :on_select, :string` in `lib/petal_components/chat.ex` (not the `JS` struct convention of `alert.ex`/`accordion.ex`) — the reorder must reach the server as a pushed event so the app can persist it.

Rendered contract (what tests assert):

- Root: `pc-sortable` (+ `pc-sortable--grid` modifier), `phx-hook="PetalSortable"`, `data-on-reorder={@on_reorder}`, `data-orientation`, `data-handle` when handle mode, `data-disabled` when disabled. Root also carries `role="list"` when items are not natively `li`.
- Items: `pc-sortable__item`, `data-sortable-id={item id}`, `role="listitem"`, and in keyboard mode `tabindex`/`aria-roledescription="sortable"` + `aria-describedby` pointing at usage instructions (see Accessibility).
- Handle (when `handle`): `pc-sortable__handle` button with grip icon, `aria-label` derived from item content or an explicit label.
- One visually hidden `aria-live="polite"` region per sortable for announcements, hook-owned.

The `:item` slot lets the app render anything inside (todo rows, image cards). The component must also work when the item list is a **LiveView stream** — items carry the stream DOM ids, and reconciliation is by `data-sortable-id`, never by index. Test this explicitly.

### Server-truth model

Optimistic DOM order on drop, then push `on_reorder`. The app's `handle_event` reorders its assigns/stream and the next patch re-renders in the new order. Because the DOM already matches, the patch is a no-op visually; if the server rejects the move, the patch snaps items back — that is the correct failure behaviour, not a bug. In addition to the pushed event, dispatch a bubbling `petal:sortable` CustomEvent with the same detail (`{id, from, to}`) for dead views / JS interop, mirroring the `petal:otp-complete` precedent in `lib/petal_components/input_otp.ex`.

## Hook justification (maintainer sign-off)

CSS + `Phoenix.LiveView.JS` cannot do pointer-tracked drag, FLIP gap animation, or the keyboard lift state machine. A `PetalSortable` hook in `assets/js/petal_components.js` (registered alongside `PetalInputOTP`, `PetalPopover`, `PetalComboBox` in the hooks index) is justified under the existing precedent. Zero npm dependencies — pointer events, `setPointerCapture`, and manual FLIP transforms only.

## Variants & states to build

- [ ] Vertical list reorder (default)
- [ ] Grid reorder (`orientation="grid"`, wrapping rows, 2D arrow-key movement)
- [ ] Whole-item drag (default) and handle-only drag (`handle`)
- [ ] Drag overlay: lifted item follows the pointer with slight scale + shadow; the original slot renders as a placeholder; siblings animate into the gap via FLIP transforms (CSS `transition` on `transform`, no layout thrash)
- [ ] Touch: long-press (~250ms) lifts; a plain tap/scroll must not lift; `touch-action` managed so page scroll still works outside a lift
- [ ] Keyboard reorder mode (see Accessibility)
- [ ] Disabled container (`disabled`) and per-item `disabled` (item skipped by pointer and keyboard)
- [ ] Reduced motion: with `prefers-reduced-motion`, gaps reposition instantly (no FLIP transition), overlay drops scale/shadow animation; rest state fully legible
- [ ] Dark mode: overlay shadow/surface uses the floating-panel treatment (`bg-white dark:bg-gray-900`, real borders) — see popover/modal in `assets/default.css`

CSS: new `pc-sortable` section in `assets/default.css` (`pc-sortable`, `pc-sortable--grid`, `pc-sortable__item`, `pc-sortable__item--dragging`, `pc-sortable__item--placeholder`, `pc-sortable__handle`, `pc-sortable__overlay`). Radius on items via the `--pc-radius` token.

## Playground page

Nav entry + hero + dials, per house style (see any recent playground page in `dev.exs`).

Dials:
- `handle`: whole item / grip handle
- `orientation`: vertical / grid
- `disabled`: on / off

Scenarios (shadcn-grade, no lorem ipsum):
1. **Todo list with handles** — realistic tasks, grip handles, a per-item disabled ("locked") row; the LiveView `handle_event` persists order and displays it, proving server truth
2. **Image grid** — `orientation="grid"` gallery reorder with thumbnail cards
3. **Disabled state** — the same list with `disabled`, showing inert styling

The page should demonstrate keyboard reorder in the hero copy (tell the visitor to Tab to an item and press Space).

## Accessibility

There is no single WAI-ARIA pattern for drag-and-drop; the bar is the dnd-kit keyboard model plus `aria-live` announcements:

- Tab reaches each sortable item (or its handle in handle mode); item has `aria-roledescription="sortable"` and `aria-describedby` referencing visually hidden instructions ("Press Space to lift, arrows to move, Space to drop, Escape to cancel")
- **Space** lifts the focused item (announce: "Picked up X, position 2 of 7")
- **Arrow keys** move the lifted item one position (vertical: up/down; grid: all four) (announce: "Moved X to position 3 of 7")
- **Space** drops (announce: "Dropped X at position 3 of 7") — only now does the `on_reorder` event fire
- **Escape** cancels, restoring the original position and announcing the cancel
- Focus stays on the moved item after drop; `aria-live="polite"` region carries all announcements
- Drag overlay and placeholder are `aria-hidden="true"`; the accessible list order is always the real DOM order

## Tests

JS suite (`test/js/sortable.test.js`, alongside `combo_box.test.js` / `popover.test.js`; run via `npm test`):
- [ ] Position maths: hit-testing which index a given pointer position maps to, vertical and grid
- [ ] Keyboard state machine: lift → move → drop event payload correctness; Escape restores order and fires nothing; per-item and container disabled short-circuit lift
- [ ] Announcement strings ("Picked up / Moved / Dropped ... position N of M") written to the live region
- [ ] Drop pushes the `data-on-reorder` event with `{id, from, to}` and dispatches `petal:sortable`
- [ ] Long-press threshold: short touch does not lift

ComponentCase (`test/petal/sortable_test.exs`, LazyHTML assertions, pattern per `test/petal/input_otp_test.exs`):
- [ ] Root renders `phx-hook="PetalSortable"`, `data-on-reorder`, `data-orientation`, list role
- [ ] Every item renders `data-sortable-id` and the a11y attributes (`aria-roledescription`, `aria-describedby`, tabindex)
- [ ] `handle` renders the grip button with an accessible name; absent otherwise
- [ ] `disabled` (container and per-item) renders the data attrs and removes items from the tab order
- [ ] `orientation="grid"` adds the modifier class
- [ ] Live region present with `aria-live="polite"`
- [ ] `rest` passthrough and `class` merging

Plus a live test (see `test/petal/a11y_live_test.exs` for the harness) rendering the sortable from a **LiveView stream**, asserting that a simulated reorder event plus stream re-insertion reconciles by `data-sortable-id` without duplication.

## Non-goals

- Cross-container drag / kanban boards — the explicit follow-up; v1's single-container contract and event shape are designed so a `group` attr and container-to-container transfer can be added without breaking changes
- Multi-select drag
- Auto-scrolling scroll containers during drag
- Nested sortables
- Client-side persistence of order (server truth only)

Document each in the moduledoc as planned follow-ups toward a kanban composition.

## Process

1. Build on a feature branch: `lib/petal_components/sortable.ex` + `pc-sortable` section in `assets/default.css` + `PetalSortable` hook in `assets/js/petal_components.js` + tests + playground page (nav entry, dials, examples). No external packages.
2. Verify in the playground (`mix run dev.exs`, http://localhost:4000) — exercise every dial, light + dark, keyboard-only reorder, touch on a real device or emulation, and `prefers-reduced-motion`.
4. Open a PR against `main` with screenshots (light + dark) of the playground page, ideally including a mid-drag shot. The automated review must pass.

## Definition of done

- `<.sortable>` component with the API above; every attr has a `doc:` string; moduledoc includes usage examples (basic list, handle mode, grid, stream-backed list with the `handle_event` snippet)
- `pc-sortable` CSS section, `--pc-radius` respected, dark mode and reduced motion covered
- `PetalSortable` hook registered in the hooks index; keyboard + pointer + touch paths all working
- JS suite, ComponentCase suite, and stream reconciliation live test green
- Playground page with dials and the three scenarios
- PR screenshots (light + dark) attached

mplatts:
Closes #620

## Summary

`<.sortable>` plus one `PetalSortable` hook: drag-to-reorder lists and grids for LiveView. Drag a row (or only its grip with `handle`) and the list reorders under the pointer with the gaps animating into place; on drop it pushes ONE `on_reorder` event carrying `%{"id", "from", "to"}` so the app persists the order. Zero npm dependencies - pointer events, `setPointerCapture` and hand-rolled FLIP transforms.

Five parts per CONTRIBUTING: module, `pc-sortable` CSS section, both test suites, showcase module + registry entry, playground page. Plus `### Unreleased` in the changelog.

All the order maths and the whole keyboard lift state machine live in an exported `SortableCore` as pure functions, so they unit-test without a DOM. The hook is only the layer that reads rects, moves nodes and talks to LiveView.

## Hook justification

CSS and `Phoenix.LiveView.JS` cannot do pointer-tracked drag, FLIP gap animation, or a keyboard lift state machine. Maintainer signed this off on the issue. The hook is kept as thin as I could make it: every decision that produces an *order* is a pure function outside it.

## How LiveView reconciliation was resolved - who owns order after a drop

**The server owns order. The client owns it only for the instant between the drop and the patch.**

The DOM reorders optimistically as you drag, so on drop it already shows the intended order. Then the event goes up, the app persists, and the patch re-renders in that same order - a visual no-op. If the server disagrees, the patch is what corrects the DOM, which is the right failure behaviour rather than a bug.

Two things fell out of building it that are documented loudly in the moduledoc:

1. **Streams do not snap back on their own.** Stream updates are ops, not a diff. A handler that rejects a move and sends no stream ops leaves the DOM in the client's order forever, silently disagreeing with the server. The "Rejecting a move" section spells this out with a `stream(..., reset: true)` example. This is the one thing that would bite a real app and that no unit test would catch.
2. **`from` can be stale, the id never is.** Under concurrency someone else may have reordered between the render and the drop. The docs tell you to treat the payload as "move `id` to index `to`" and use `from` as a sanity check only.

Mid-drag patches: because there is no cloned overlay (see deviations), a patch can replace the nodes being animated. The hook abandons that drag rather than animating a detached tree. A keyboard lift is cheaper and survives - `updated()` re-resolves it by `data-sortable-id`, never by the remembered index. Both documented.

One bug this actually surfaced in the browser: the live region is written by the hook and rendered empty by the server, so the drop's own patch wiped the announcement back out. It now carries `phx-update="ignore"`, with a test pinning that.

## Keyboard map

Tab reaches each item, or its grip in handle mode.

| Key | Does |
|---|---|
| `Space` | lift the focused item - "Picked up X, position 2 of 7" |
| `Arrow` keys | move it one position; up/down in a list, all four in a grid - "Moved X to position 3 of 7" |
| `Space` | drop - "Dropped X at position 3 of 7". Only now does `on_reorder` fire, and not at all if it landed where it started |
| `Escape` | cancel, restore the original order, announce it, push nothing |

Focus follows the item after a drop (the grip in handle mode). Announcements go to a visually hidden `aria-live="polite"` region, pointer drops included, not just keyboard ones. Items carry `aria-roledescription="sortable"` and `aria-describedby` pointing at hidden instructions. Disabled items and disabled containers leave the tab order.

1. **Added `dom_id` to the `:item` slot.** Streams need the stream dom id on the element while the server wants the record id in the payload. Two ids, two attrs; `dom_id` defaults to `id`. The sketch omitted it and stream usage does not work without it.
2. **No `pc-sortable__overlay` / `pc-sortable__item--placeholder`.** The issue's class list implies dnd-kit's model - original stays as a faded placeholder, a clone follows the pointer. I used SortableJS's instead: the real element is moved through the list and gets a transform plus scale and shadow. The issue also says "the accessible list order is always the real DOM order", which the clone model fights; a clone additionally breaks form state, media and measured widths inside a row, and doubles the a11y surface for no gain. The classes shipped are `pc-sortable`, `--grid`, `--disabled`, `--dragging`, `__item`, `__item--dragging`, `__item--lifted`, `__item--disabled`, `__handle`, `__grip`, `__sr`.
3. **The live region and the instructions are siblings of the list, not children.** `phx-update="stream"` requires the container's children to be exactly the stream items, so a nested live region would break every stream user. `aria-describedby` works across the document. Both are `sr-only`, so absolutely positioned and invisible to a flex or grid parent.
4. **Grid hit-testing does not use reading-order midpoints.** A list uses midpoints (hysteresis matters when rows have different heights). A grid uses the cell under the pointer, because the same midpoint rule sends an item somewhere it was never dragged - hover the bottom-left cell from its left half and reading-order counting lands the item top-right. This is safe because the list is reordered live, so the rects are the current arrangement and the item lands where the pointer already is. Falls back to counting in the gutters and past the end. My own test expectations caught this; both models are covered by specs.
5. **Added a `target` attr** rendering `phx-target`, so a sortable inside a LiveComponent can own its own event.
6. **Added `label` and `class` to the `:item` slot.** `label` is the accessible name used in the handle's `aria-label` and in every announcement, rendered as `data-sortable-label` for the hook; it falls back to the item's text.
7. **Per-item `disabled` means "cannot be grabbed", not "cannot move".** A disabled row is still displaced when its neighbours move around it. Fixed-anchor semantics is a different feature (pinned rows). Called out in the attr doc since "locked row" invites the other reading.
8. **Auto-scroll is not implemented** - the issue lists it under non-goals. It is in the moduledoc's planned follow-ups alongside cross-container drag, multi-select and nested sortables.

## What I verified interactively vs by unit test

Driven in a real browser on the playground page (`/c/sortable`), light and dark:

- **Keyboard reorder end to end.** Focused a grip, Space, two ArrowDowns, Space. The list moved, the live region read "Dropped Cut the 4.1 release build at position 2 of 5", and the server log showed `Cut the 4.1 release build: 1 -> 2`. Nothing was pushed until the drop - the log still read "nothing moved yet" mid-lift, which is the `lifted.png` screenshot.
- **Escape.** Lifted, moved twice, Escape: order restored, "Reorder cancelled. Prep Monday standup returned to position 5 of 5", server log unchanged.
- **Grid 2D arrows.** In a 3-column grid, ArrowDown moved position 1 to position 4 (a full row), ArrowRight to 5, drop persisted as `Harbour at dawn: 1 -> 5`.
- **A real pointer drag.** Dragged a grip down the list; the row moved and the server logged `Cut the 4.1 release build: 1 -> 4`. That event can only have come from the hook, so the pointer path is genuinely exercised, not just asserted.
- **The `phx-update="ignore"` bug**, found by reading the live region after a drop and seeing it empty.

Unit-tested rather than driven: touch long-press thresholds (fake timers), pointer slop, `petal:sortable` dispatch, `pushEventTo` when `phx-target` is set, the patch-mid-lift re-resolution, and all the pure maths. **Not tested on a real touch device** - the touch path is jsdom specs plus the standard `touch-action` / non-passive `touchmove` / `setPointerCapture` combination. That deserves a phone before release.

## Tests

| Suite | Before | After |
|---|---|---|
| `mix test` | 913, 1 skipped | **928**, 1 skipped, 0 failures |
| `npm test` | 157 | **214**, 0 failures |

`mix credo` vs `origin/main`: **zero new entries** (diffed the full sorted output both ways).

15 ComponentCase tests cover the container attrs, item a11y attrs, `dom_id`, tab order, handle mode, both disabled levels, the grid modifier, `class` merging, `rest` passthrough, the live region (including `phx-update="ignore"`), and a stream-backed render asserting a reordered stream keeps every id exactly once and reconciles by `data-sortable-id` rather than position.

57 JS tests cover `arrayMove`, vertical and grid hit-testing, arrow stepping and clamping, the announcement strings, the full lift/move/drop/cancel state machine including both disabled short-circuits and the no-op drop, plus hook wiring: which keys are swallowed, what reaches the server, focus after a drop, the long-press threshold, and teardown.

## Not done

- A LiveView `live_isolated` stream test. `test/petal/a11y_live_test.exs` is a skipped stub, not a harness, and there is no endpoint wired for live tests in this repo. Stream reconciliation is covered at the render level (ids survive a reorder, once each, keyed by `data-sortable-id`) and the playground proves the round trip against a real LiveView. Happy to wire a live test if you want one, but it needs test infrastructure this repo does not have yet.
- Real-device touch, as above.

## Screenshots

Light, dark and a keyboard lift mid-move (item at position 3, server log still "nothing moved yet") are attached below.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,38 @@
 # Changelog
+
+### Unreleased
+
+#### Added
+
+- **New `sortable` component: drag-to-reorder lists and grids, with a
+  keyboard path that is not an afterthought.** Drag a row (or only its
+  grip, with `handle`) and the list reorders under your finger with the
+  gaps animating into place; on drop it pushes ONE `on_reorder` event
+  carrying `%{"id", "from", "to"}` so your `handle_event` can persist the
+  order. LiveView stays the source of truth: the DOM has already moved,
+  so the patch is a visual no-op, and a server that says no snaps the
+  items back. `orientation="grid"` wraps items into a grid and teaches
+  the arrow keys to move in two dimensions, reading the column count off
+  the live layout so a responsive grid steps correctly at every
+  breakpoint. Works with LiveView streams - pass `phx-update="stream"`
+  through and give items a `dom_id`; reconciliation is by
+  `data-sortable-id`, never by index.
+
+  Keyboard reorder is a first-class path, not a fallback: Space lifts,
+  the arrows move, Space drops, Escape cancels, and every transition is
+  announced through a visually hidden `aria-live` region ("Picked up
+  Discovery call, position 1 of 5"). Focus follows the item it moved.
+  On touch a long press lifts, so a tap or a scroll still does what it
+  always did.
+
+  One `PetalSortable` hook, zero npm dependencies - pointer events,
+  `setPointerCapture` and hand-rolled FLIP transforms. All the order
+  maths and the whole lift state machine are pure functions, so they are
+  unit-tested without a DOM. There is no cloned drag overlay: the real
+  element moves through the list, which keeps the accessible order equal
+  to the DOM order at every instant and keeps form state and media
+  inside a row intact.
+
 ### 4.14.0 - 2026-08-11
 
 #### Added
--- a/assets/default.css
+++ b/assets/default.css
@@ -7586,3 +7586,109 @@
       transition-duration: 1ms;
     }
   }
+
+/* Sortable - drag-to-reorder lists and grids.
+ * The dragged element IS the overlay: it is really moved through the list as
+ * you drag, so there is no clone and the accessible order is always the DOM
+ * order. Siblings FLIP into the gap; the transform transition below is the
+ * whole animation, which is why reduced motion is one media query at the
+ * bottom rather than a branch in the hook. */
+@layer components {
+  .pc-sortable {
+    @apply flex flex-col gap-2;
+  }
+
+  .pc-sortable--grid {
+    @apply grid grid-cols-2 gap-3 sm:grid-cols-3;
+  }
+
+  .pc-sortable__item {
+    @apply relative flex items-center gap-3 px-3 py-2.5 bg-white border border-gray-200 text-gray-900 shadow-xs;
+    @apply dark:bg-gray-400/8 dark:border-gray-400/20 dark:text-gray-100;
+    @apply focus:outline-hidden focus-visible:ring-2 focus-visible:ring-primary-500/50 focus-visible:border-primary-500;
+    /* manipulation, not none: the page must still scroll under a list that
+       happens to be sortable. A lift vetoes the scroll itself, from the
+       hook's non-passive touchmove, once a long press has earned it. */
+    touch-action: manipulation;
+    transition:
+      transform 200ms cubic-bezier(0.2, 0, 0, 1),
+      box-shadow 150ms ease-out;
+    border-radius: var(--pc-radius, 0.625rem);
+  }
+
+  /* Whole-item drag: the row itself is the grab target. In handle mode the
+     grip owns the cursor instead, so the row stays an ordinary row. */
+  .pc-sortable:not([data-handle="true"]):not([data-disabled="true"])
+    .pc-sortable__item:not([data-disabled="true"]) {
+    @apply cursor-grab;
+  }
+
+  /* The lifted item, pointer or keyboard: the floating-panel treatment used
+     by popover and modal, so it reads as being above the list. */
+  .pc-sortable__item--dragging,
+  .pc-sortable__item--lifted {
+    @apply z-30 bg-white shadow-lg border-primary-500 dark:bg-gray-900 dark:border-primary-500/70;
+  }
+
+  .pc-sortable__item--dragging {
+    /* pointer-driven, so a transition here would lag the finger */
+    @apply cursor-grabbing select-none;
+    transition: none;
+    scale: 1.02;
+  }
+
+  .pc-sortable__item--lifted {
+    @apply ring-2 ring-primary-500/50;
+  }
+
+  /* Nothing under the pointer should paint a hover or select text mid-drag. */
+  .pc-sortable--dragging {
+    @apply select-none;
+  }
+
+  .pc-sortable--dragging .pc-sortable__item:not(.pc-sortable__item--dragging) {
+    @apply pointer-events-none;
+  }
+
+  .pc-sortable__item--disabled {
+    @apply opacity-60 cursor-not-allowed bg-gray-50 dark:bg-gray-400/4;
+  }
+
+  .pc-sortable__handle {
+    @apply flex shrink-0 items-center justify-center -ml-1 p-1 cursor-grab text-gray-400 transition-colors;
+    @apply hover:text-gray-600 dark:text-gray-500 dark:hover:text-gray-300;
+    @apply focus:outline-hidden focus-visible:ring-2 focus-visible:ring-primary-500/50;
+    /* the grip is the one surface that trades page scroll for dragging */
+    touch-action: none;
+    border-radius: max(calc(var(--pc-radius, 0.625rem) - 0.25rem), 0.25rem);
+  }
+
+  .pc-sortable__handle[aria-disabled="true"] {
+    @apply cursor-not-allowed opacity-50 hover:text-gray-400 dark:hover:text-gray-500;
+  }
+
+  .pc-sortable--dragging .pc-sortable__handle {
+    @apply cursor-grabbing;
+  }
+
+  .pc-sortable__grip {
+    @apply w-4 h-4;
+  }
+
+  /* The keyboard instructions and the announcement region: present for
+     assistive tech, absent from layout (sr-only is absolutely positioned, so
+     a flex or grid parent never sees them as children). */
+  .pc-sortable__sr {
+    @apply sr-only;
+  }
+
+  @media (prefers-reduced-motion: reduce) {
+    .pc-sortable__item {
+      transition: none;
+    }
+
+    .pc-sortable__item--dragging {
+      scale: 1;
+    }
+  }
+}
--- a/assets/js/petal_components.js
+++ b/assets/js/petal_components.js
@@ -4925,6 +4925,691 @@ export const PetalComboBox = {
   },
 };
 
+// Sortable: every order decision as a pure function.
+//
+// Drag-to-reorder is one of the few things CSS and Phoenix.LiveView.JS
+// genuinely cannot do - pointer tracking, FLIP gap animation and a keyboard
+// lift state machine all need imperative code. So the hook exists, but the
+// thinking does not live in it: the reorder maths, the arrow-key stepping,
+// the announcement strings and the whole lift state machine are pure
+// functions here, unit-tested in test/js/sortable.test.js without a DOM. The
+// hook below is only the layer that reads rects, moves nodes and talks to
+// LiveView.
+export const SortableCore = {
+  // A tap is not a drag and a scroll is not a lift. Touch waits for a long
+  // press; mouse and pen wait for real movement so a plain click still is one.
+  LONG_PRESS_MS: 250,
+  TOUCH_SLOP_PX: 8,
+  POINTER_SLOP_PX: 4,
+
+  arrayMove(list, from, to) {
+    const next = list.slice();
+    if (from < 0 || from >= next.length) return next;
+    const clamped = Math.max(0, Math.min(to, next.length - 1));
+    const [moved] = next.splice(from, 1);
+    next.splice(clamped, 0, moved);
+    return next;
+  },
+
+  // Where a pointer at (x, y) wants the dragged item, given every item's rect
+  // in current DOM order.
+  //
+  // The two orientations genuinely want different rules. A list has rows of
+  // varying height, so it uses midpoints: an item is displaced only once the
+  // pointer crosses its centre line, which is the hysteresis that stops two
+  // unequal rows flickering back and forth across one boundary. A grid has
+  // uniform cells and reflows in two dimensions, where the same rule sends an
+  // item somewhere it was never dragged (hover the bottom-left cell from the
+  // left and reading-order counting lands you top-right). So a grid uses the
+  // cell under the pointer directly. That works because the list is reordered
+  // live as you drag: the rects ARE the current arrangement, so the cell you
+  // are over is exactly the slot you are asking for, and once the item lands
+  // there the pointer is still inside it - stable, not oscillating.
+  indexFromPoint({ rects, x, y, orientation = "vertical", activeIndex = -1 }) {
+    if (orientation === "grid") {
+      const over = rects.findIndex(
+        (r) =>
+          x >= r.left && x <= r.left + r.width && y >= r.top && y <= r.top + r.height,
+      );
+
+      // Only fall through to counting when the pointer is in a gutter or off
+      // the end of the grid, where no cell can answer.
+      if (over !== -1) return over;
+    }
+
+    let passed = 0;
+
+    rects.forEach((rect, i) => {
+      if (i === activeIndex) return;
+      if (SortableCore.pointerPassed(rect, x, y, orientation)) passed += 1;
+    });
+
+    return Math.max(0, Math.min(passed, rects.length - 1));
+  },
+
+  pointerPassed(rect, x, y, orientation) {
+    if (orientation === "grid") {
+      // Reading order: a whole row above the pointer is behind it, a whole
+      // row below is ahead, and within its own row it is a left/right call.
+      if (y > rect.top + rect.height) return true;
+      if (y < rect.top) return false;
+      return x > rect.left + rect.width / 2;
+    }
+
+    return y > rect.top + rect.height / 2;
+  },
+
+  // One arrow-key step, clamped to the list. Returns null for a key this
+  // component does not own, so the hook leaves that event alone.
+  nextKeyboardIndex({
+    index,
+    key,
+    orientation = "vertical",
+    columns = 1,
+    count,
+  }) {
+    const span = Math.max(1, columns);
+
+    const step =
+      orientation === "grid"
+        ? { ArrowLeft: -1, ArrowRight: 1, ArrowUp: -span, ArrowDown: span }[key]
+        : { ArrowUp: -1, ArrowDown: 1 }[key];
+
+    if (step === undefined) return null;
+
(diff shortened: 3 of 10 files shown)


mplatts:
## Screenshots

**Item lifted** with the drop indicator

Sortable, item lifted

**Light / dark**

Sortable, light
Sortable, dark

### Verified independently

| Check | Result |
|---|---|
| `mix test` | 928 tests, 0 failures, 1 skipped (+15) |
| `npm test` | 214 passing (**+57**) |
| new dependencies | none — no SortableJS, no dragula |
| slugs : clauses | 61 : 61 |

## The reconciliation notes are the valuable part

Server owns order; the client owns it only between drop and patch. Two consequences documented loudly, and both are the kind of thing that bites in production rather than in review:

- **Streams don't snap back on rejection.** They're ops, not a diff — so if the server rejects a reorder, the app has to explicitly re-insert. Silent divergence otherwise.
- **`from` can be stale under concurrency while the id can't**, so handlers should place by id.

Grid hit-testing using the hovered cell rather than reading-order midpoints came out of the author's own test expectations catching that midpoints send a bottom-left drop to the top-right. That's a test earning its keep.

A **real pointer drag** was driven successfully here (row moved, server logged `1 -> 4`), so the `position: fixed` coordinate problem seen on #637 is situational rather than universal. It also drove out a real bug: the live region was being wiped by the drop's own patch, now fixed with `phx-update="ignore"` and a test.

## Two honest gaps, both stated in the PR

- **No `live_isolated` stream test** — `a11y_live_test.exs` is a skipped stub rather than a harness and there's no test endpoint in this repo. Covered at render level plus the playground round trip instead. If you want real stream-reconciliation coverage, that harness is the prerequisite and it's a separate piece of work.
- **Touch not verified on a real device** — jsdom specs plus the standard `touch-action` / non-passive `touchmove` / `setPointerCapture` combination. Worth a phone before release.

Auto-scroll was skipped: my brief mentioned it, but the issue lists it as a non-goal. The issue won, correctly.

nhobes:
Grid reorder bug found during the omnibus eye-test, fixed in the new commit:

**Symptoms** (playground grid example): dragging the first item never displaced anything; dragging a later item to the front worked, but the drag then became one-way — the item couldn't walk back to its original slot mid-drag and was forced to drop into first place. Vertical lists were always fine.

**Root cause**: `SortableCore.indexFromPoint`'s grid branch searched every rect for the one under the pointer — including the dragged item's own rect, which carries the pointer-tracking transform and therefore sits under the pointer *by construction*. `findIndex` returns the first DOM-order match, so the answer was always `min(draggedIndex, targetIndex)`: index 0 shadowed the whole grid, and any item that reached the front inherited the same shadow.

**Fix**: skip `activeIndex` in the cell search (the vertical counting path already did). The pointer then falls through to the existing reading-order counting when it's over the slot the item owns — stable and bidirectional.

**Evidence**: verified in a real Chrome session over CDP — both directions reorder, and an item walks back to its original slot mid-drag; vertical drag sequence byte-identical before/after. `mix test` 931/0 (exit 0), vitest 226 passed with 4 new specs — 3 of which fail with the fix reverted (the old pure-grid specs modelled a static grid with the dragged rect still home, the one state that can't occur mid-drag).

nhobes:
Second eye-test follow-up — the "intense jittering" while other items shuffle was diagnosed, and it was a real logic bug feeding itself, not just missing polish:

**Oscillation (primary)**: the hit test read every item's `getBoundingClientRect` per pointermove — but mid-drag those are *paint* positions, part-way through the 200ms FLIP, so several travelling rects can cover the pointer at once. Each wrong answer reorders; each reorder restarts the animations; the grid shakes itself. Held perfectly still over one cell, the order flipped on **29 of 30 consecutive frames**, cycling three arrangements.

**Fix: hit-test slots, not paint.** Slot geometry (layout positions, transforms stripped) is measured at the lift and at every reorder, cached container-relative for the drag, and re-anchored to the container's live rect per move — so mid-drag page scroll still resolves correctly (verified with a 121px scroll). Bonus: one rect read per move instead of one per item. `SortableCore`'s reorder maths untouched — the bidirectional fix's 4 specs stand.

**FLIP interruption (secondary)**: the invert measured the post-reorder rect with the previous animation's transform still applied — interrupted reorders snapped by whatever was left in flight (single frames of 430px). Transforms now come off before measuring, so a mid-animation reorder retargets from the current visual position.

**Reduced motion**: items still land in their slots, they just arrive instead of gliding — the CSS media query stays the primary gate, the hook now skips the invert work too (matchMedia optional, so consumer jsdom suites keep working).

**Measured, same slow grid drag**: 58 order changes → **4**; 74 direction reversals → **2** (both the dragged item tracking the pointer); biggest single-frame move 430px → **73px** (one ease-out frame over a row wrap). Held pointer: 0 flips. Fast flick: no pile-up, no stuck transforms 600ms after drop. Vertical list byte-identical metrics before/after.

Gates: vitest 231 (exit 0, +5 specs, 4 fail against the old code), `mix test` 931/0, format clean.

PR #642, opened 2026-08-12
feat(calendar): month grid and date picker on plain Elixir Date

### LINKED ISSUE #610: Component: Calendar & Date Picker
## Summary

A two-tier date selection system: `<.calendar>`, a pure server-rendered month grid built on Elixir `Date`, and `<.date_picker>`, an input-plus-calendar composition that puts the grid in a popover with a formatted, editable text input. The calendar works with or without LiveView (month navigation via events or plain links with query params), supports single/range/multiple selection, min/max and disabled dates, and follows the WAI-ARIA grid pattern for keyboard use. This is the flagship component of the release and the most-requested gap in the library; every serious form eventually needs one, and there is currently no native HEEx answer.

Budget accordingly: this is a week-plus build, not days.

## References

Study behaviour and variants only; we build native HEEx/LiveView from scratch.

- reui.io Calendar (30 examples) and Date Selector (4 examples) - the visual and scope reference
- react-day-picker (https://daypicker.dev) - the behaviour reference: selection modes, disabled matchers, outside days, fixed weeks, min/max
- shadcn/ui Calendar + Date Picker - the popover composition pattern
- WAI-ARIA APG grid pattern and Date Picker Dialog example - the a11y contract

In this repo, study before starting:

- `lib/petal_components/popover.ex` - the popover machinery `<.date_picker>` composes (anchored and `top_layer` modes, Escape/refocus handling, `compose_js/2`)
- `lib/petal_components/field.ex` and `lib/petal_components/input.ex` - field surface, `Phoenix.HTML.FormField` integration, `clearable` precedent, label/errors/help_text wiring
- `lib/petal_components/input_otp.ex` and `assets/js/petal_components.js` - the precedent for when a hook is justified and how hooks are registered
- `lib/petal_components.ex` - confirm no name collisions for `calendar/1` and `date_picker/1`
- `assets/default.css` - the `pc-*` section pattern new components add to

## Design decisions (locked)

- **Elixir `Date` throughout.** No time selection, no timezones, no DateTime. Values in and out are `%Date{}`; form posting uses ISO 8601 strings.
- **No cldr or any date-localisation hex dep.** Day and month names are plain attrs with English defaults; the docs show gettext wiring as a recipe. Zero new required deps.
- **Works without LiveView.** Month navigation renders as plain links carrying query params when no event handler is wired; document both wirings (event-based and link/query-param based) in the moduledoc.
- **Keyboard navigation is a small hook**, not server round-trips per arrow key. CSS + LiveView.JS cannot implement roving tabindex across a month grid; this meets the bar set by `PetalInputOTP` and `PetalPopover`. The grid must remain fully usable without the hook (every day is a real button; only roving-focus arrow keys degrade).

## API sketch

Starting proposal; refine and flag deviations in the PR.

### `<.calendar>` - `lib/petal_components/calendar.ex`

```elixir
attr :id, :string, doc: "autogenerated if not set"
attr :mode, :string, default: "single", values: ~w(single range multiple),
  doc: "selection behaviour"
attr :value, :any, default: nil,
  doc: "the selection: a Date (single), a {from, to} tuple or map with :from/:to (range, either side may be nil mid-selection), or a list of Dates (multiple)"
attr :month, :any, default: nil,
  doc: "the displayed month as any Date within it; defaults to the selected value's month, else today"
attr :min, :any, default: nil, doc: "earliest selectable Date (inclusive)"
attr :max, :any, default: nil, doc: "latest selectable Date (inclusive)"
attr :disabled_dates, :any, default: nil,
  doc: "a list of Dates, or a 1-arity function Date -> boolean; disabled days render but are not selectable"
attr :starts_on, :integer, default: 1, values: [1, 2, 3, 4, 5, 6, 7],
  doc: "first day of the week, ISO day numbers (1 = Monday, 7 = Sunday)"
attr :show_outside_days, :boolean, default: true,
  doc: "render leading/trailing days from adjacent months"
attr :fixed_weeks, :boolean, default: false,
  doc: "always render 6 week rows so the grid height never jumps"
attr :day_names, :list, doc: "7 short day labels ordered Monday-first; default English"
attr :month_names, :list, doc: "12 month labels; default English"
attr :on_select, :any, default: nil,
  doc: "event name pushed when a day is clicked (LiveView wiring); payload includes the ISO date"
attr :on_month_change, :any, default: nil,
  doc: "event name for prev/next month; when nil, nav renders as patch links using the month_param query param"
attr :month_param, :string, default: "month",
  doc: "query param used for link-based month navigation"
attr :rest, :global
```

Notes:

- With `on_select`/`on_month_change` set, day buttons and nav buttons use `phx-click` with `phx-value-date`/`phx-value-month` (ISO strings). With them nil, nav renders `<.link patch={...}>` carrying `month_param`, and days render as submit-capable buttons inside whatever form wraps the calendar. Both wirings get a full moduledoc example.
- Month/day maths is pure `Date` arithmetic in private helpers (`Date.beginning_of_month/1`, `Date.add/2`, `Date.day_of_week/2` with `starts_on`). No sigils, no external libs.
- Range mode with only `:from` set renders the anchored day as selected and hover-preview via CSS on the rest (preview is cosmetic; the real state lives server-side).

### `<.date_picker>` - `lib/petal_components/date_picker.ex`

Composes `<.calendar>` inside the existing popover machinery with a field-surface input.

```elixir
attr :id, :string, doc: "autogenerated if not set"
attr :field, Phoenix.HTML.FormField, default: nil,
  doc: "form field; wires name, value, errors like <.field>"
attr :name, :any
attr :value, :any, default: nil
attr :label, :string, default: nil
attr :mode, :string, default: "single", values: ~w(single range)
attr :format, :string, default: "%Y-%m-%d",
  doc: "strftime format string for the display input (Calendar.strftime/2); the posted value is always ISO 8601"
attr :placeholder, :string, default: nil
attr :clearable, :boolean, default: false, doc: "show a clear button when a value is set"
attr :two_months, :boolean, default: false,
  doc: "range mode only: render two month panes side by side"
# pass-through calendar attrs: min, max, disabled_dates, starts_on,
# show_outside_days, fixed_weeks, day_names, month_names
attr :on_select, :any, default: nil
attr :rest, :global
```

Notes:

- The visible input shows the formatted date and allows manual typing. Parse-on-blur contract: on blur (or Enter), attempt `Date.from_iso8601/1` and then a lenient parse of the configured format; on success update the value and calendar month, on failure revert the input to the last valid formatted value. Document this contract explicitly in the moduledoc.
- The posted form value is a hidden input carrying ISO 8601 (`YYYY-MM-DD`; range posts `from`/`to` sub-fields), regardless of display `format`.
- Reuse the popover's open/close, Escape-and-refocus, and click-away behaviour rather than reimplementing; on open, focus moves into the grid (selected day, else today, else first enabled day) and returns to the input on close.

## Variants & states to build

- Calendar modes: `single`, `range` (including half-selected mid-range state and hover preview), `multiple`
- Month navigation: event-based (LiveView) and link/query-param based (dead views) - both demonstrated
- `min`/`max` window enforcement, `disabled_dates` as list and as function
- `starts_on` for all 7 values (visually verify at least Monday and Sunday)
- `show_outside_days` on/off; `fixed_weeks` on/off
- Today marker (distinct from selected; both can coincide)
- Date picker: closed/open, empty/filled, clearable, disabled, with-errors (field integration), two-month range pane
- Dark mode for everything; floating-panel surface per popover/modal, field surface per input; `--pc-radius` token per the house radius rules

## Playground page

Dials (segmented controls via `handle_event`, per the house pattern in `dev.exs`):

- `mode` (single / range / multiple)
- `starts_on` (Monday / Sunday at minimum)
- `show_outside_days` toggle
- min/max window toggle (e.g. "next 30 days only")
- two-month range toggle (picker section)

1. **Booking range** - a check-in/check-out range picker with `min` = today and a two-month pane
2. **Birthday picker** - single mode with a year-select header variant for fast decade jumps (if the year select is cut for scope, note it in Non-goals follow-ups rather than half-shipping)
3. **Deadline picker in a form** - `<.date_picker>` wired to a `Phoenix.HTML.FormField` inside a form with validation errors, demonstrating ISO posting

The page must also demonstrate keyboard-only operation.

## Accessibility

Follow the WAI-ARIA grid pattern (APG Date Picker Dialog example):

- The month grid is `role="grid"` with `role="row"`/`role="gridcell"`, column headers for day names, and an accessible name including month + year
- Roving tabindex: exactly one day in the grid is tabbable at a time
- Keyboard map: Arrow keys move by day/week; PageUp/PageDown move by month (Shift+PageUp/Down by year is a nice-to-have); Home/End jump to start/end of the week; Enter/Space selects
- `aria-selected="true"` on selected days; disabled days get `aria-disabled="true"` and remain focusable per APG guidance
- Month changes announced via an `aria-live="polite"` region
- Picker: the input labels the popup (`aria-haspopup="dialog"`, `aria-expanded`), Escape closes and restores focus to the input, focus moves into the grid on open
- Grid remains operable without the hook: every day is a native `<button>`, so Tab + Enter always works

## Tests

`test/petal/calendar_test.exs` and `test/petal/date_picker_test.exs` using `ComponentCase` + LazyHTML:

- Month maths edge cases: December to January navigation (year rollover), February in a leap year (2024) vs non-leap, `fixed_weeks` always yielding 6 rows
- `starts_on` reorders the header and grid correctly
- Selection modes: correct `aria-selected` for single, range (from/to and every day between), multiple
- Half-open range (`from` set, `to` nil) renders without crashing
- `min`/`max` and `disabled_dates` (list and function forms): disabled days carry `aria-disabled` and no selection wiring
- `show_outside_days` on/off DOM assertions; outside days marked distinctly
- Today marker present on today's cell
- Link-based nav renders patch links carrying `month_param`; event-based nav renders `phx-click` with ISO `phx-value-month`
- Keyboard map: hook attributes/data present on the grid; roving tabindex initial state (exactly one `tabindex="0"` day)
- Picker: hidden input posts ISO regardless of `format`; display input shows the strftime-formatted value; range posts from/to sub-fields; clearable button renders only when a value is set; field errors render
- ARIA assertions for grid roles, column headers, `aria-live` region, popup wiring
- JS hook tests (`npm test`) for the roving-focus keyboard handler if implemented as a hook

## Non-goals (v1)

- Time selection (no DateTime, no time inputs)
- Presets rail ("Today", "Last 7 days" etc.) - document as a follow-up in the PR description
- Natural-language parsing ("tomorrow", "next friday")
- Localised month/day names beyond the attr mechanism (no cldr dependency; gettext wiring is documented, not implemented)
- Multi-calendar layouts beyond the two-month range pane

## Process

1. Build on a feature branch: `lib/petal_components/calendar.ex`, `lib/petal_components/date_picker.ex`, a `pc-calendar`/`pc-date-picker` section in `assets/default.css`, the keyboard hook in `assets/js` (registered in the hooks index), tests, and the playground page (nav entry + dials + the three scenarios). CSS/JS live in this repo - no external packages.
2. Verify in the playground (`mix run dev.exs`, http://localhost:4000) - exercise every dial, light + dark, keyboard-only, and `prefers-reduced-motion`.
3. Run the suites: `mix test` and `npm test` (the hook adds JS tests). All green before PR.

## Definition of done

- `<.calendar>` and `<.date_picker>` implemented per the API sketch (deviations flagged in the PR), no name collisions with `lib/petal_components.ex`, `Phoenix.Component`, or core_components
- Both LiveView and no-LiveView wirings work and are documented with moduledoc examples, including the gettext localisation recipe and the parse-on-blur contract
- `pc-*` CSS section added to `assets/default.css`; dark mode and `--pc-radius` respected
- Full test checklist above green (`mix test` + `npm test`)
- Playground page with all dials and the three scenarios; keyboard demo works
- PR screenshots light + dark
- Docs-ready attr docs: every attr has a `doc:` string; moduledocs include usage examples for both components

mplatts:
Closes #610

## Summary

Two components, shipped together as the issue asks: `<.calendar>`, a server-rendered month grid on Elixir's own `Date`, and `<.date_picker>`, that grid in a panel under a text input.

No new dependencies. Month maths is `Date.beginning_of_month/1`, `Date.end_of_month/1`, `Date.add/2` and `Date.day_of_week/1` plus modular arithmetic for the week rotation. Display formatting is `Calendar.strftime/2`. Nothing converts a timezone, because a `%Date{}` does not have one.

**`<.calendar>`**

- `single`, `range` (including the half-open `{from, nil}` mid-selection state) and `multiple`
- `min`/`max` inclusive windows, `disabled_dates` as a list or a 1-arity function. Disabled days render and stay focusable per the APG, they just lose their click wiring
- `starts_on` for all seven ISO week starts, `show_outside_days`, `fixed_weeks`, and a today marker (a dot under the number) that still reads when the day is also selected
- two wirings: `on_select`/`on_month_change` push `phx-click` with ISO `phx-value-date`/`phx-value-month`; with them nil the nav renders `<.link patch>` on `month_param`, which is a plain query-string anchor in a dead view
- the range band lives on the `<td>` so it runs edge to edge between days; the day button keeps its own rounding

**`<.date_picker>`**

- display input is strftime and unnamed, so it never posts; a hidden input carries ISO 8601 whatever `format` says. Range posts `name[from]` / `name[to]`
- parse on blur or Enter: ISO first, then a lenient parse of the configured format (padding-insensitive, case-insensitive, accepts `%B`/`%b` month names and discards `%A`/`%a`), and reverts to the last valid value on failure rather than silently clearing
- `two_months`, `clearable`, `disabled`, and the `<.field>` surface for label, errors, help text and the required marker, including `Phoenix.HTML.FormField` integration

## Scope

Everything in the issue's "Variants & states to build", the playground spec and the test checklist is in this PR. Deliberately **not** in it, per the issue's own Non-goals:

- time selection, natural-language parsing, a presets rail, cldr
- **the birthday year-select header**. The issue says to note it rather than half-ship it, so: the birthday scenario is on the playground page and it works, but it gets there via parse-on-blur (type `12 Jun 1987`, the grid follows) plus Shift+PageUp/PageDown for year jumps, not a decade dropdown. A year/month select header is a good follow-up and wants its own API decision (a `caption` slot, probably, so callers can put anything in the header).

One more follow-up worth naming: in `two_months` mode both arrows sit on the left pane and the right pane gets none, so paging always moves the pair by one month. Splitting them (prev on the left header, next on the right) needs the right pane's arrow to target base+1 rather than its own next month, which is a second nav-target attr I did not think earned its place yet.

All additive; nothing in the sketch was dropped or renamed.

- **`today` attr on both.** Defaults to `Date.utc_today/0`. The issue is right that a `Date` has no timezone, which is exactly why the component must not pick one for you: a user in Melbourne gets "today" wrong for ten hours a day if UTC is hardcoded. Passing a date is the honest fix, and it makes every test deterministic.
- **`day_names_long` attr.** `day_names` is 7 short labels, which is what a column header shows. The screen-reader name for the column needs the full word, and deriving "Monday" from "Mo" is not possible once the attr is localised. English defaults, same as the others.
- **`name` attr on `<.calendar>`.** The sketch says days "render as submit-capable buttons inside whatever form wraps the calendar" but a submit button needs a name to post anything. With `name` set (and no `on_select`) days render `type="submit" name=... value={iso}`, so the no-JS story is real rather than notional. `on_select` wins when both are set, and a disabled day never carries the name.
- **`nav` attr on `<.calendar>`** (`both` / `prev` / `next` / `none`). Two side-by-side panes rendering two independent sets of arrows was confusing in the playground. A suppressed arrow leaves a spacer so the caption stays optically centred.
- **`target`, `on_clear`, `range_separator`, `prev_label`/`next_label`/`open_label`/`clear_label`, `panel_class`, `month`/`on_month_change` on the picker.** `target` is for LiveComponents; the labels exist because otherwise the only localisable strings in the component would be hardcoded English; `range_separator` has to be an attr because the hook parses the display value back apart with it.
- **`<.date_picker>` does not literally wrap `<.popover>`.** The popover's `:trigger` slot renders *inside* a `<button>`, and a date picker's trigger is an `<input>` plus a toggle button. It reuses the popover's mechanics instead (`JS.show`/`JS.hide` with the same transitions, `phx-click-away`, `phx-keydown` Escape plus `JS.focus` back to the input) and its own `pc-date-picker__panel` surface.

## Hook decisions

Two hooks, both justified by the bar the issue sets.

**`PetalCalendar`** does roving tabindex and the arrow map. CSS cannot move focus and a server round trip per arrow key is not navigation, it is latency. Keyboard map: arrows by day and week, Home/End to the week bounds (respecting `starts_on`), PageUp/PageDown by month, Shift+PageUp/PageDown by year, Enter/Space to select (swallowed on a disabled day). Arrowing or paging out of the displayed month clicks the nav element and restores focus in `updated()`, so it works identically in event mode and link mode.

One thing worth flagging in review: **clicking the arrow as rendered only ever moves one month**, so a year jump landed eleven months short. I caught this driving the real playground, not in a unit test. The hook now retargets the arrow (`phx-value-month` in event mode, the month param in the `href` in link mode) before clicking it, which keeps `phx-target` and the patch URL as LiveView's own rather than forging an event. There are three JS specs pinning that.

**`PetalDatePicker`** does open-focus (move into the grid on open, back to the input on Escape), parse-on-blur, and client-side selection when no `on_select` is wired so the picker works in a dead view. When the server owns selection the hook stays out of the value entirely.

Both degrade: every day is a real `<button>`, so Tab plus Enter works with the hooks deleted. Only roving focus and the arrow map are lost.

## Verification

| check | result |
|---|---|
| `mix format --check-formatted` | pass |
| `mix compile --force --warnings-as-errors` | pass |
| `mix credo` | 45 issues, same 45 as `origin/main`. Zero new |
| `mix test` | 913 → **972**, 0 failures, 1 skipped (+59) |
| `npm test` | 157 → **195** (+38) |

Date edge cases have real tests: December to January in both directions, February 2024 vs 2026, `fixed_weeks` always six rows, the first and last week of the grid carrying the adjacent months, `min`/`max` inclusivity at both boundaries, all seven `starts_on` values, a backwards range pair getting ordered, and clamping 31 January to 28 February on a month page.

Verified by hand in the playground, keyboard only: ArrowRight then ArrowDown moved 12 Aug to 20 Aug, Home snapped to 17 Aug and End to 23 Aug, PageDown moved to September 2026 keeping the 23rd focused, and Shift+PageUp landed on September 2025 with focus intact. Light and dark both checked, and the two-month range panel was opened and screenshotted.


mplatts:
## Screenshots

**Date picker open** — two-month range panel, 12-16 Aug selected

Date picker open

**Keyboard** — focus ring on Sept 2025 after a Shift+PageUp year jump

Calendar keyboard navigation

**Calendar, light / dark**

Calendar, light
Calendar, dark

**Date picker, light / dark**

Date picker, light
Date picker, dark

### Verified independently

| Check | Result |
|---|---|
| `mix test` | 972 tests, 0 failures, 1 skipped (**+59**) |
| `npm test` | 195 passing (+38) |
| new dependencies | none — stdlib `Date`/`Calendar` only |
| slugs : clauses | 62 : 62, two components, two pages |

The largest issue in the set and it came in clean. Both components carry moduledocs, both are registered, and the 1:1 playground mapping holds at 62.

## The bug worth reading

Clicking the nav arrow as rendered only ever moves one month, so **year jumps landed eleven months short**. The hook now retargets the arrow (`phx-value-month`, or the month param in the `href`) before clicking, keeping `phx-target` and the patch URL as LiveView's own. Three JS specs pin it.

The author's own note: "I'd have shipped that broken on unit tests alone." That's the second time this run that driving the real browser caught something no unit test could — the other was File Upload's `:accept` string-vs-list. Both argue for the CONTRIBUTING rule that the playground page is part of the deliverable.

## Deferred deliberately, not silently

- **The birthday year-select header.** The issue says note it rather than half-ship it, and that's what happened — the birthday scenario works via parse-on-blur and Shift+PageUp/Down instead of a decade dropdown. A `caption` slot is the suggested follow-up API.
- **Split arrows in `two_months`.** Both arrows sit on the left pane and paging moves the pair by one month; splitting them needs a second nav-target attr the author judged hadn't earned its place. Your call.

The deviations are all additive and each has a concrete reason — `today` for timezone honesty (UTC is wrong for 10 hours a day in Melbourne), `day_names_long` because you can't derive "Monday" from a localised "Lu", `name` because a submit button needs one to post anything.

Worth noting the picker does **not** literally wrap `&lt;.popover&gt;`: popover's `:trigger` renders inside a `&lt;button&gt;` and a picker's trigger is an `&lt;input&gt;`, so it reuses the mechanics rather than the component. Reasonable, and stated rather than glossed.

nhobes:
Maintainer eye-test against shadcn/reui surfaced range + dark-mode styling issues. Root cause first, then the restyle — new commit:

**The "gray chips with gaps" and "gray bg + white text" complaints were ONE dark-mode bug.** `selected?/2` was true for every day in a range, so range middles wore `--selected` and relied on `--in-range` painting them back out. Light mode worked by source-order accident (both selectors 0-1-0); dark mode didn't — `dark:bg-primary-500` compiles to `.dark .pc-calendar__day--selected` (0-2-0) and outranked the override, so every day in a dark range rendered as its own rounded gray chip. Fixed structurally: the chip class is now `selected and not range_middle` — no specificity fight to lose.

**The restyle** (shadcn's structure, house tokens):
- **Selected/endpoints**: the maximal-contrast inverse chip — `bg-gray-900 text-white`, dark `bg-white text-gray-900`.
- **Range band**: continuous — `border-spacing: 0 0.25rem` (the horizontal 0 is load-bearing; commented), band on the cells edge-to-edge, endpoints round outward only at the band's radius so chip and band share one corner. Week-edge cases take their inward corner back in both directions. One-day and half-open ranges stay plain four-corner chips (`banded?/1`).
- **Middles**: subordinate — soft gray wash (dark rides gray-400/12 alpha per house convention), text one step back, deeper wash on hover, outside-days stay muted inside a band.
- **Rows**: 4px vertical breathing room.
- **Today dot inverts with the chip** (white-on-dark chip, dark-on-white chip) via a `--pc-calendar-dot` custom property on the button — a `dark:` variant applied straight to the `::after` compiles to a trailing `:where(.dark, .dark *)` that can never match (same trap the range-input track documents).
- Bonus hardening: disabled+selected in dark spelled out so the chip can't win.

**Measured**: 0px horizontal seams across every adjacent cell pair in both themes; endpoint chips fill their cells exactly with identical radii at the join; today+selected verified in dark (white chip, black number, black dot); two-month picker panel unaffected.

**Three taste calls flagged, not taken**: band and plain-day hover are both gray-100 in light (shadcn has the same collision); endpoint chips round at 10px vs plain chips' 6px (deliberate — they share the band's corner); row gap is 4px vs shadcn's 8px — one token if you want it airier.

Gates: `mix test` 979/0 (exit 0; calendar tests re-pinned to the new contract), format clean.

nhobes:
Maintainer follow-up on the restyle: completing a range visibly re-rounded the endpoint — a lone selected chip wears the day radius (~6px at default) but the band's outer ends wore the full `--pc-radius` (~10px), so the corners jumped the moment the second click landed. New commit unifies every day-shaped corner (chip, band outer ends, week-edge take-backs, nav buttons) onto one `--pc-calendar-day-radius` token on the calendar root — an endpoint is now the exact shape of the chip it was before the range existed, so nothing re-rounds. The date-picker panel keeps its container radius (it's a surface, not a day). This resolves the second of the three flagged taste calls. 979 tests exit 0, format clean.

nhobes:
Maintainer caught a regression from the radius-token commit: the calendar rendered square at every radius dial setting. Honest cause: the unification's search-and-replace matched its own definition line, leaving `--pc-calendar-day-radius: var(--pc-calendar-day-radius);` — a dependency cycle, which computes to guaranteed-invalid, so every consumer fell back to a 0 radius. Fixed, and `css_assets_test` now scans both shipped stylesheets for self-referential custom properties so this edit-class mistake fails the suite instead of shipping (verified the guard bites: reintroducing the cycle fails the test). 981 tests exit 0.

nhobes:
Two maintainer rulings from the comparison against reui/shadcn, recorded here:

1. **Drag-to-resize range endpoints: parked, demand-gated.** Click-built ranges cover the real use cases; drag-resize needs optimistic client-side range preview (the calendar renders server-side per click), month-edge auto-paging mid-drag, and touch/scroll arbitration — a large interaction surface for a marginal gain, and none of the sortable drag machinery transfers (moving DOM elements ≠ dragging a selection boundary across data). Same treatment as nav_menu's auto-flip: build when demand shows. If it's ever built, revisit the endpoint shape toward the floating-chip look at the same time — a draggable handle should look like one.

2. **Card/bordered presentation: composition, not variants** — matching the actual convention in both reference libraries (shadcn's calendar goes transparent inside card/popover slots and its bordered look is plain utilities; reui wraps theirs in a Card). Showcase examples demonstrating composition are being added rather than any `variant` attr, along with a `--pc-calendar-cell-size` token and a `:day` slot for booking-style custom cell content (pricing under dates) — the one real capability gap against both libraries.

nhobes:
The parity gap against reui/shadcn is closed — new commit adds the two capabilities plus the composition examples:

**`--pc-calendar-cell-size`** (default 2.25rem, today's rendering byte-identical — proven by md5/cmp across 9 configurations against a pristine pre-change server). One token sizes days, weekday header widths, and (via a derived `--pc-calendar-nav-size` preserving the one-step-smaller relationship) the nav arrows, so the grid stays in register at any size. Deliberately a CSS token and not a `size` attr — same contract as `--pc-radius`: set it in a class or style anywhere above the calendar. Key implementation discipline: the token is *read with a fallback, never declared on the calendar root* — declaring it there would shadow a wrapper's value and break the documented `<div style=...>` path (pinned by test).

**`:day` slot** — replaces the day button's content (the consumer renders the number); receives the day map the grid already builds (`:date`, `:iso`, `:today`, `:outside`, `:disabled`, `:selected`, `:range_start/_middle/_end`). `:selected` is load-bearing — it's how a muted price line stays readable when the inverse chip lands under it. State classes, wiring, roving tabindex and the full-date `aria-label` remain the component's job; the slot is what a sighted user sees, not what a screen reader is told. Empty slot = the bare number, byte-identical.

**Examples**: `:booking` (3.5rem cells, weekend/weeknight pricing under each date, range mode — the reui pricing-calendar look in house style) and `:in_card` (composition inside `<.card>` — zero calendar CSS changes needed, it already computes transparent bg). The playground gains a cell-size dial (2/2.25/3/4rem) written as an inline style, since a runtime-built arbitrary-value class is one Tailwind's scanner never sees.

CDP-verified both themes on a patched-vs-pristine server pair: default probes identical on every field; booking grid band continuity + chip legibility + today-dot clearance all hold at 3.5rem; header/cell columns stay aligned at 4rem.

Gates: `mix test` 988/0 (exit 0, +7 specs), format clean, --warnings-as-errors clean.

nhobes:
Three maintainer-requested examples, new commits:

- **`:availability`** — the artist-booking pattern: `disabled_dates` (list or 1-arity function) blocks booked days, and the `:day` slot crosses them out via the `:disabled` flag. Real find during verification: the naive strike inherited the disabled button's `text-gray-300` and probed **1.48:1** in light — illegible. The example lifts struck days to gray-400/gray-500 (2.62:1 / 3.97:1 — clearly muted, clearly readable, and `text-decoration-color` tracks the span so strike and digits mute together).
- **`:appointments`** — the composed picker: `<.card>` with header, calendar + scrollable time-slot rail side by side, and a **bordered `card_footer`** carrying the summary + confirm. One deliberate taste call: slot buttons are gray outline (selected = gray solid), not primary — flipping the playground palette dial to blue showed seven blue-outlined slots fighting the blue Confirm; with gray, the card's only primary element is the action. The rail is a plain `max-h-56 overflow-y-auto` div since scroll_area lives on the unmerged #630 — swapping it in post-merge is a noted follow-up.
- **`:booking`** gains the deal-price accent: cheapest fares in the success colour, selected-chip prices explicitly not green (the `cond` puts `:selected` first — chip legibility probes 9.14:1 light / 6.57:1 dark). Gatekeeper adjustment on the agent's honest flag: `success-600` at `text-xs` probes **3.22:1** on white, under the AA 4.5:1 line — the light stop is `success-700` (commented why); dark keeps the house `success-400`.

All probes are measured (contrast ratios, rects, computed border tokens), both themes. `mix test` 988/0 exit 0, format clean, zero warnings.

nhobes:
Maintainer eye-test find on /c/date-picker, root-caused and fixed in the new commit — the flagship range picker rendered patchy pre-restyle chips while the Booking example looked perfect.

**Why they differed**: two selection paths. Booking carries `data-select-event`, so the SERVER re-renders every click through the restyled anatomy. The flagship is client-owned: `PetalDatePicker.paint/2` stamps selection classes in the browser — and it still spoke the pre-restyle vocabulary (`--selected` on every range day, no day-level rounding classes — a comment even asserted they "deliberately do not exist" — cell band classes toggled without the `banded?` guard, and the server's rounding classes never cleared).

**Fix**: `paint` now mirrors `build_day/4` exactly — endpoints-only chips, `banded?`-guarded band classes on cells AND days, `--in-range` middles, honest `aria-selected` — plus `updated()` re-binds and repaints from the hidden inputs after any LiveView patch, so the grid can never disagree with what will post.

**The parity lock (the real deliverable)**: `test/fixtures/calendar_selection_classes.json` — one fixture, six scenarios, read by BOTH suites (Jason / node:fs), comparing every rendered day. The twin tests cross-reference each other, and flipping one expected class turns both suites red at once — the two writers of this contract can no longer drift silently.

**CDP, both themes**: the flagship's client-painted digest (classes + aria + computed backgrounds/text/radii/dot var per day) is byte-identical to the server-painted Booking picker for the same dates, including a 27 Aug → 2 Sep range crossing a week AND month boundary across two panes; clear/re-pick/range-restart/nav-patch all land on the server matrix.

Gates: `mix test` 989/0, vitest 222 (+11 parity specs, −1 spec that pinned the OLD contract; 10 of 11 fail with the hook fix stashed — clear was the one thing the old painter got right), format clean.

nhobes:
Second maintainer find from the same eye-test pass: the date-picker's trailing adornments (clear × + calendar toggle) diverged from the combobox's — solid-weight X at a shared generic size, tighter spacing, optically misaligned pairing. New commit brings them to the combobox's contract, and the measurements found real bugs underneath:

- **The two buttons overlapped by −4px**, silently killing the clear's right 4px of tap area; and `pr-10` let long values run 16px UNDER the clear button (`:has` bumps it to `pr-16` when the clear renders).
- Clear is now `hero-x-mark-mini` at the combobox's exact clear-icon size (ink 8.03 ≡ combobox's), both adornments share a 16px glyph box in 24×24 chips, 6px apart (the combobox's gap), same muted colors and hover wash (byte-identical computed hover values both schemes), house invisible-extension keeping each strip field-height.
- **No optical nudge — measured before deciding**: `hero-calendar`'s ink is dead-centred in its box (0.00px offset at both sizes); the residual weight difference is heroicons' own sparse-vs-full grid, and both alternatives measured worse. Glyph-box centre delta between the pair: 0.00px (the combobox's own pair is 0.86px apart).
- The contract is a CSS comment naming combo_box as the precedent: mini-weight icons, matched boxes, clear leftmost, 6px gap, tap-sized hits.

**Clearables audit** (13 surfaces, in the report): follow-ups outside this file's scope — the combobox's own trigger-variant clear gap is 3px vs its input variant's 6px; `pc-clearable-field-icon` (field/input) is still a 20px solid X with no dark rest colour; the date-input icon is a div in field.ex but a button in input.ex; the data-table filter clear diverges on shape/wash. Flagged for a later sweep, not smuggled in here.

Gates: `mix test` 989/0, vitest 222, format clean; hit-tested tap areas 24×41 / 32×41 at 1280 and 375.

